PR #5, opened 2026-07-16
fix(install): skip CLAUDE.md pointer when install destination is gitignored

### LINKED ISSUE #4: install.sh (non-dev): appends tracked CLAUDE.md pointer even when target repo gitignores the install destination
## Summary

When installing into a consumer repo that **gitignores the install destination** (`.claude/commands` and/or `.claude/skills`), the non-dev `install.sh` path writes the command/skill files (which are then ignored → machine-local) but **still appends a tracked pointer block to `CLAUDE.md`**. The result is an inconsistency: the committed pointer advertises `/repo:*` commands whose files are not committed.

Discovered installing Repo Skills v0.4.0 into [rjwalters/loom](https://github.com/rjwalters/loom) on 2026-07-16.

## Repro

In a target repo whose `.gitignore` contains `.claude/commands` and `.claude/skills` (Loom does this intentionally — co-install hygiene so a foreign tool's files don't pollute Loom's distribution artifact):

```bash
./install.sh -y /path/to/loom
git -C /path/to/loom status --short
```

Observed:

```
 M CLAUDE.md          # tracked — the REPO-SKILLS pointer block
```

The 14 `.claude/commands/repo/*.md` files and `.claude/skills/repo/SKILL.md` are silently gitignored:

```
$ git check-ignore -v .claude/commands/repo/release.md
.gitignore:104:.claude/commands   .claude/commands/repo/release.md
```

So the operator is left with a choice, both bad:
- **Commit `CLAUDE.md`** → teammates who clone get a committed "This repo has Repo Skills v0.4.0 installed — run `/repo:help`" pointer to command files they don't have (and won't, until they run the installer themselves).
- **Don't commit** → the pointer block lingers as a permanent uncommitted diff in the working tree.

## Why this is inconsistent with existing behavior

The installer's **dev-mode** path already reasons about exactly this case and does the right thing — it ensures `.claude/` is gitignored and **deliberately skips the CLAUDE.md block**:

> `# 4. CLAUDE.md block ... Skipped in dev mode: the symlinked install is machine-local ... instead of advertising it in a committed CLAUDE.md we ensure .claude/ is gitignored and leave CLAUDE.md untouched.`

The **non-dev** path has no symmetric detection: it appends the CLAUDE.md pointer unconditionally, even when the destination it just wrote to is gitignored (i.e. the install is effectively machine-local anyway).

## Suggested fix

In the non-dev path, before appending the CLAUDE.md block, check whether the install destination is gitignored in the target (e.g. `git -C "$TARGET" check-ignore -q .claude/commands/repo` / `.claude/skills/repo`). If the files won't be tracked, treat it like dev mode for the CLAUDE.md step — skip the pointer block (or print a notice and require an explicit `--claude-md` opt-in), since a committed pointer to uncommitted files is never what the operator wants.

Alternatively: detect the mismatch and warn loudly at the end of install (`⚠ install destination is gitignored but a tracked CLAUDE.md pointer was written; you probably don't want to commit it`), leaving the decision to the operator.

## Severity: low

The install functions correctly and the `/repo:*` commands work in-session. This is a hygiene/consistency papercut, not a breakage — but it forces an awkward commit-or-not decision on every install into a Loom-style repo, and a naive `git commit -a` would ship a misleading pointer.

---

### Root cause (grounded in current `install.sh`)

`install.sh` has one branch that already handles the "install is effectively machine-local" case and one that does not:

- **Dev-mode CLAUDE.md skip** — lines **203–216**. The `if [[ "$DEV" == true ]]` block (line 207) deliberately does *not* write the pointer: it ensures `.claude/` is gitignored (lines 208–212) and `exit 0` (line 215) before the CLAUDE.md logic runs. The rationale comment is on lines 204–206.
- **Non-dev CLAUDE.md pointer-append** — lines **218–246**. `CLAUDE_MD="$TARGET/CLAUDE.md"` (line 218), the marker-bounded block is built into `$BLOCK_FILE` (lines 222–231), then it is either spliced in place via `awk` (lines 233–241, when `$MARKER_BEGIN` already exists) or appended (lines 242–245, `else` branch). **This runs unconditionally** — there is no check for whether the destination the installer just wrote to (`$TARGET/.claude/commands/repo`, `$TARGET/.claude/skills/repo`) is gitignored in `$TARGET`.

Relevant path/variable names:
- `TARGET` — absolute path to the consumer repo (resolved at line 100).
- Install destinations written earlier: `$TARGET/.claude/skills/repo` (line 179) and `$TARGET/.claude/commands/repo` (line 184).
- `MARKER_BEGIN` / `MARKER_END` (lines 41–42) bound the pointer block.
- `warning()` helper (line 35) already exists for the notice path.

### Recommended approach

Add a gitignore-detection guard immediately before the non-dev CLAUDE.md logic (i.e. right after `CLAUDE_MD="$TARGET/CLAUDE.md"` on line 218, before the `$BLOCK_FILE` build on line 222). Mirror the dev-mode reasoning:

1. Probe the destination with a **quiet** `check-ignore` against the exact subdirectories the installer wrote to:
   ```bash
   if git -C "$TARGET" check-ignore -q .claude/commands/repo .claude/skills/repo 2>/dev/null; then
     # destination is gitignored → committed pointer to uncommitted files
     warning "Install destination (.claude/commands, .claude/skills) is gitignored in $TARGET;"
     warning "skipping the CLAUDE.md pointer block (a committed pointer to uncommitted command files"
     warning "is not what you want). The /repo:* commands still work in-session."
     echo ""
     success "Repo Skills v$VERSION installed. Try /repo:help in Claude Code."
     rm -f "$BLOCK_FILE"   # if BLOCK_FILE built before the guard; otherwise omit
     exit 0
   fi
   ```
   Place the guard *before* `$BLOCK_FILE` is created so no temp file leaks, and keep the final success line so the operator still gets a clean "installed" confirmation.

2. **Notes on `check-ignore` semantics** (important for correctness):
   - `git check-ignore -q A B` exits `0` if **any** path is ignored, so pass both destinations to catch a repo that ignores only one. If the fix must require *both* to be ignored before skipping, invoke once per path and `&&` them. Recommendation: skip if **either** destination is ignored, since a split state is itself broken.
   - Run the probe relative to `$TARGET` via `git -C "$TARGET"`. The paths are repo-relative (`.claude/commands/repo`), matching the issue's repro (`.gitignore:104:.claude/commands`).
   - Guard against `$TARGET` not being a git repo: `check-ignore` returns non-zero (exit 128) outside a repo, which correctly falls through to the normal append path. The `2>/dev/null` suppresses the noise. The installer already tolerates non-git targets (lines 106–112).
   - The `--dry-run` output (lines 147–158) unconditionally lists `CLAUDE.md (marker-bounded REPO-SKILLS block)` on line 156. If feasible, reflect the skip there too (e.g. detect gitignore and print `CLAUDE.md (skipped — destination gitignored)`), but this is secondary — the primary fix is the real install path.

3. Prefer the **skip + notice** behavior (Suggested fix, option 1) over the warn-but-write-anyway alternative: it matches dev mode's precedent and avoids leaving an uncommitted diff. An opt-in flag (e.g. `--claude-md`) to force-write is optional and out of scope unless trivially added.

### Scope guard

- Single-file change to `install.sh`; well under the size/complexity thresholds. Do not refactor the dev-mode branch or the awk-splice logic.
- Do not change what the pointer block says, the marker format, or the copy/render pipeline.
- The `--claude-md` force-write opt-in is **optional / out of scope** unless implemented in the same trivial guard; the core deliverable is the skip-when-gitignored behavior + notice.

## Acceptance Criteria

- [ ] Installing (non-dev) into a repo whose `.gitignore` ignores `.claude/commands` (and/or `.claude/skills`) leaves `CLAUDE.md` **byte-for-byte unchanged**: `git -C "$TARGET" status --short` shows no `M CLAUDE.md` after `./install.sh -y "$TARGET"`.
- [ ] In that gitignored case, the installer prints a clear notice explaining the CLAUDE.md pointer was skipped because the destination is gitignored (and that `/repo:*` still works in-session).
- [ ] Installing (non-dev) into a **normal** repo (destination not gitignored) still appends/updates the marker-bounded REPO-SKILLS block in `CLAUDE.md` exactly as before — no regression to lines 218–246 behavior.
- [ ] The re-install / in-place update path (existing `MARKER_BEGIN` present, lines 233–241) is unaffected when the destination is not gitignored: an existing block is still replaced, not duplicated.
- [ ] Installing into a target that is **not** a git repo still writes the CLAUDE.md block (the `check-ignore` probe fails safe / falls through), matching current behavior.
- [ ] No temp `$BLOCK_FILE` is left behind on the skip path (`mktemp` either not called before the guard, or cleaned up on the early exit).
- [ ] `--dev` behavior is unchanged (still skips CLAUDE.md and gitignores `.claude/`, lines 203–216).

## Affected Files

- `install.sh` - Add a gitignore-detection guard in the non-dev CLAUDE.md path (immediately after line 218, before building `$BLOCK_FILE` at line 222): probe `git -C "$TARGET" check-ignore -q .claude/commands/repo .claude/skills/repo`; on match, print a `warning()` notice, print the final `success` line, and `exit 0` without writing/updating `CLAUDE.md`. Optionally reflect the skip in the `--dry-run` output (line 156).

## Test Plan

- [ ] Manual — gitignored destination: create a temp git repo, add `.claude/commands` and `.claude/skills` to its `.gitignore`, run `./install.sh -y "$repo"`, confirm `git -C "$repo" status --short` shows the `.claude/**` files ignored and **no** `CLAUDE.md` diff, and that the skip notice was printed.
- [ ] Manual — normal destination: create a temp git repo with no such `.gitignore` entries, run the installer, confirm the REPO-SKILLS block is appended to `CLAUDE.md` and the `.claude/**` files are tracked/untracked (not ignored).
- [ ] Manual — re-install idempotency (normal repo): run the installer twice; confirm the CLAUDE.md block is updated in place (single block, not duplicated).
- [ ] Manual — non-git target: run against a plain directory (answer `y` to the not-a-git-repo prompt or pass `-y`); confirm the CLAUDE.md block is still written (probe fails safe).
- [ ] Manual — dev mode unchanged: `./install.sh --dev .` still adds `.claude/` to `.gitignore` and leaves `CLAUDE.md` untouched.
- [ ] Automated: N/A (no test harness in repo; verification is via the manual installer scenarios above).
- [ ] Edge cases: repo that ignores only `.claude/commands` but not `.claude/skills` (skip should still trigger); `--dry-run` output review.

rjwalters:
## Summary

The non-dev `install.sh` path appended a tracked `REPO-SKILLS` pointer block to `CLAUDE.md` unconditionally — even when the target repo gitignores `.claude/commands` / `.claude/skills` (as Loom-style repos do). The result was a committed pointer advertising `/repo:*` commands whose files are never tracked, forcing an awkward commit-or-not decision on every install. Dev mode already avoids this; this brings the non-dev path to parity.

## Changes

- `install.sh`: added a gitignore probe in the non-dev CLAUDE.md path, immediately after `CLAUDE_MD="$TARGET/CLAUDE.md"` and **before** any `$BLOCK_FILE` is created. If either destination (`.claude/commands/repo`, `.claude/skills/repo`) is gitignored in `$TARGET`, it prints a `warning()` notice, prints the final `success` line, and `exit 0` — mirroring dev mode and leaving `CLAUDE.md` untouched with no temp file leaked.
- Probe each destination separately and OR the results: `git check-ignore -q` rejects multiple pathnames (`--quiet is only valid with a single pathname`), so the curator's single two-arg invocation would have silently failed to 128 and never triggered. Skipping if *either* path is ignored also correctly covers the split-state edge case.
- Mirrored the skip in `--dry-run` output: shows `CLAUDE.md (skipped — install destination is gitignored)` for gitignored targets and a `.gitignore` line for dev mode, else the normal marker-block line.

| Gitignored dest leaves CLAUDE.md byte-for-byte unchanged | done | S1: shasum before==after; `git status --short` shows no `M CLAUDE.md` |
| Clear skip notice printed | done | S1: prints "Install destination ... is gitignored ... skipping the CLAUDE.md pointer block ... /repo:* commands still work in-session" |
| Normal dest still appends/updates the block (no regression) | done | S2: `<!-- BEGIN REPO-SKILLS -->` appended; dest confirmed not gitignored |
| Re-install idempotency (block replaced, not duplicated) | done | S3: second run prints "Updated REPO-SKILLS block"; BEGIN marker count == 1 |
| Non-git target still writes the block (fail-safe) | done | S4: probe returns non-zero outside a git repo; block written |
| No temp `$BLOCK_FILE` left behind on skip | done | Guard exits before `mktemp` is called; tmp-dir count unchanged before/after |
| `--dev` unchanged (skips CLAUDE.md, gitignores `.claude/`) | done | S5: CLAUDE.md shasum unchanged; `.claude/` added to `.gitignore` |
| Only-one-of-two destinations ignored still skips | done | S6: `.gitignore` with only `.claude/commands`; CLAUDE.md unchanged |
| `--dry-run` reflects the skip | done | S7: gitignored dry-run shows "skipped"; normal dry-run shows the marker block |

## Test Plan

Ran `./install.sh -y <target>` against seven scratch repos under a temp dir (git repos created on the fly): gitignored (both dests), normal, re-install of normal, non-git dir, `--dev`, split-state (one dest ignored), and `--dry-run` for gitignored vs normal. All seven scenarios passed. `bash -n install.sh` clean.

Closes #4

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -153,7 +153,14 @@ if [[ "$DRY_RUN" == true ]]; then
   while IFS= read -r cmd; do
     echo "  $TARGET/.claude/commands/repo/$cmd.md"
   done <<<"$COMMANDS"
-  echo "  $TARGET/CLAUDE.md (marker-bounded REPO-SKILLS block)"
+  if [[ "$DEV" == true ]]; then
+    echo "  $TARGET/.gitignore (.claude/ entry; CLAUDE.md skipped in dev mode)"
+  elif git -C "$TARGET" check-ignore -q .claude/commands/repo 2>/dev/null \
+    || git -C "$TARGET" check-ignore -q .claude/skills/repo 2>/dev/null; then
+    echo "  $TARGET/CLAUDE.md (skipped — install destination is gitignored)"
+  else
+    echo "  $TARGET/CLAUDE.md (marker-bounded REPO-SKILLS block)"
+  fi
   exit 0
 fi
 
@@ -216,6 +223,28 @@ if [[ "$DEV" == true ]]; then
 fi
 
 CLAUDE_MD="$TARGET/CLAUDE.md"
+
+# If the install destination is gitignored in the target, the command/skill
+# files we just wrote are effectively machine-local (they won't be committed),
+# so a tracked CLAUDE.md pointer would advertise /repo:* commands whose files
+# aren't in the repo. Mirror dev mode here: skip the CLAUDE.md block entirely.
+# Probe each destination separately (check-ignore -q accepts only one pathname)
+# and skip if *either* is ignored, since a split state is itself broken. The
+# probe exits non-zero outside a git repo or when nothing is ignored, which
+# correctly falls through to the write path.
+dest_is_gitignored() {
+  git -C "$TARGET" check-ignore -q .claude/commands/repo 2>/dev/null \
+    || git -C "$TARGET" check-ignore -q .claude/skills/repo 2>/dev/null
+}
+if dest_is_gitignored; then
+  warning "Install destination (.claude/commands, .claude/skills) is gitignored in $TARGET;"
+  warning "skipping the CLAUDE.md pointer block (a committed pointer to uncommitted command"
+  warning "files is not what you want). The /repo:* commands still work in-session."
+  echo ""
+  success "Repo Skills v$VERSION installed. Try /repo:help in Claude Code."
+  exit 0
+fi
+
 # The block is intentionally lightweight: a pointer to the real docs, not an
 # inlined command dump. `/repo:help` and SKILL.md carry the authoritative,
 # always-current command list; duplicating it here just goes stale.


rjwalters:
✅ **Approved!**

Solid, well-scoped fix that brings the non-dev CLAUDE.md path to parity with dev mode. I independently verified every acceptance criterion by running the actual PR branch's `install.sh` against scratch git repos (not relying on self-reported output).

## Correctness verification

**The load-bearing `check-ignore` claim is correct — and the builder caught a real bug in the curator's suggested probe.** I reproduced it (git 2.55.0):

```
$ git check-ignore -q .claude/commands/repo .claude/skills/repo
fatal: --quiet is only valid with a single pathname
exit=128
```

The curator's single two-arg `git check-ignore -q A B` would have fataled to 128 every time (non-zero → fall through to the write path), so the skip would **never** have triggered. The per-path probe-and-OR (`dest_is_gitignored()`) is the right correction, and it also naturally covers the split-state edge case.

## Acceptance criteria — all pass (verified end-to-end)

| Criterion | Result |
|-----------|--------|
| Gitignored dest → CLAUDE.md byte-for-byte unchanged | ✅ shasum identical before/after; no `M CLAUDE.md` |
| Clear skip notice printed | ✅ `warning()` notice + `/repo:*` in-session note |
| Normal dest still appends block | ✅ block appended, one BEGIN marker |
| Re-install idempotency | ✅ 2nd run "Updated ... in place", marker count stays 1 |
| Non-git target fails safe (writes block) | ✅ probe non-zero outside repo → block written |
| No temp `$BLOCK_FILE` leaked on skip | ✅ guard `exit 0`s *before* `mktemp` at line 222 (structural) |
| `--dev` unchanged | ✅ CLAUDE.md untouched, `.claude/` gitignored |
| Split-state (only one dest ignored) | ✅ skip triggers whether only `.claude/commands` or only `.claude/skills` is ignored |
| `--dry-run` reflects skip | ✅ gitignored → "skipped — install destination is gitignored"; normal → marker-block line; dev → `.gitignore` line |

## Shell correctness

- `git -C "$TARGET"` scopes the probe to the target repo; `2>/dev/null` suppresses the exit-128 noise on non-git targets; quoting is correct throughout.
- Guard is placed immediately after `CLAUDE_MD=` and before the `mktemp`, so the append/awk-splice/marker-idempotency logic (lines 233–246) is completely untouched on the normal path — no regression.
- `bash -n` clean; dry-run branch reuses the same per-path OR for consistency.

Merge state is CLEAN. This repo has no configured CI workflows (`.github/workflows` absent), so "no checks reported" is expected rather than a pending/failing state. PR body uses `Closes #4`.

Nice work correcting the probe rather than copying the issue's suggestion verbatim.

PR #10, opened 2026-07-16
feat(remote): load shared cloud creds from ~/.config/repo/remote.env

rjwalters:
## Summary

`repo:remote` currently reads **everything** — cloud provisioning credentials *and* per-repo machine config — from a single `.env` at the repo git root. That forces you to paste the same AWS IAM keys / GCP project into every repo's `.env` you want to launch from.

This makes config **two-layer**, loaded in order:

1. **`~/.config/repo/remote.env`** (shared, loaded first) — cloud identity reused across every repo: `AWS_ACCESS_KEY_ID`/`AWS_SECRET_ACCESS_KEY`/`AWS_REGION` or `GCP_PROJECT`/`GCP_ZONE`/`GOOGLE_APPLICATION_CREDENTIALS`, plus shared defaults like `REPO_REMOTE_PROVIDER` and `REPO_REMOTE_SSH_KEY`. Honors `$XDG_CONFIG_HOME`.
2. **`<repo>/.env`** (per-repo, loaded second → **overrides**) — `REPO_REMOTE_*` machine/software/session settings and the pinned `REPO_REMOTE_INSTANCE_ID`.

Because the repo file is sourced last, a repo that needs a different account/region can still override any shared value.

## Changes (prose-only, `commands/repo/remote.md`)

- **Config section** rewritten to document the two layers and which vars belong where.
- **Step 2 load snippet** now sources `${XDG_CONFIG_HOME:-$HOME/.config}/repo/remote.env` first, then the repo `.env`.
- **`--configure`** writes credentials to the shared file by default (offers repo `.env` for a repo-specific account), machine settings to the repo `.env`, and protects both (gitignore the repo `.env`, `chmod 600` the shared file).
- **Instance-ID writeback** explicitly always targets the **repo** `.env` (per-repo handle), never the shared file.
- **Hygiene / Safety Rule 4** updated for the two-file model.
- Backward-compatible: a repo `.env` that still carries creds keeps working (it just overrides an absent shared layer).

## Test / verification

Prose command-doc change — verified by reading: the load order produces repo-overrides-home precedence, the `${XDG_CONFIG_HOME:-$HOME/.config}` expansion is correct, and the credential-hygiene + writeback rules are internally consistent.

### DIFF
--- a/commands/repo/remote.md
+++ b/commands/repo/remote.md
@@ -12,15 +12,17 @@ Stand up (or reuse) a cloud VM with this repository cloned and synced, then
 open a live SSH session in a new terminal window — landing in the repo, ready
 to run `claude` and continue the work on the remote host.
 
-Configuration and credentials are read from the target repo's **`.env`**
-(see below). The provisioning credentials there are used locally to drive the
-cloud CLI; they are **never** copied to the VM.
+Configuration is read from two layers (see below): shared cloud credentials
+from **`~/.config/repo/remote.env`** (reused across every repo), with the
+target repo's **`.env`** layered on top for per-repo machine settings — and
+free to override any shared value. The provisioning credentials are used
+locally to drive the cloud CLI; they are **never** copied to the VM.
 
 ## Usage
 
 ```
 /repo:remote                   # Read .env, bring up / reuse the host, open SSH
-/repo:remote --configure       # Guided setup: create/update .env (provider, creds, machine)
+/repo:remote --configure       # Guided setup: shared creds (~/.config/repo/remote.env) + repo .env (machine)
 /repo:remote gcp               # Override REPO_REMOTE_PROVIDER for this run
 /repo:remote aws
 /repo:remote --status          # List instances created by this command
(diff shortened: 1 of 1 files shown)


rjwalters:
✅ **Approved!**

Prose-only change to `commands/repo/remote.md` making `repo:remote` config two-layer. Verified by close reading of the diff and the full file. All design-intent points hold.

**1. Load-order precedence — CORRECT (repo overrides shared).** Step 2 snippet sources the shared file first, then the repo `.env`:
```bash
CONFIG_HOME="${XDG_CONFIG_HOME:-$HOME/.config}/repo/remote.env"
set -a
[ -f "$CONFIG_HOME" ] && . "$CONFIG_HOME"
[ -f "$(git rev-parse --show-toplevel)/.env" ] && . "$(git rev-parse --show-toplevel)/.env"
set +a
```
With `set -a`, the later source wins, so the repo `.env` overrides the shared file — exactly the repo-wins precedence intended. The prose ("Repo values override shared ones because the repo file is sourced last") matches the code.

**2. Path + optionality.** `${XDG_CONFIG_HOME:-$HOME/.config}/repo/remote.env` is a correct expansion honoring `$XDG_CONFIG_HOME` with the `$HOME/.config` fallback. Both `[ -f … ] && . …` guards make each file optional — a repo `.env` still carrying creds keeps working with no shared file present. Backward-compatible.

**3. Instance-ID writeback (Step 4)** explicitly targets the repo `.env`: "write the new ID back to the repo's `.env` (the git root, never the shared file — the instance handle is per-repo)." Correct.

**4. Credential hygiene** is coherent for the two-file model: repo `.env` must be gitignored (refuse to run otherwise); shared file should be `chmod 600`. Safety Rule 4 updated to match.

**5. `--configure` wizard** writes creds to the shared file by default and machine settings to the repo `.env`, protects both, and offers repo-`.env` creds only for a repo-specific account. No contradictions between the Config section, wizard, load step, and Safety Rules.

**6. No precedence-inverting stale phrasing.** Main-body prose consistently describes the two layers; `--status` and writeback references are consistent with the model.

**7. Embedded bash is shell-valid** (`bash -n` passes).

Merge state CLEAN; no CI checks configured on this docs repo (expected). No `Closes #N` — expected for this direct feature change per the operator.

PR #11, opened 2026-07-16
feat(remote): dogfood dev env — Claude Code + multi-account token pool + gh label auth

rjwalters:
## Summary

Makes `/repo:remote` stand up a **working dev session** on the VM, not just a bare host — verified end-to-end by dogfooding it on a live `t3.small`. Adds the dev-environment Dockerfile, installs the Claude Code CLI, and introduces **dev-session auth** as a credential class distinct from provisioning creds.

Builds on the two-layer config from #10 (shared `~/.config/repo/remote.env` + repo `.env` override).

## What's in here

- **`Dockerfile`** (new) — Ubuntu 24.04 dev image with the repo-skills toolchain (bash, git, gh, shellcheck, ripgrep, jq) **plus the Claude Code CLI**. Auth is injected at runtime, never baked in.
- **`commands/repo/remote.md`** — dev-session auth:
  - **Two secret classes.** Provisioning creds (`AWS_*`/`GCP_*`) stay local, never touch the VM. Dev-session auth (`REPO_REMOTE_GH_TOKEN`, the `ACCOUNT_*` Claude pool) goes to the VM *by design*.
  - **Multi-account Claude pool** (the Loom pattern): `ACCOUNT_EMAIL/KEY/TOKEN_FILE_N` registry + `.loom/tokens/*.token` files, resolved **shared-then-repo (repo wins)**, copied to the VM's `.loom/tokens/` at chmod 600 so a Loom install there rotates accounts. Account 1 becomes the default `CLAUDE_CODE_OAUTH_TOKEN`.
  - **gh token** injected inline over SSH into the container env (no file on disk). PAT guidance: fine-grained, repo-scoped, **Contents + Issues + Pull requests** write so Loom-style label workflows work.
  - Safety rules split the two classes; teardown wipes the pool with the disk. Note that `gh` repo-scoped commands need the clone path (or `-R owner/repo`).
- **`.gitignore`** — ignore the per-repo `.env`.

## Verified live (on a real t3.small)

- Container builds; Claude Code v2.1.211 + gh 2.96.0 present.
- Claude **account 1** authenticates (`AUTHOK`); when it hit its session limit mid-test, **account 2** authenticated (`ROTATEOK`) — the pool proving its exact purpose.
- `gh` authenticates via `GH_TOKEN`; **issue and PR label add/remove** both succeed from the VM container.
- Provisioning `.env` correctly excluded from the VM sync.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,3 +32,6 @@
 .loom/state.json
 .loom/*.log
 .loom/*.sock
+
+# repo:remote per-repo config (may hold overrides; never commit)
+.env
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,24 @@
+# Dev environment for working on repo-skills — and for dogfooding /repo:remote.
+# /repo:remote builds and runs this on the remote host when the repo's .env sets
+# REPO_REMOTE_DOCKERFILE=./Dockerfile. It carries the toolchain needed to work on
+# the /repo:* commands (bash, git, gh, shellcheck, ripgrep, jq). Claude Code is
+# installed in-session by /repo:remote, not baked in, to keep the image lean.
+FROM ubuntu:24.04
+
+ENV DEBIAN_FRONTEND=noninteractive
+
+RUN apt-get update && apt-get install -y --no-install-recommends \
+        bash git ca-certificates curl jq ripgrep shellcheck less openssh-client \
+    && rm -rf /var/lib/apt/lists/*
+
+# GitHub CLI (gh) from GitHub's official apt repo
+RUN curl -fsSL https://cli.github.com/packages/githubcli-archive-keyring.gpg \
+        -o /usr/share/keyrings/githubcli-archive-keyring.gpg \
+    && chmod go+r /usr/share/keyrings/githubcli-archive-keyring.gpg \
+    && echo "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/githubcli-archive-keyring.gpg] https://cli.github.com/packages stable main" \
+        > /etc/apt/sources.list.d/github-cli.list \
+    && apt-get update && apt-get install -y --no-install-recommends gh \
+    && rm -rf /var/lib/apt/lists/*
+
+WORKDIR /work
+CMD ["sleep", "infinity"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,9 @@
 # Dev environment for working on repo-skills — and for dogfooding /repo:remote.
 # /repo:remote builds and runs this on the remote host when the repo's .env sets
 # REPO_REMOTE_DOCKERFILE=./Dockerfile. It carries the toolchain needed to work on
-# the /repo:* commands (bash, git, gh, shellcheck, ripgrep, jq). Claude Code is
-# installed in-session by /repo:remote, not baked in, to keep the image lean.
+# the /repo:* commands (bash, git, gh, shellcheck, ripgrep, jq) plus the Claude
+# Code CLI. Auth tokens (gh, claude) are injected at container-run time via the
+# environment — never baked into this image and never committed.
 FROM ubuntu:24.04
 
 ENV DEBIAN_FRONTEND=noninteractive
@@ -19,6 +20,11 @@ RUN curl -fsSL https://cli.github.com/packages/githubcli-archive-keyring.gpg \
         > /etc/apt/sources.list.d/github-cli.list \
     && apt-get update && apt-get install -y --no-install-recommends gh \
     && rm -rf /var/lib/apt/lists/*
+
+# Claude Code CLI (native installer). Auth is provided at runtime via
+# CLAUDE_CODE_OAUTH_TOKEN in the container env, not baked in here.
+RUN curl -fsSL https://claude.ai/install.sh | bash
+ENV PATH="/root/.local/bin:${PATH}"
 
 WORKDIR /work
 CMD ["sleep", "infinity"]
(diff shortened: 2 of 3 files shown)


rjwalters:
❌ **Changes Requested**

Reviewed the three files by close reading of the diff plus the surrounding `remote.md` context. The design is sound and I can see the behavior was live-verified on a real t3.small (account rotation + gh label writes) — but one committed artifact contradicts itself and should be fixed before merge.

## Blocking: stale absolute "never copy credentials to the VM" wording in step 5

`commands/repo/remote.md:316-319` (step 5, "sync uncommitted work") still carries the pre-#11 absolute form:

> **Never copy `.env`, keys, or anything credential-like to the VM** — ... The VM authenticates to the forge on its own (e.g. `gh auth login` there), not with your local secrets.

This now directly contradicts the new **step 6a** (`:333-395`):
- "anything credential-like" includes the Claude pool `*.token` files that 6a **does** copy to `~/<repo-name>/.loom/tokens/`.
- "The VM authenticates to the forge on its own ... not with your local secrets" is precisely the interactive-login path that 6a **replaces** by injecting `GH_TOKEN` inline over SSH.

Every other cross-reference was correctly re-scoped to *provisioning* creds (intro `:18-19`, Config `:116-117`, Safety Rule 3 `:457-461`) — this step-5 note is the one spot left in its old absolute form. Because the runbook is executed top-to-bottom, an agent reading step 5 ("don't copy credentials; the VM auths on its own") can reasonably skip or be confused by step 6a's auth wiring.

**Suggested fix** — scope it to provisioning creds and defer dev-session auth to 6a, e.g.:

> **Don't copy the provisioning `.env` or cloud/SSH keys to the VM** — the `.gitignore` filter already excludes a gitignored `.env`; double-check it's excluded and call it out. Dev-session auth (gh token + Claude pool) is wired separately and by design in step 6a — provisioning credentials never leave the local host.

## Non-blocking nit to fold into the same pass: rsync `--chmod` portability

`commands/repo/remote.md:384` hardcodes `rsync -az --chmod=D700,F600 ...`. `--chmod` is a GNU-rsync feature; the macOS system rsync (BSD/openrsync) doesn't support it, so this exact invocation fails on an unpatched macOS operator box. Since the doc already says the files must land at `chmod 600` / dir `700`, either add a one-line note that `--chmod` is GNU-only (set perms via a follow-up `ssh ... chmod` on macOS), or drop `--chmod` and set perms explicitly after the transfer. Minor, but worth doing while the file is open.

## Verified good (no action needed)

- **Dockerfile**: valid syntax; `ubuntu:24.04` + `DEBIAN_FRONTEND=noninteractive`; installs gh from the official apt repo (keyring + `signed-by` source) and Claude Code via the native `claude.ai/install.sh` installer; `ENV PATH="/root/.local/bin:${PATH}"`; no secrets baked in; `CMD ["sleep","infinity"]`. Nothing would break `docker build`.
- **.gitignore**: ignoring `.env` is correct and consistent with the per-repo config model.
- **Pool resolution** prose (shared registry + `~/.config/repo/tokens/`, repo `.env` `ACCOUNT_*` + `.loom/tokens/` wins) is coherent and matches the #10 two-layer model; account 1 → default `CLAUDE_CODE_OAUTH_TOKEN`.
- **Token-files-as-files** rationale (Loom rotates by globbing files, not the registry) is correct.
- **gh clone-path caveat** present (6a point 2: `gh` infers repo from `.git`; rsync-only tree needs `-R <owner>/<repo>`).
- No leftover single-token (`REPO_REMOTE_CLAUDE_CODE_TOKEN`) references — grep is clean.
- No `Closes #N` expected here (direct feature work, not issue-tracked) — not faulted.

Merge state is CLEAN and no CI is configured on this repo, so the only gate is the step-5/step-6a consistency fix above. Flip back to `loom:review-requested` once addressed and I'll re-check.

rjwalters:
✅ **Approved!** (Second pass — both requested fixes verified resolved)

Re-reviewed the full `gh pr diff 11`. Both changes-requested items from the first pass are correctly applied, and nothing regressed.

**1. Step 5 / step 6a self-contradiction — RESOLVED.**
The old absolute wording is gone (grep confirms `authenticates to the forge on its own` and `not with your local secrets` no longer appear anywhere in `commands/repo/remote.md`). Step 5 is now re-scoped to the **provisioning-credential class** ("Never copy the *provisioning* `.env` or cloud keys to the VM … This is the provisioning-credential class (Safety Rule 3)") and explicitly points at step 6a for opt-in dev-session auth that *is* placed on the VM by design. No remaining contradiction with 6a.

**2. rsync `--chmod` portability — RESOLVED.**
The step-6a rsync invocation no longer carries `--chmod` (`rsync -az -e "ssh -i \$REPO_REMOTE_SSH_KEY" …`). Perms are now set in a follow-up `ssh … 'chmod 700 …/.loom/tokens && chmod 600 …/*.token'`. The only remaining `--chmod` occurrence is the explanatory NOTE that it is GNU-rsync-only and fails on macOS's system rsync. Correct fix.

**Rest of first-pass review re-confirmed:**
- Dockerfile valid (Ubuntu 24.04, toolchain + gh + Claude Code CLI, auth injected at runtime, `CMD ["sleep","infinity"]`).
- Two-secret-class framing consistent across intro ("Two classes of secret"), the `.env` Config block (provisioning vs dev-session auth sections), step 6a, and Safety Rule 3 ("Two secret classes, handled differently").
- Pool-resolution prose (shared vs repo, repo wins) reads consistently in both the Config note and step 6a.1.
- `.gitignore` adds `.env`.
- No single-token / single-account leftovers.

CI: no GitHub Actions workflows configured; `pnpm check:ci` is a passing stub. `mergeStateStatus` is CLEAN / MERGEABLE.

PR #16, opened 2026-07-17
docs(install): describe apply-by-default hygiene behavior in CLAUDE.md block

### LINKED ISSUE #12: Installer's CLAUDE.md block still says hygiene commands are 'report-first' — contradicts the 0.4.0 apply-by-default change
## Original Issue

Observed installing v0.4.1 (88188e4) into a consumer repo.

The 0.4.0 changelog declares a behavior change: fix-capable hygiene commands now **apply safe fixes by default** and `--ask` restores the old review-and-confirm flow. But the block `install.sh` writes into the consumer's `CLAUDE.md` still says:

> Hygiene commands are report-first: they present findings and wait before changing anything.

Since that block is loaded into every AI session's context in the consumer repo, agents will describe (and possibly enforce) the old confirm-first behavior, contradicting what the commands actually do. The template text in `install.sh` needs updating to match the 0.4.0 contract, e.g. "apply safe, reversible fixes by default and report each change; run with --ask to review first; irreversible removals stay behind --prune."

---

### Root Cause

`install.sh` writes a "REPO-SKILLS" pointer block into the consumer repo's
`CLAUDE.md` (between `<!-- BEGIN REPO-SKILLS -->` / `<!-- END REPO-SKILLS -->`
markers, see `install.sh:41-42`). That block is generated at
`install.sh:251-260` and one line is stale:

```bash
# install.sh:257
  echo "guide. Hygiene commands are report-first: they present findings and wait before changing"
```

This describes the pre-0.4.0 behavior. The 0.4.0 CHANGELOG entry (`CHANGELOG.md:35-41`)
documents the actual current contract:

> **Behavior change — apply safe fixes by default.** Fix-capable hygiene
> commands now apply their safe, reversible fixes (doc/link/gitignore edits,
> `tidy`'s regenerable SAFE clutter) automatically and report each change,
> instead of only reporting. Run any of them with `--ask` to restore the
> old review-and-confirm flow. Irreversible removals stay gated behind an
> explicit opt-in (`--prune`); `tidy`'s old `--apply` is now the default (kept
> as an alias).

Because this block is loaded into every AI agent's context in every consumer
repo that has run `install.sh`, the stale line actively misinforms agents
about what the hygiene commands will do — they may explain (or behave as if)
findings require manual confirmation, when in fact safe/reversible fixes are
applied automatically and only destructive removals require `--prune`.

### Scope confirmation (verified, not just a starting point)

Ran a repo-wide search for the same wording pattern to confirm this is an
isolated defect and not one of several stale copies:

```
git ls-files | xargs grep -n "report-first"
```

Results:
- `install.sh:257` — **stale**, this is the bug (unconditionally says
  hygiene commands are report-first, no mention of the 0.4.0 default).
- `commands/repo/all.md:39-40` — **not stale**, correctly scoped: "Under
  `--ask`, every stage reverts to report-first" — this accurately describes
  the *opt-in* confirm-first flow, consistent with the 0.4.0 contract.
- `CHANGELOG.md:72` — **not stale**, historical 0.2.0 changelog entry
  describing behavior at that point in time; changelogs are not living docs
  and should not be edited.

So `install.sh:257` is the only file requiring a change.

### Implementation Guidance

Update the echo statement at `install.sh:257` to describe the current
contract instead of the old one. Suggested replacement text (keep the
existing line-wrapping style used by the surrounding `echo` calls at
lines 254-258):

```bash
  echo "guide. Hygiene commands apply safe, reversible fixes by default and report"
  echo "each change; run with --ask to review first. Irreversible removals stay"
  echo "behind --prune. Managed by `install.sh` — edit outside the markers only."
```

(This merges into the existing block structure — adjust wrapping/line count
as needed to match `install.sh`'s existing 80-col-ish echo style; the exact
line breaks are a style choice, the content is what matters.)

Because this text is baked into every previously-installed consumer
`CLAUDE.md`, also note in the PR description that consumers must re-run
`install.sh` (or `/repo:update-tools`, if that's the intended upgrade path)
to pick up the corrected block — this is a doc-only fix, no version bump
required unless project convention says otherwise.

### Acceptance Criteria

- [ ] `install.sh`'s generated CLAUDE.md block no longer contains the phrase
      "report-first" as an unconditional description of hygiene command
      behavior.
- [ ] The block instead describes: safe/reversible fixes applied by default
      and reported; `--ask` restores confirm-first; `--prune` gates
      irreversible removals — consistent with the 0.4.0 CHANGELOG entry.
- [ ] Re-running `install.sh` against a target repo that already has the
      REPO-SKILLS block (the "Updated REPO-SKILLS block" path at
      `install.sh:262-270`) correctly replaces the stale line with the new
      text (marker-based replacement, not just append).
- [ ] No other occurrence of the stale wording remains in `install.sh`.

### Affected Files

- `install.sh` — update the `echo` line(s) generating the REPO-SKILLS
  CLAUDE.md block (currently line 257) to match the 0.4.0 apply-by-default
  contract.

### Test Plan

- [ ] Manual verification: Run `./install.sh --dry-run /tmp/some-target`
      (or a real target) and inspect the generated block between
      `<!-- BEGIN REPO-SKILLS -->` / `<!-- END REPO-SKILLS -->` markers;
      confirm it no longer says commands are "report-first" and instead
      matches the 0.4.0 CHANGELOG wording.
- [ ] Marker-replace path: run `install.sh` twice against the same target
      (first install, then re-install) and confirm the second run replaces
      the stale block content in place (per the `awk`-based marker rewrite
      at `install.sh:262-270`), not just on a fresh install.
- [ ] Automated tests: `.loom/scripts/tests/test-install-pr-markers.sh`
      exercises the marker-replace path — run it to confirm no regression;
      no test currently pins the exact block wording, so none should break,
      but consider (optional, not required for this fix) adding an assertion
      that the block does not contain "report-first" as a follow-up guard
      against future drift.
- [ ] Edge cases: verify the fresh-append path (`install.sh:272`, no
      existing marker block) also produces the corrected text.

rjwalters:
## Summary

The REPO-SKILLS pointer block that `install.sh` writes into a consumer repo's `CLAUDE.md` described hygiene commands as "report-first: they present findings and wait before changing anything." That is pre-0.4.0 behavior and directly contradicts the 0.4.0 contract in `CHANGELOG.md:35-41`. Because the block is loaded into every AI session's context in every consumer repo, the stale line actively misinformed agents about what the hygiene commands do.

This updates the generated block text to match the 0.4.0 apply-by-default contract.

## Changes

- `install.sh` (block generation at lines 253-259): replaced the "report-first" line with text describing the current contract — safe, reversible fixes applied by default and reported; `--ask` restores review-and-confirm; `--prune` gates irreversible removals. The trailing "Managed by \`install.sh\` — edit outside the markers only." note is preserved.

Both install paths emit the corrected text from the same generated block: the fresh-append path (line 272) and the `awk` marker-replace path (lines 262-270).

| Generated block no longer contains "report-first" as an unconditional description | Done | `grep -c report-first` on generated `CLAUDE.md` returns 0 |
| Block describes default apply + report, `--ask` confirm-first, `--prune` for irreversible removals | Done | Confirmed in generated block output |
| Re-run (marker-replace path) replaces stale line in place, not append | Done | Ran `install.sh --yes` twice against a temp target; BEGIN marker count stays 1 and block content is replaced |
| No other stale occurrence remains in `install.sh` | Done | `grep -n report-first install.sh` returns no matches |

## Test Plan

- Ran `./install.sh --yes <temp-target>` against a fresh git repo (fresh-append path): block contains the corrected text.
- Re-ran `./install.sh --yes <temp-target>` (marker-replace path): block is replaced in place, single BEGIN marker, no `report-first`.
- `.loom/scripts/tests/test-install-pr-markers.sh` is a Loom-shipped test that references `scripts/install/create-pr.sh` (absent from this repo); it errors independent of this change — a pre-existing, non-applicable test, not a regression.

## Upgrade note

This text is baked into every previously-installed consumer `CLAUDE.md`. Consumers must re-run `install.sh` (or `/repo:update-tools`) to pick up the corrected block. Doc-only fix; no version bump.

Closes #12


rjwalters:
✅ **Approved!**

Small, focused, and correct doc fix. Verified against the 0.4.0 contract in `CHANGELOG.md:35-41`.

**Correctness / accuracy**
- The replacement text in the generated block (`install.sh:257-259`) faithfully mirrors the 0.4.0 apply-by-default contract: safe, reversible fixes applied by default and reported; `--ask` restores review-and-confirm; `--prune` gates irreversible removals. This matches CHANGELOG lines 35-41 exactly.
- The stale unconditional "report-first: they present findings and wait before changing anything" line is gone.

**Both install paths covered**
- The block is generated once into `$BLOCK_FILE`, then consumed by both paths: the `awk` marker-replace path (`install.sh:263-271`) and the fresh-append path (`install.sh:273`). Both therefore emit the corrected text — confirmed.

**No regressions**
- `grep report-first install.sh` on the PR head returns no matches (exit 1).
- `bash -n install.sh` passes; the trailing "Managed by \`install.sh\` — edit outside the markers only." note and em-dash are preserved.
- Two `shellcheck` SC2094 info notes (lines 217, 273) are pre-existing and on unchanged lines — not introduced by this PR.

**Meta**
- Merge state CLEAN; PR body uses "Closes #12". No `.github/workflows` in this repo, so "no checks reported" is expected rather than a CI failure.

Nice catch on stale context that every consumer session was loading.

PR #17, opened 2026-07-17
fix: move machine-local source path + timestamp out of tracked install metadata

### LINKED ISSUE #14: install-metadata.json commits a machine-local absolute source path
## Summary

The installer writes `.claude/skills/repo/install-metadata.json` with an absolute `source` field pointing at the installer's checkout on the machine that ran it. Since this file lands in the consumer repo's tracked `.claude/` tree, it gets committed — leaking a machine-local path into shared history. The `installed_at` timestamp is similarly machine/run-specific, making two installs of the identical version+commit+skill-set produce different tracked bytes.

## Example

Installing from `~/GitHub/repo` into a consumer repo produced:

```json
{
  "version": "0.4.1",
  "commit": "88188e4",
  "installed_at": "2026-07-17T16:27:16Z",
  "dev": false,
  "source": "/Users/rwalters/GitHub/repo",
  "commands": ["all","audit","branches", ...]
}
```

`source` (`/Users/rwalters/GitHub/repo`) is specific to the installing machine and has no meaning in the committed repo — a teammate cloning it sees a path that doesn't exist for them.

## Root cause (exact code)

`install.sh` (repo root), step 3 — "Install metadata" — writes the JSON directly to the target's tracked path:

```bash
# install.sh:197-208
# 3. Install metadata
{
  echo "{"
  echo "  \"version\": \"$VERSION\","
  echo "  \"commit\": \"$COMMIT\","
  echo "  \"installed_at\": \"$(date -u +%Y-%m-%dT%H:%M:%SZ)\","
  echo "  \"dev\": $DEV,"
  echo "  \"source\": \"$SOURCE_ROOT\","
  echo "  \"commands\": [$(echo "$COMMANDS" | sed 's/.*/"&"/' | paste -sd, -)]"
  echo "}"
} >"$TARGET/.claude/skills/repo/install-metadata.json"
success "Wrote install-metadata.json"
```

`SOURCE_ROOT` (install.sh:36) is `$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)` — always an absolute, machine-local path to wherever the installer script lives when invoked.

Unlike dev-mode symlink installs — which explicitly add `.claude/` to the target's `.gitignore` (install.sh:210-218, "the symlinked install is machine-local ... must not be committed") — the normal (non-dev) install path has no such guard, so `install-metadata.json`, including its `source`/`installed_at` fields, is committed by default whenever the target's `.claude/` tree isn't already gitignored.

### `source` is not decorative — it's consumed by `/repo:update-tools`

`commands/repo/update-tools.md` (the `/repo:update-tools` skill) explicitly depends on this field:

> Key names vary by tool (`version` vs `loom_version` / `anvil_version`, `source` vs `loom_source` / `anvil_source`) — read whichever variant is present. Each file gives: installed version, installed commit, install date, and **the path of the local source clone it was installed from**.
>
> "For each tool, prefer **the local source clone recorded in the metadata** ..." (falls back to `gh api .../tags` / releases if the source clone no longer exists.)

So simply dropping `source` (fix option 1 from the original report) works but degrades `/repo:update-tools` to its GitHub-API fallback path on every run, even for the common case where the local clone that was used to install still exists on the same machine.

### Existing precedent in this codebase for the split

This repo already solves the identical problem for Loom's own self-install, via `.loom/loom-source-path` — a separate, plain-text, **gitignored** sidecar file (see `.gitignore`: `.loom/loom-source-path`) that stores exactly this kind of machine-local absolute path, read by `.loom/scripts/lib/loom-tools.sh:45-53`. The committed `.loom/install-metadata.json` counterpart also currently embeds a `loom_source` absolute path directly (`.loom/scripts/lib/loom-tools.sh:56-65` reads it as a *fallback* when the sidecar is missing) — i.e. the same anti-pattern this issue is about exists in Loom's own installer too, just outside this issue's scope (that installer lives in the separate `rjwalters/loom` repo, not here). Worth a follow-up issue there, but not part of this fix.

## Recommended fix: Option 3 (split tracked vs. machine-local fields), following the `loom-source-path` precedent

- **Tracked** `$TARGET/.claude/skills/repo/install-metadata.json` keeps only fields that are identical for any machine installing the same version/commit/skill-set: `version`, `commit`, `dev`, `commands`. This makes repeat installs of the same release byte-reproducible and removes the path/timestamp leak from history.
- **Untracked sidecar**, gitignored, holds the machine-local fields: `source` (absolute path to `SOURCE_ROOT`) and `installed_at`. Suggested location/format to match the existing convention: `$TARGET/.claude/skills/repo/.install-local.json` (JSON, since `update-tools.md` already expects to parse a JSON file for `source`/date — a plain-text file like `.loom/loom-source-path` would require `update-tools.md`'s two consumers, the doc-described discovery step, to special-case one more format).
- `install.sh` must add this sidecar path to `$TARGET/.gitignore` unconditionally (not just in dev mode) — mirror the existing `install.sh:215-218` block, e.g. append `.claude/skills/repo/.install-local.json` (or a broader `.claude/skills/*/install-local.json` if other tools in the family adopt the same split) with a comment explaining why.
- Update `commands/repo/update-tools.md`'s "Discover installed tools" step to also read `.claude/skills/repo/.install-local.json` (or `.claude/skills/*/.install-local.json`) for the `source` path, falling back to "local source unknown → use GitHub" when the sidecar is absent (e.g. a fresh clone on a different machine, or an older install predating this change) — this must not be treated as an error.
- Backward compatibility: older, already-committed `install-metadata.json` files (pre-fix) still contain `source`/`installed_at` inline. `update-tools.md` should keep reading `source` from the legacy location too if the new sidecar isn't present, so existing consumer-repo installs don't silently lose the "prefer local source clone" behavior until they re-run `install.sh`.

Why not Option 1 (drop `source` entirely) or Option 2 (relative path): Option 1 is simplest but silently degrades `/repo:update-tools`'s fast path on every affected repo. Option 2 is weaker per the reporter's own note — the installer source usually isn't inside/near the consumer repo, so a relative path is frequently meaningless or `../../../..`-style noise. Option 3 keeps the tracked file reproducible while preserving full `/repo:update-tools` functionality on the machine where the install actually happened, and reuses a pattern already established in this codebase.

## Acceptance Criteria

- [ ] `install.sh`'s non-dev install path no longer writes `source` or `installed_at` into the **tracked** `$TARGET/.claude/skills/repo/install-metadata.json`. That file contains only `version`, `commit`, `dev`, `commands`.
- [ ] `install.sh` writes `source` and `installed_at` to a new, gitignored sidecar file (e.g. `$TARGET/.claude/skills/repo/.install-local.json`).
- [ ] `install.sh` ensures the sidecar path is present in `$TARGET/.gitignore` on every (not just dev-mode) install, adding it if missing, using the same idempotent append pattern already used for `.claude/` in dev mode (install.sh:215-218).
- [ ] `--dry-run` output (install.sh:150-164) reflects both the tracked metadata file and the new gitignored sidecar as things that "would" be written.
- [ ] Running `./install.sh -y <target>` twice from two different `SOURCE_ROOT` checkouts of the *same* version+commit+skill selection produces byte-identical tracked `install-metadata.json` content (only `commands` ordering was already deterministic via `sort -u`; the new fix additionally removes `source`/`installed_at` as a source of diff).
- [ ] `commands/repo/update-tools.md`'s "Discover installed tools" step is updated to read `source` from the new sidecar (`.claude/skills/repo/.install-local.json` or equivalent) first, falling back to reading legacy inline `source` from `install-metadata.json` for pre-fix installs, and finally to "source clone unknown → check GitHub" when neither is present — this fallback chain must not be treated as an error condition anywhere in the doc.
- [ ] `commands/repo/help.md`'s reference to `cat .claude/skills/repo/install-metadata.json` (help.md:28) still works and is not broken by the split (it only needs `version`/`commands`/`dev`, all of which remain in the tracked file).
- [ ] No regression to dev-mode installs (`--dev`): dev-mode already skips the tracked metadata's problematic path via full-tree `.claude/` gitignoring; verify the new sidecar logic doesn't conflict with or duplicate that existing dev-mode gitignore block.

## Test Plan

1. `./install.sh -y /tmp/consumer-repo` (non-dev) into a fresh git repo; inspect `git status` in the target — confirm `install-metadata.json` is tracked/stageable but contains no absolute paths or timestamps, and the new sidecar file is untracked (shows under `.gitignore`d, not in `git status --short`).
2. `git diff` the tracked `install-metadata.json` between two installs run from differently-named checkout directories of this repo at the same commit — expect no diff.
3. `./install.sh --dry-run -y /tmp/consumer-repo` — confirm the dry-run listing includes the new sidecar path alongside the existing `install-metadata.json` line.
4. `./install.sh --dev -y .` (dogfood/dev mode) — confirm no behavior change (still exits early per install.sh:210-221, `.claude/` gitignored wholesale).
5. Manually run through `/repo:update-tools`'s discovery step against both a freshly-installed target (has sidecar) and a pre-fix target with only legacy inline `source` in `install-metadata.json` — confirm both resolve a local source path, and a target with neither resolves to the documented GitHub-fallback path without erroring.

## Affected Files

- `install.sh` (installer: metadata write, gitignore management, dry-run listing)
- `commands/repo/update-tools.md` (discovery step must read the new sidecar + legacy fallback)
- `commands/repo/help.md` (verify no regression to its `install-metadata.json` read)

rjwalters:
## Summary

`install.sh` wrote an absolute `source` path (the installer's checkout on the installing machine) and a run-specific `installed_at` timestamp into the **tracked** `.claude/skills/repo/install-metadata.json`. Because that file lands in the consumer's tracked `.claude/` tree, it got committed — leaking a machine-local path into shared history and making two installs of the same version+commit+skill-set produce different tracked bytes.

`source` is not decorative: `/repo:update-tools` reads it to prefer the local source clone over the GitHub API. So it is split out rather than dropped, following the existing `.loom/loom-source-path` gitignored-sidecar precedent already used for the identical Loom-self-install problem.

## Changes

- **`install.sh`**: tracked `install-metadata.json` now contains only `version`, `commit`, `dev`, `commands`. `source` + `installed_at` move to a new gitignored sidecar `.claude/skills/repo/.install-local.json`. The installer adds that path to the target `.gitignore` on every (not just dev-mode) install — idempotently, and only when not already covered by a broader ignore (`git check-ignore` guard). `--dry-run` lists the sidecar and its gitignore entry.
- **`commands/repo/update-tools.md`**: discovery step now resolves `source` from the sidecar first, falls back to legacy inline `source` in `install-metadata.json` for pre-fix installs, then to "unknown → GitHub" — none treated as an error.
- **`commands/repo/help.md`**: reads the install date from the sidecar (or legacy inline) and omits `(installed <date>)` gracefully when neither is present. (The curator's AC assumed help.md did not use the date; the doc's overview template does, so it is handled here.)

| Non-dev tracked metadata has no `source`/`installed_at` | ✅ | Installed into fresh repo; tracked file = version/commit/dev/commands only |
| Sidecar holds `source`/`installed_at`, gitignored | ✅ | Sidecar written; absent from `git status --short` after `git add -A` |
| `.gitignore` gets sidecar entry on every install | ✅ | Entry added in non-dev; dev-mode `.claude/` wholesale ignore covers it |
| `--dry-run` reflects sidecar + gitignore | ✅ | Dry-run lists `.install-local.json` and the gitignore entry |
| Two installs from differently-named checkouts → byte-identical tracked metadata | ✅ | `diff` of the two tracked files: identical |
| `update-tools.md` sidecar → legacy → GitHub fallback, none an error | ✅ | Discovery step rewritten with explicit 3-step chain |
| `help.md` unbroken by the split | ✅ | Reads version/commands/dev from tracked file; date from sidecar/legacy, omitted if absent |
| No dev-mode regression | ✅ | `--dev` still exits early, `.claude/` gitignored wholesale; sidecar covered, no duplicate entry; metadata + sidecar remain real files (not symlinks) |

## Test Plan

Ran locally against fresh `git init` targets:
1. Non-dev install → tracked metadata clean, sidecar gitignored (not in `git status`).
2. Byte-identical tracked metadata across two differently-named source checkouts at the same commit.
3. `--dry-run` lists sidecar + gitignore entry.
4. Idempotent re-install → no duplicate `.gitignore` line.
5. `--dev` install → `.claude/` wholesale ignore, sidecar covered, real files, no redundant entry.
6. Destination-already-gitignored install → no redundant sidecar line (check-ignore guard).

Closes #14

### DIFF
--- a/commands/repo/help.md
+++ b/commands/repo/help.md
@@ -26,11 +26,20 @@ commands.
 ```bash
 ls .claude/commands/repo/*.md
 cat .claude/skills/repo/install-metadata.json
+cat .claude/skills/repo/.install-local.json 2>/dev/null   # machine-local, gitignored
 ```
 
 Build the command list from the files present. For each, pull the one-line
 `description:` from its frontmatter. Do not list commands that are not
 installed here.
+
+`install-metadata.json` (tracked) carries `version`, `commands`, and `dev`. The
+install date lives in the gitignored `.install-local.json` sidecar
+(`installed_at`), so it is present only on the machine that ran the install;
+older installs still keep `installed_at` inline in `install-metadata.json`. Use
+whichever is present for the date below, and simply omit "(installed <date>)"
+when neither has it (e.g. a fresh clone on another machine) — its absence is
+expected, not an error.
 
 ### 2. Present the overview
 
--- a/commands/repo/update-tools.md
+++ b/commands/repo/update-tools.md
@@ -38,8 +38,28 @@ ls .claude/skills/*/install-metadata.json 2>/dev/null
 
 Key names vary by tool (`version` vs `loom_version` / `anvil_version`, `source`
 vs `loom_source` / `anvil_source`) — read whichever variant is present. Each
-file gives: installed version, installed commit, install date, and the path of
-the local source clone it was installed from.
+file gives: installed version, installed commit, install date, and (for the
+"prefer local source clone" fast path) the path of the local source clone it
+was installed from.
+
+**Locating the local source path (`source`).** The absolute source path and
+install timestamp are machine-local — they mean nothing in another clone — so
+newer installers keep them out of the tracked metadata file and write them to a
+gitignored sidecar instead. Resolve `source` in this order, and treat every
+step failing as "source clone unknown" rather than an error:
+
+1. **Sidecar first.** For Repo Skills, read
+   `.claude/skills/repo/.install-local.json` (generally
+   `.claude/skills/*/.install-local.json`); it holds `source` and
+   `installed_at`. Loom uses the plain-text `.loom/loom-source-path` sidecar for
+   the same purpose. A sidecar is gitignored, so it is present only on the
+   machine that ran the install — a fresh clone elsewhere legitimately has none.
+2. **Legacy inline fallback.** Older (pre-split) installs still embed `source` /
+   `installed_at` directly in `install-metadata.json` — read them from there if
+   no sidecar exists, so existing installs keep their fast path.
+3. **Unknown → GitHub.** If neither yields a usable path, the local source clone
+   is simply unknown; skip to the GitHub check in step 2. This is normal (fresh
+   clone on a different machine), not a failure.
 
 Known family members: Loom (`.loom/`), Anvil (`.anvil/`), Repo Skills
 (`.claude/skills/repo/`), kicad-tools, and anything else that follows the same
--- a/install.sh
+++ b/install.sh
@@ -150,6 +150,7 @@ if [[ "$DRY_RUN" == true ]]; then
   echo "Would write:"
   echo "  $TARGET/.claude/skills/repo/SKILL.md"
   echo "  $TARGET/.claude/skills/repo/install-metadata.json"
+  echo "  $TARGET/.claude/skills/repo/.install-local.json (machine-local, gitignored)"
   while IFS= read -r cmd; do
     echo "  $TARGET/.claude/commands/repo/$cmd.md"
   done <<<"$COMMANDS"
@@ -159,6 +160,7 @@ if [[ "$DRY_RUN" == true ]]; then
     || git -C "$TARGET" check-ignore -q .claude/skills/repo 2>/dev/null; then
     echo "  $TARGET/CLAUDE.md (skipped — install destination is gitignored)"
   else
+    echo "  $TARGET/.gitignore (.claude/skills/repo/.install-local.json entry)"
     echo "  $TARGET/CLAUDE.md (marker-bounded REPO-SKILLS block)"
   fi
   exit 0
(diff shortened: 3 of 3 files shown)


rjwalters:
✅ **Approved!**

Thorough review of the tracked/machine-local metadata split. I checked out `feature/issue-14` and executed `install.sh` against fresh `git init` targets to verify every acceptance criterion empirically — not just by reading the diff.

## Verification (all executed)

**1. Tracked metadata is clean and byte-reproducible**
Non-dev install into `t1` produced:
```json
{ "version": "0.4.1", "commit": "8947e91", "dev": false, "commands": [...] }
```
No `source`, no `installed_at`, no absolute path. Installed from two differently-named checkouts (`srcA`, `srcB`) at the same commit → `diff` of the tracked `install-metadata.json` = IDENTICAL. The machine-local leak (#14) is gone.

**2. Sidecar holds machine-local fields and is gitignored**
`.install-local.json` carries `source` + `installed_at`. After `git add -A`, `git status --short` does NOT list it — correctly ignored via the `.gitignore` entry the installer appended.

**3. `.gitignore` handling is idempotent and correctly scoped**
- Non-dev install adds `.claude/skills/repo/.install-local.json` with an explanatory comment.
- Re-install → still a single entry (`grep -cxF` = 1); the `check-ignore` + `grep -qxF` guards prevent duplication.
- **Dev mode** (`--dev`): `.claude/` is ignored wholesale, the sidecar block is correctly skipped (no redundant line), and both `install-metadata.json` and `.install-local.json` are written as **real files, not symlinks**. Exit 0, no regression.
- **Destination-already-gitignored** (pre-existing `.claude/`): the `check-ignore -q` guard suppresses a redundant sidecar line — verified, entry count = 0.

**4. `--dry-run`** lists both the sidecar (`machine-local, gitignored`) and the `.gitignore` entry it would add.

**5. Backward-compatible doc fallbacks**
- `update-tools.md`: rewritten discovery step resolves `source` sidecar-first → legacy inline → "unknown → GitHub", each explicitly documented as *not* an error. Pre-fix installs keep their local-source fast path.
- `help.md`: reads the date sidecar-first with legacy inline fallback. The builder's noted deviation from the curator's AC — the overview template's `(installed <date>)` line — is handled soundly: the doc instructs omitting `(installed <date>)` when neither source has it, framing absence (fresh clone elsewhere) as expected rather than an error. Correct call; the curator's AC assumed help.md didn't use the date, but the template does.

Confirmed `help.md` and `update-tools.md` are the only consumers of this metadata, so no other code path breaks.

## Non-blocking note
`shellcheck` flags SC2094 (info) on the new `{ [[ -s "$GITIGNORE" ]] && echo ""; ...; } >>"$GITIGNORE"` block. This is not a real bug — `>>` appends without truncating, so the `-s` size test reads the pre-existing content, and my tests confirmed the blank-line separator is added only when the file is non-empty. The pattern is already established elsewhere in this same file (dev-mode block, CLAUDE.md block), so the new code is stylistically consistent. No action required.

Merge state CLEAN, no CI workflows configured on this repo, PR uses `Closes #14`. Solid, well-tested fix that reuses the existing `loom-source-path` sidecar precedent.

PR #26, opened 2026-07-22
docs(update-tools): commit + land tool updates on the default branch by default

### LINKED ISSUE #25: update-tools: land the update on the default branch by default (not just leave it uncommitted)
## Request

Make `/repo:update-tools` **commit the update and merge it into the default branch (`main`) by default**, rather than leaving the changed files uncommitted in the working tree.

## Current behavior

`update-tools` runs each tool's own installer (which writes to the working tree), then stops there. Per the command's tail and Safety Rule 4:

> After updating, re-read each metadata file … and show a summary of what changed (`git status --short`) so the user can review and commit the update.
>
> 4. **The update touches the working tree** — leave the changes uncommitted for the user to review

So today the human has to branch/commit/merge the install output by hand every time.

## Why change the default

- The other `/repo:*` hygiene commands apply safe, reversible fixes by default and report each change; a tool-version bump run through the tool's own idempotent installer is squarely in that "safe, reversible" class (it's re-runnable and version-controlled), so leaving it as a manual finish is an inconsistent exception.
- In practice the update is almost always wanted on `main` — the manual commit step is friction, not a safety gain, because the change is already confirmed before it's applied (Safety Rule 1).

## Design considerations (for the maintainer to decide)

- **Mechanism:** branch → commit → merge to `main` (fast-forward/squash) vs. commit directly on `main`. If the session is already on `main`, branching first matches the common "don't commit straight to main" convention.
- **Commit message:** auto-generate per tool, e.g. `chore(tooling): update <tool> <old>→<new>` (mirrors how this repo already records such bumps).
- **Push:** landing on the local `main` is reversible; **pushing to the remote is outward-facing** and should stay gated / opt-in, not part of the default.
- **Confirmation:** keep Safety Rule 1 (confirm installed→latest before applying); the merge happens after that existing gate.
- **Escape hatch:** a flag to preserve today's behavior (e.g. `--no-commit` / `--stage-only`) for anyone who wants to review the diff before it lands.
- Update Safety Rule 4 accordingly so it no longer contradicts the new default.

- [ ] By default, after a confirmed update, `update-tools` commits the installer's changes and lands them on `main` (with a clear, per-tool commit message).
- [ ] The default does **not** push to any remote without explicit opt-in.
- [ ] A documented flag restores the current "leave it uncommitted for review" behavior.
- [ ] Safety Rules updated to match; dirty pre-existing working tree is handled safely (don't fold unrelated changes into the update commit).

## Context

Surfaced while running the repo-skills hygiene commands on botho-project/botho (repo skills v0.4.3). Filed via `/repo:followups`-style routing.

rjwalters:
## Summary

`/repo:update-tools` previously ran each tool's installer and stopped, leaving
the changes uncommitted in the working tree for the human to branch/commit/merge
by hand. This changes the default so a **confirmed** update is committed and
landed on the default branch (`main`) automatically, matching the "safe,
reversible fixes applied by default" convention of the other `/repo:*` hygiene
commands. Pushing stays a separate, explicit action, and a `--no-commit` escape
hatch restores the old review-first behavior.

## Changes

- **Usage**: added the `--no-commit` flag and a note that confirmed updates are
  committed + landed on the default branch by default (never pushed).
- **New step 5 "Land the update (default)"**: after the existing confirmation
  gate (step 4), commit the installer's output and land it on the default branch:
  - Snapshots the working tree before/after the installer and stages **only** the
    paths the installer actually changed (`comm -13` of pre/post `git status`),
    so a pre-existing dirty tree is never folded into the update commit.
  - Commits per tool with a `chore(tooling): update <tool> <old>→<new>` message.
  - When on the default branch: commits on a short-lived branch and
    `git merge --ff-only`s it in (lands on `main` without a straight-to-main
    commit); when already on a feature branch: commits in place and reports where
    it landed rather than switching branches mid-session.
  - Reports the commit and reminds the user it has not been pushed.
- **Safety Rules**: rewrote Rule 4 (was "leave the changes uncommitted for the
  user to review") to describe the new commit-and-land default + `--no-commit`
  escape hatch, and added Rule 5 ("Never push").

Only `commands/repo/update-tools.md` is touched (its `.claude/…` counterpart is a
symlink to this source file). README's classification of `update-tools` as a
"confirm-first" command remains accurate and is unchanged.

| By default, after a confirmed update, commit the installer's changes and land them on `main` with a per-tool commit message | ✅ | Step 5 + Safety Rule 4; commit msg `chore(tooling): update <tool> <old>→<new>`; branch→commit→`--ff-only` merge to default (or in-place on a feature branch) |
| Default does not push without explicit opt-in | ✅ | Step 5 and Safety Rule 5 state it never pushes; no `git push` anywhere in the flow |
| A documented flag restores the "leave it uncommitted for review" behavior | ✅ | `--no-commit` (alias `--stage-only`) added to Usage, step 5, and Safety Rule 4 |
| Safety Rules updated; dirty pre-existing tree handled safely | ✅ | Rules 4–5 rewritten; step 1 of the landing sequence stages only the installer's own paths via pre/post snapshot diff, never `git add -A` |
| On a non-`main` branch behaves sensibly | ✅ | Branch check commits in place and reports the landing branch instead of switching branches mid-session |

## Test Plan

This command family is markdown agent-instruction files with no executable
harness for the command itself. Verified:
- Scope check: `git diff --stat` shows only `commands/repo/update-tools.md`.
- `npm run check:ci` passes; `npm test` (hook guard suite) passes 49/49 —
  confirms no unrelated breakage.
- Manual read-through of the new step 5 / Safety Rules against each acceptance
  criterion (table above).

Closes #25

### DIFF
--- a/commands/repo/update-tools.md
+++ b/commands/repo/update-tools.md
@@ -15,15 +15,22 @@ to update the stale ones.
 ## Usage
 
 ```
-/repo:update-tools             # Report, then offer updates
-/repo:update-tools --check     # Report only
-/repo:update-tools loom        # Only check/update one tool
+/repo:update-tools               # Report, then offer updates (commit + land on the default branch)
+/repo:update-tools --check       # Report only, never writes
+/repo:update-tools loom          # Only check/update one tool
+/repo:update-tools --no-commit   # Update the working tree but leave it uncommitted for review
 ```
 
 An update runs the tool's own installer (executing code from its source repo
 and rewriting `.claude/`), so unlike the safe-fix hygiene commands this one is
 **not** auto-applied — it reports and confirms before updating. `--check` is
 the report-only form.
+
+Once an update is confirmed, it is committed and landed on the default branch
+(`main`) by default — it does **not** push, and it never folds a pre-existing
+dirty working tree into the update commit. Pass `--no-commit` (alias
+`--stage-only`) to restore the old behavior of leaving the changes uncommitted
+for manual review. See step 5 and the Safety Rules for details.
 
 ## Steps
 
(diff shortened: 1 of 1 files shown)


rjwalters:
## ✅ Approved

Evaluating this PR per explicit user request (`/loom:sweep 25`).

Scope is exactly one file — `commands/repo/update-tools.md` (78 insertions, 8 deletions). Verified `.claude/commands/repo/update-tools.md` is a symlink to that source, so editing only the source is correct and the PR body's symlink note is accurate.

### Correctness of the new git-plumbing (step 5)
Walked the four edge cases the change has to get right:

- **Not on the default branch:** the `if [ "$CUR" = "$DEFAULT" ]` branch commits in place and prints `Landed the update on '$CUR' (not '$DEFAULT')…` instead of switching branches mid-session. Matches the PR's claim and the issue's "behaves sensibly on a non-main branch" criterion.
- **`git merge --ff-only` failing on a diverged main:** not reachable in the documented flow — the short-lived `tmp` branch is cut from the current `$DEFAULT` tip and adds a single commit while `$DEFAULT` stays put, so the fast-forward is guaranteed within the session. Correct use of `--ff-only` to avoid a straight-to-main commit.
- **Zero installer-touched paths:** guarded — empty `changed.txt` is treated as a no-op ("already current", skip the commit), so `git commit` is never invoked with nothing staged.
- **Pre-existing dirty tree:** isolated via the pre/post `git status --porcelain` snapshot + `comm -13`, staging only installer paths with `git add -- …` and explicitly never `git add -A`. This satisfies the issue's "don't fold unrelated changes into the update commit" requirement.

### Acceptance criteria
The PR body's table maps 1:1 onto issue #25's "Suggested acceptance criteria" (commit+land on main / no push by default / documented `--no-commit` escape hatch / Safety Rules updated + dirty-tree safety), plus a fifth row for the non-main-branch edge case from the curator's test plan. Safety Rule 4 was correctly rewritten (no longer contradicts the new default) and Rule 5 ("Never push") added. `Closes #25` present.

### Checks
- `npm run check:ci` — passes (repo has no CI checks configured; no GitHub Actions workflows, `gh pr checks` reports none).
- `npm test` — 49/49 hook-guard tests pass, confirming no unrelated breakage.
- `mergeStateStatus: CLEAN`, `mergeable: MERGEABLE`.

### Non-blocking notes (illustrative pseudo-code, no fix required)
- `changed.txt` is written into the repo cwd; it's captured after the `post` snapshot so it never enters the diff, and it's never staged — a harmless leftover temp file. Consider `mktemp` for it too if you revisit.
- `git add -- $(cat changed.txt)` word-splits, so paths with spaces would break — not a concern for the `.claude/…` install footprint, which has none.

Clean, well-scoped documentation change with sound, well-explained git plumbing. Approving.

Issue #7, opened 2026-07-16
repo:release — Phase 1.5 CHANGELOG gate assumes bracketed [x.y.z] headers, false-negatives on bracket-less format


rjwalters:
## Summary

Phase 1.5's CHANGELOG completeness gate greps for Keep-a-Changelog bracketed headers (`## [0.4.0]`), but this repo's `CHANGELOG.md` uses a bracket-less format (`## 0.4.0 (2026-07-15)`). Result: the gate reported `v0.4.0`, `v0.3.0`, `v0.2.0` as **MISSING** when the 0.4.0 entry is literally at the top of the file.

## Contradiction within the command

Phase 4 ("Draft the CHANGELOG") instructs: *"study its format and draft a new entry matching it."* So the command correctly adapts to the repo's real format when writing — but Phase 1.5's completeness check hardcodes the bracketed style when reading. The two phases disagree about what a CHANGELOG header looks like.

## Repro

```bash
# In a repo whose CHANGELOG uses "## 0.4.0 (date)" (no brackets):
/repo:release
# Phase 1.5 reports v0.4.0/v0.3.0/v0.2.0 as MISSING (false negative)
```

## Suggested fix

Make the gate's version-header match format-agnostic: accept both `## [x.y.z]` and `## x.y.z` (and an optional leading `v`), e.g. match `^##\s+v?\[?<version>\]?` rather than requiring brackets. Ideally infer the repo's header style once (as Phase 4 does) and reuse it for the completeness check.

## Severity: low

False-negative only — it over-reports gaps, so worst case is a spurious backfill prompt; it won't skip a real gap. Cosmetic/annoyance. Discovered cutting v0.4.1 on 2026-07-16.

---

### Affected Files
- `commands/repo/release.md` — Phase 1.5 "CHANGELOG completeness gate" (verified present on `origin/main`)

### Exact location to change
Phase 1.5 lives at **`commands/repo/release.md` lines 42–57**, and the version-header check is the prose/comment block at **lines 48–52**:

```bash
[ -f CHANGELOG.md ] || echo "(no CHANGELOG.md — skipping gate)"
# For each of the last ~5 tags (git tag --sort=-v:refname), check that
# CHANGELOG.md contains a header for its version (strip a leading 'v').
```

[...]

Issue #6, opened 2026-07-16
repo:release — version-tool detection prefers npm/package.json over VERSION even when they disagree


rjwalters:
## Summary

Cutting v0.4.1 via `/repo:release` surfaced a detection bug: Phase 2's version-tool detection tries `npm` (whenever `package.json` exists) **before** the plain `VERSION` file. In this repo the two disagree:

- `VERSION` = `0.4.0` — the real source of truth (every tag `v0.4.0` / `v0.3.0` / `v0.2.0` matches it)
- `package.json` = `1.0.0` — a vestigial placeholder, never maintained

Had the command run its detected tool (`npm version patch`), it would have bumped `package.json` 1.0.0 → 1.0.1 and tagged **v1.0.1** — wrong version, wrong file. The operator had to override to `version-file` semantics by hand.

## Why the drift gate doesn't catch it

Phase 2's "Drift gate (multi-file tools only)" explicitly treats `npm` as single-source / drift-free and skips the agreement check. But a repo can carry BOTH a `package.json` and a `VERSION` file that disagree — exactly this case — and npm-first detection silently picks the non-authoritative one.

## Repro

```bash
# In a repo with VERSION=0.4.0 and package.json version=1.0.0 and tags matching VERSION:
/repo:release
# Phase 2 reports: "npm (package.json present)"  ← wrong; VERSION is authoritative
```

## Suggested fix (pick one or combine)

1. When both `package.json` and a root `VERSION` file exist, cross-check them; if they disagree, do NOT silently pick npm — surface the drift and ask which is authoritative (extend the drift gate to the multi-*source* case, not just multi-file-within-one-tool).
2. Prefer the file that agrees with the latest git tag as a tie-breaker.
3. Let the operator pin the tool/source (e.g. a `.repo-release.toml` or a CLAUDE.md hint).

## Related latent issue

`package.json` in this repo is stuck at `1.0.0` (placeholder). It should be reconciled — either dropped from version-bearing status or synced to the real line — so future detection isn't ambiguous. Deliberately left untouched during the v0.4.1 cut.

## Severity: medium

Not a breakage (operator caught it), but a naive run would have shipped a wrong-versioned tag. Discovered cutting v0.4.1 on 2026-07-16.

---

## Curator Enhancement

[...]

PR #3, opened 2026-07-15
docs(tidy): make SAFE an allowlist and add never-delete denylist for secrets/venvs

rjwalters:
## Summary

`/repo:tidy` inventoried deletion candidates with `git clean -ndX` (every gitignored file on disk) and defined the auto-deleted **SAFE** category as "gitignored build output and caches," gated only by two negative guards (not tracked, no source extension). Neither guard excludes `.env` (secrets) or `.venv/` (expensive to rebuild), so a faithful reading of the spec would sweep them into SAFE and `git clean -fdX` them. This reworks `commands/repo/tidy.md` so **gitignored ≠ safe to delete**.

## Changes

- **Step 2 (Categorize)**: SAFE is now an explicit **allowlist** of known build/cache/droppings patterns (`__pycache__/`, `*.pyc`, `.pytest_cache/`, `.mypy_cache/`, `.ruff_cache/`, `dist/`, `.turbo/`, `.astro/`, `htmlcov/`, `.coverage`, OS/editor droppings, merge/patch leftovers, empty dirs) — not "everything `git clean -ndX` lists minus exclusions."
- Added a **never-delete denylist**, checked first and overriding SAFE regardless of gitignore status: `.env`, `.env.*` (excluding `.env.example`/`.env.sample`), `*.pem`, `*.key`, `*.keystore`, `*.p12`, `*.pfx`, `id_rsa*`, and virtualenvs `.venv/`/`venv/`/`env/`. Denylist matches route to **ASK**, never KEEP.
- ASK now explicitly covers denylisted gitignored files **and** any gitignored file not on the SAFE allowlist (unrecognized → ASK).
- **Step 3 (Report)**: example now shows `.env` and `.venv/` under ASK with annotations.
- **Step 4 (Apply)**: default auto-delete is scoped to SAFE-allowlisted paths only; denylisted/unrecognized gitignored paths must never be passed to `git clean -fdX`.
- **Safety Rules**: added rule 6, "Gitignored ≠ safe to delete."

## Acceptance Criteria Verification

| Criterion | Status | Verification |
|-----------|--------|--------------|
| SAFE rewritten as an allowlist, not "everything `git clean -ndX` lists" | ✅ | Step 2 SAFE now enumerates explicit patterns; intro to step 2 states the allowlist principle |
[...]

PR #34, opened 2026-07-27
feat(hooks): surface a pending /repo:handoff note via a SessionStart hook

### LINKED ISSUE #32: install.sh: offer to wrap the user's shell `claude` command so a pending /repo:handoff note is surfaced at session start
## Original Issue

`/repo:handoff` writes `.claude/handoff.md` and adds a pointer line to the agent's `MEMORY.md`. That makes the note reliably findable **by Claude** — the memory index is read at session start.

It does nothing for **the human**. You quit, run `claude update`, relaunch, and there is no signal that the previous session left instructions. The one-shot contract ("read it, then delete it") also means a note that *doesn't* get absorbed lingers silently — and a stale handoff is explicitly called out in `handoff.md` as "a trap of its own." Nothing currently surfaces that staleness.

The gap is small and the fix is ~40 lines of shell, but it has to be **opt-in and careful**, because it is the one part of the install that touches a file outside the repo.

## Proposal

During install (and gated behind an explicit confirm, default **N**), offer to install a shell wrapper that prints a compact banner when a handoff note exists in the current repo.

Implemented and running in a consumer repo (2AMLogic/studio) as of 2026-07-27:

```zsh
_claude_handoff_banner() {
  local root note
  root="$(git rev-parse --show-toplevel 2>/dev/null)" || root="$PWD"
  note="$root/.claude/handoff.md"
  [[ -r "$note" ]] || return 0

  local now mtime age_h age_label
  now=$(date +%s)
  mtime=$(stat -f %m "$note" 2>/dev/null) || return 0   # BSD stat; GNU is -c %Y
  age_h=$(( (now - mtime) / 3600 ))
  if   (( age_h < 1 ));  then age_label="just now"
  elif (( age_h < 48 )); then age_label="${age_h}h old"
  else                        age_label="$(( age_h / 24 ))d old"
  fi

  local y=$'\033[33m' d=$'\033[2m' r=$'\033[0m' b=$'\033[1m'
  printf '%s┌─ %sHandoff note pending%s%s — %s (%s)%s\n' \
    "$y" "$b" "$r" "$y" "${note/#$HOME/~}" "$age_label" "$r"

  grep -E '^#{1,2} ' "$note" 2>/dev/null | head -9 | sed 's/^#* *//' | while IFS= read -r line; do
    printf '%s│%s  %s\n' "$y" "$r" "$line"
  done

  if (( age_h >= 168 )); then
    printf '%s└─ %sSTALE (>7d) — a handoff describes one moment; verify before trusting.%s\n' "$y" "$b" "$r"
  else
    printf '%s└─%s %sClaude reads it via MEMORY.md; delete note + pointer once absorbed.%s\n' "$y" "$r" "$d" "$r"
  fi
}

_claude_is_session() {
  case "$1" in
    update|doctor|mcp|config|install|migrate-installer|setup-token) return 1 ;;
    -v|--version|-h|--help) return 1 ;;
    *) return 0 ;;
  esac
}

claude() {
  _claude_is_session "$1" && _claude_handoff_banner
  command claude --dangerously-skip-permissions "$@"
}
```

Sample output:

```
┌─ Handoff note pending — ~/GitHub/studio/.claude/handoff.md (just now)
│  Handoff — 2026-07-27
│  1. In-flight — resolve before anything else
│  2. Decisions — settled, do not relitigate
│  3. Traps discovered empirically — these cost real time
│  4. Research pool — what is citable
│  5. Repo state
│  6. The precise next action
└─ Claude reads it via MEMORY.md; delete note + pointer once absorbed.
```

It prints section headers rather than the file — the real note was 7.5 KB, far too much to dump on every launch, but the headers convey the shape in eight lines.

## Design constraints found while building it

These are the parts that are easy to get wrong:

1. **The existing `claude` may be an alias carrying flags.** This user's was `alias claude='command claude --dangerously-skip-permissions'`. A naive append silently loses the flag, or the alias shadows the new function depending on load order. The wrapper must **parse the existing alias, preserve its flags, and remove the alias** — aliases and functions of the same name do not compose.
2. **Recursion.** The function body must call `command claude`, never `claude`.
3. **Gate on subcommands.** `claude update`, `claude mcp`, `claude --version` are not sessions; a banner there is noise. The `_claude_is_session` list above is the minimum.
4. **`stat` is not portable.** `-f %m` is BSD/macOS, `-c %Y` is GNU. Detect or fall back.
5. **Shell detection.** `$SHELL` / `$ZSH_VERSION` / `$BASH_VERSION`. The `local`/`(( ))`/`${var/#$HOME/~}` forms above are zsh+bash compatible, but fish needs a separate translation — probably better to *skip* fish with a clear message than to half-support it.
6. **Idempotency + uninstall.** Wrap in `# BEGIN REPO-SKILLS CLAUDE WRAPPER` / `# END` markers, same convention as the CLAUDE.md block, so re-install replaces in place and `uninstall.sh` can remove it. Today `uninstall.sh` has no reason to touch shell rc files; this would give it one.
7. **Back up the rc file** before editing it.

## Why gate it hard

`install.sh` is otherwise scrupulous about staying inside the target repo — the CLAUDE.md block is marker-bounded, `settings.json` is jq-merged rather than clobbered, and the destructive-guard wiring defers to an existing guard. Editing `~/.zshrc` is categorically different: it is global, it affects every repo and every shell, and it changes the behaviour of a command the user runs constantly.

Suggested: **not** part of the default flow. Either a `--shell-wrapper` opt-in flag, or a confirm defaulting to N with the diff shown first. And it should be a no-op under `--yes` unless the flag is explicit, so unattended installs never silently rewrite a shell rc.

## Alternative worth considering

A `SessionStart` hook in `.claude/settings.json` could inject the note into the session's context directly — repo-scoped, no shell edit, no portability surface. That is arguably the more correct mechanism. It would not show the human anything *before* Claude starts, which is part of the value here, but the two are complementary and the hook is strictly safer. Might be worth doing that first and treating the shell wrapper as the optional extra.

Filed from studio (canary consumer), where `/repo:handoff` was run end-to-end for the first time today.

---

### Decided scope: SessionStart hook only (shell wrapper is OUT OF SCOPE for this issue)

The issue's own "Alternative worth considering" section correctly identifies that the `SessionStart` hook is the safer, more architecturally-consistent mechanism, and explicitly floats deferring the shell wrapper. This curation pass makes that call explicit rather than leaving it open:

- **In scope for this issue**: a `SessionStart` hook, wired into the target repo's `.claude/settings.json` via the same jq-merge pattern `install.sh` already uses for the `PreToolUse` guard hook (`merge_settings_hook`). It reads `.claude/handoff.md` in the target repo (read-only) and, if present, emits `hookSpecificOutput.additionalContext` summarizing it (section headers + age, same compact style as the shell-wrapper banner in the original proposal). This never touches anything outside the target repo, so the "gate it hard" / opt-in-default-N constraint from the original proposal does not apply to it (see hard requirement below — it's satisfied vacuously because no file outside the repo is touched).
- **Out of scope for this issue**: the shell `claude` wrapper (editing `~/.zshrc`/`~/.bashrc`/etc.). All seven "design constraints found while building it" above are specific to that wrapper (alias parsing, recursion guard, subcommand gating, fish detection, rc backup, wrapper-specific idempotency markers) and remain unaddressed by the hook. If a human still wants the shell wrapper after this ships, file it as a **separate follow-up issue** — bundling both mechanisms into one PR would exceed a reasonable single-PR size (new hook script + tests + install/uninstall wiring for the hook, *plus* alias-parsing/rc-editing/fish-detection/backup logic for the wrapper is comfortably >8 files and >400 LOC combined).

This resolves the issue's own open question rather than leaving the Builder to choose.

### Root-cause / current-state verification (against `origin/main`)

Verified directly against the current codebase (not just the issue's assumptions):

- `install.sh` has **no** shell-rc-editing code path today, and **no** `SessionStart` hook wiring today (`grep -rn "SessionStart"` across the repo returns nothing). Both are genuinely new.
- `install.sh` already has a proven, idempotent, coexistence-aware pattern for JSON-merging a hook into a consumer's `.claude/settings.json` without clobbering other content: `merge_settings_hook()` (`install.sh:247-296`), which is careful to (a) refuse to touch malformed JSON, (b) detect its own hook already present and no-op, (c) detect a *different* tool's guard already wired for the same purpose and defer rather than double-register. The new `SessionStart` hook wiring should follow this same shape (new sibling function, e.g. `merge_settings_sessionstart_hook`), reusing the temp-file-and-mv write discipline (never redirect `jq` straight onto the file).
- `uninstall.sh` has a mirror-image `remove_settings_hook()` (`uninstall.sh:29-51`) that removes only the Repo-Skills-owned entry and prunes now-empty `hooks`/`PreToolUse` containers. A `remove_settings_sessionstart_hook()` counterpart is needed for symmetry, plus removal of the new hook script file (already covered for free by the existing `rm -rf "$TARGET/.claude/skills/repo"` in `uninstall.sh:85`, since the new script should live under that tree — see Affected Files).
- The existing destructive-command guard hook script (`hooks/repo/guard-destructive.sh`) establishes the input-handling convention to reuse: JSON on stdin, parsed with `jq -r '.cwd // empty'` etc., fail-open on any internal error (never a non-zero exit, never invalid output), errors logged rather than surfaced. The new hook should follow this convention: fail-open (silently emit nothing / exit 0) rather than ever breaking session start.
- `commands/repo/handoff.md` confirms the note path is `.claude/handoff.md` relative to the repo root, and that `/repo:handoff` already gitignores it and is responsible for the one-shot delete-after-absorb contract — the new hook only needs to *read* this path, never write or delete it.
- Existing test harness precedent: `hooks/repo/tests/run.sh` + `hooks/repo/tests/test-guard-destructive.sh` — a new hook should ship with an equivalent `hooks/repo/tests/test-session-start-handoff.sh` wired into `run.sh`.
- `install.sh --dry-run` currently enumerates every file/setting it would write (`install.sh:189-218`); the new hook's file + settings.json wiring line must be added to that enumeration so `--dry-run` stays accurate.

### Hard requirement (encode as acceptance criteria)

Nothing this issue implements may edit any file **outside** the target repo. There is no opt-in flag needed to satisfy this for the hook-only scope, because the hook script and its `.claude/settings.json` wiring are both entirely inside the target repo — same risk class as the existing `PreToolUse` guard hook, which is already wired unconditionally (no confirm prompt) during install. Per that precedent, the `SessionStart` hook is wired **unconditionally** as part of the normal (non-`--dev`) install flow — no new confirm/flag is required, and no new prompt should be added under `--yes`/non-interactive installs (it must behave identically whether `--yes` is passed or not, since it never risks unattended damage). This constraint remains binding for any *future* shell-wrapper issue, which by definition edits a file outside the repo (`~/.zshrc` etc.) and therefore MUST default to N / be strictly no-op under `--yes` unless an explicit opt-in flag is passed.

### Acceptance Criteria

- [ ] New hook script (see Affected Files) reads Claude Code's `SessionStart` JSON payload from stdin (fields include at least `.cwd` and `.source`) and resolves `.claude/handoff.md` relative to the repo root derived from `.cwd` (fall back gracefully, fail-open, if `.cwd` is missing/invalid — never break session start).
- [ ] Hook fires only for `source == "startup"` or `source == "resume"`; it is a no-op (silent, exit 0) for `source == "clear"` (a `/clear` isn't a process relaunch, and firing there would repeat the banner without the note having been touched).
- [ ] If `.claude/handoff.md` does not exist or is not readable, the hook exits 0 with no output (matches the shell-wrapper banner's original `[[ -r "$note" ]] || return 0` behavior).
- [ ] If the note exists, the hook emits `{"hookSpecificOutput": {"hookEventName": "SessionStart", "additionalContext": "..."}}` containing: the note's path, its age (portable stat handling — BSD `stat -f %m` vs GNU `stat -c %Y`, per design constraint #4 above, which still applies here since this script runs on the install target's machine), a staleness flag when age ≥ 7 days, and a compact section-header summary (`grep -E '^#{1,2} '`-style, headers only — never the full file body, per the original proposal's "far too much to dump on every launch" rationale).
- [ ] The hook is read-only: it never writes, deletes, or modifies `.claude/handoff.md` or any other file.
- [ ] The hook never exits non-zero and never emits malformed JSON on any error path (fail-open, consistent with `guard-destructive.sh`'s ERR-trap convention) — a hook crash must never block or corrupt session start.
- [ ] `install.sh` wires the hook into `.claude/settings.json` via a new function mirroring `merge_settings_hook()`'s idempotency (re-running install doesn't duplicate the entry) and coexistence-awareness (a pre-existing `SessionStart` hook entry with a *different* command is preserved, not clobbered).
- [ ] The hook is wired unconditionally in the default (non-`--dev`) install flow — no new confirm prompt, no new flag, identical behavior under `--yes` and interactively (see Hard Requirement above).
- [ ] `install.sh --dev` mode: hook script is symlinked like other skill files (consistent with existing `install_file()` dev-mode behavior); no `.claude/settings.json` wiring changes needed beyond what dev mode already skips for CLAUDE.md (verify against current dev-mode branch at `install.sh:371-381` — dev mode does NOT currently write `.claude/settings.json` at all is false, since `merge_settings_hook` runs unconditionally at `install.sh:365` before the dev-mode early-exit at `install.sh:371`; the new SessionStart wiring should run at the same point, before the dev-mode branch).
- [ ] `install.sh --dry-run` output lists the new hook script path and a line describing the `SessionStart` settings.json wiring (extend the existing enumeration at `install.sh:189-218`).
- [ ] `uninstall.sh` removes the `SessionStart` hook's settings.json entry via a new function mirroring `remove_settings_hook()`, pruning now-empty containers, and leaves any unrelated `SessionStart` hooks (hand-authored, or another tool's) untouched.
- [ ] New automated test file exercises: note present (fresh) → context emitted with correct fields; note present (stale ≥7d) → staleness flag set; note absent → silent no-op exit 0; `source == "clear"` → silent no-op even if note present; malformed/missing stdin JSON → fail-open, exit 0, no crash.
- [ ] Shell-wrapper design constraints #1 (alias parsing), #2 (recursion), #3 (subcommand gating), #5 (shell/fish detection), #6 (wrapper-specific idempotency markers + rc uninstall), #7 (rc backup) are explicitly OUT OF SCOPE for this issue — they apply only to the shell wrapper, which is deferred to a future issue (see Decided Scope above). Design constraint #4 (stat portability) DOES apply, since the hook script itself needs the note's mtime.

### Implementation Guidance

- Script location: `hooks/repo/session-start-handoff.sh`, following `hooks/repo/guard-destructive.sh`'s header/input-handling conventions (read stdin once, `jq -r '.field // empty'`, ERR trap that fail-opens and logs rather than surfaces).
- Portable mtime: try `stat -f %m "$note" 2>/dev/null` first (BSD/macOS), fall back to `stat -c %Y "$note" 2>/dev/null`; if both fail, fail-open (skip the banner rather than error).
- Reuse the compact section-header rendering approach from the original shell-wrapper proposal (`grep -E '^#{1,2} ' | head -9 | sed 's/^#* *//'`) when building the `additionalContext` string — plain text is fine; `additionalContext` does not need ANSI color codes since it is consumed by Claude/the transcript, not a raw terminal.
- `install.sh` change: add a new function alongside `merge_settings_hook()` (e.g. `merge_settings_sessionstart_hook()`) using the same `.matcher`/`.hooks` merge shape but for `.hooks.SessionStart` instead of `.hooks.PreToolUse` (note: `SessionStart` hook matchers key on `source`, e.g. `"startup"`/`"resume"`, rather than a tool matcher like `"Bash"` — confirm the exact matcher shape against current Claude Code hook documentation before implementing, since this repo has no existing `SessionStart` example to copy from). Call it in the same place as the existing `merge_settings_hook` call (`install.sh:365`), before the dev-mode branch.
- `uninstall.sh` change: add `remove_settings_sessionstart_hook()` mirroring `remove_settings_hook()`, called alongside the existing call (`uninstall.sh:91`).
- Test wiring: add `hooks/repo/tests/test-session-start-handoff.sh` and register it in `hooks/repo/tests/run.sh` next to the existing guard test.

### Affected Files

- `hooks/repo/session-start-handoff.sh` — NEW. The `SessionStart` hook script (read-only; emits `additionalContext` summarizing a pending handoff note).
- `hooks/repo/tests/test-session-start-handoff.sh` — NEW. Automated tests per Acceptance Criteria above.
- `hooks/repo/tests/run.sh` — MODIFY. Register the new test file (existing file confirmed present on `origin/main`).
- `install.sh` — MODIFY. Add `merge_settings_sessionstart_hook()` (or equivalent), call it alongside the existing `merge_settings_hook` call (~`install.sh:365`), install/chmod the new hook script (mirroring the `guard-destructive.sh` install block at `install.sh:360-364`), and extend the `--dry-run` enumeration (`install.sh:189-218`).
- `uninstall.sh` — MODIFY. Add `remove_settings_sessionstart_hook()` (or equivalent), call it alongside the existing `remove_settings_hook` call (~`uninstall.sh:91`), and extend the pre-uninstall "will remove" summary (`uninstall.sh:70-78`) to mention the new settings.json entry when present.

### Test Plan

- [ ] Manual verification: run `./install.sh --dry-run /path/to/test-repo` and confirm the new hook file + `SessionStart` wiring line appear in the output.
- [ ] Manual verification: run `./install.sh /path/to/test-repo`, create `.claude/handoff.md` in that repo with a couple of `##` headers, then manually pipe a synthetic `SessionStart` JSON payload (`{"cwd": "/path/to/test-repo", "source": "startup"}`) into `hooks/repo/session-start-handoff.sh` and confirm well-formed `hookSpecificOutput.additionalContext` JSON is emitted on stdout.
- [ ] Manual verification: repeat with no `.claude/handoff.md` present — confirm silent, exit 0, no output.
- [ ] Manual verification: repeat with `"source": "clear"` and the note present — confirm silent, exit 0, no output.
- [ ] Manual verification: run `./uninstall.sh -y /path/to/test-repo` and confirm the `SessionStart` entry is removed from `.claude/settings.json` (and unrelated hook entries, e.g. hand-authored ones, survive).
- [ ] Automated tests: `hooks/repo/tests/test-session-start-handoff.sh` covering fresh note, stale note (≥7d), missing note, `source == "clear"`, and malformed stdin — run via `hooks/repo/tests/run.sh`.
- [ ] Re-run existing `hooks/repo/tests/test-guard-destructive.sh` to confirm no regression from the shared `install.sh`/`uninstall.sh` edits (i.e. the `PreToolUse` guard wiring/removal still works unchanged alongside the new `SessionStart` wiring/removal).
- [ ] Idempotency: run `./install.sh -y /path/to/test-repo` twice in a row and confirm `.claude/settings.json`'s `SessionStart` entry is not duplicated.

rjwalters:
## Summary

`/repo:handoff` writes `.claude/handoff.md` and adds a pointer to the agent's memory index, but nothing *announces* the note at session start — and the command's one-shot "read it, then delete it" contract means a note that never gets absorbed lingers silently. `handoff.md` itself calls a stale handoff "a trap of its own"; nothing surfaced that staleness.

This implements the curated scope: a **`SessionStart` hook only**. The shell `claude` wrapper from the original proposal is deliberately **not** implemented (see Out of Scope).

## Changes

- **NEW `hooks/repo/session-start-handoff.sh`** — read-only `SessionStart` hook. Parses the payload from stdin (`.cwd`, `.source`), resolves the repo root via `git rev-parse --show-toplevel` (falling back to `cwd`), and emits `{"hookSpecificOutput":{"hookEventName":"SessionStart","additionalContext":"..."}}` with the note's path, age, a `STALE` flag past 7 days, and a **headers-only** summary (`^#{1,2} `, capped at 9 — never the body; the real note that motivated this was 7.5 KB). Portable mtime (BSD `stat -f %m` → GNU `stat -c %Y`). Follows `guard-destructive.sh`'s conventions: `ERR` trap, diagnostics to `../logs/hook-errors.log`, **always exit 0**, never malformed output.
- **NEW `hooks/repo/tests/test-session-start-handoff.sh`** — 79 cases.
- **`hooks/repo/tests/run.sh`** — delegates to the new suite so `pnpm test` covers both hooks.
- **`install.sh`** — installs + `chmod +x` the script alongside `guard-destructive.sh`; new `merge_settings_sessionstart_hook()` mirroring `merge_settings_hook()`; hook constants moved above the `--dry-run` enumeration so listing and doing read from the same values; `--dry-run` extended.
- **`uninstall.sh`** — new `remove_settings_sessionstart_hook()` mirroring `remove_settings_hook()`; "will remove" summary extended.

### On the `SessionStart` matcher shape

Verified against the current hook docs rather than assumed. `SessionStart` matchers key on the session **source** (`startup` / `resume` / `clear` / `compact` / `fork`), and the output envelope is `hookSpecificOutput.additionalContext`. Alternation (`"startup|resume"`) is documented for *tool* matchers but **not** for `SessionStart`, and a matcher that silently matched nothing would leave the hook inert with no error — so install wires **two literal entries** using only documented values. The script re-checks `.source` itself, so a hand-edited `settings.json` routing `clear` here still no-ops.

|---|---|---|
| Reads `SessionStart` stdin JSON (`.cwd`, `.source`), resolves `.claude/handoff.md` from repo root, fails open | ✅ | Cases: subdir cwd, non-git cwd fallback, nonexistent cwd, missing cwd |
| Fires only for `startup`/`resume`; silent no-op for `clear` | ✅ | Cases for `clear`, `compact`, `fork` — all exit 0, no output despite note present |
| Note absent/unreadable → exit 0, no output | ✅ | Cases: no note, `chmod 000` note, `handoff.md` as a directory |
| Emits `hookSpecificOutput` with path, age, staleness ≥7d, headers-only summary | ✅ | Cases assert `hookEventName == "SessionStart"`, path, `just now`/`5h old`/`3d old`/`8d old`, `STALE` at 8d but not 6d, headers 1–9 present, header 10 dropped, body prose absent, `###` not treated as a section |
| Hook is read-only | ✅ | Content `cksum` + mtime asserted unchanged across a run |
| Never exits non-zero, never malformed JSON | ✅ | Empty stdin, `not json at all {{{`, JSON array payload, missing fields — all exit 0, no output; success path asserted valid JSON, exactly one line |
| `install.sh` wires idempotently and coexistence-aware | ✅ | Second install → still exactly 2 entries; foreign `clear` hook and a foreign command *inside our `startup` group* both survive |
| Wired unconditionally, no new prompt, identical under `--yes` | ✅ | Called at the same point as `merge_settings_hook`, before the `--dev` early-exit; no `confirm` added |
| `--dev` symlinks the script; settings wiring runs before the dev-mode branch | ✅ | Uses the shared `install_file()`; verified in code that both merges precede the `--dev` `exit 0` |
| `--dry-run` lists the script + the `SessionStart` wiring line | ✅ | Asserted; also asserted `--dry-run` writes nothing |
| `uninstall.sh` removes the entry, prunes empty containers, leaves unrelated hooks | ✅ | Asserted removal, `.hooks.SessionStart` container pruned when empty, both foreign hooks survive |
| Automated tests for fresh / stale / absent / `clear` / malformed stdin | ✅ | All present |
| Shell-wrapper constraints #1,2,3,5,6,7 out of scope; #4 (stat portability) applies | ✅ | Wrapper not implemented; portable stat implemented and exercised |

## Test Plan

```
$ ./hooks/repo/tests/run.sh                        # exit 0 — PASS: 56  FAIL: 0
    (includes) test-session-start-handoff.sh       -> 79 cases pass
$ ./hooks/repo/tests/test-guard-destructive.sh     # exit 0 — Total: 444  Passed: 444  Failed: 0
```

The pre-existing 444-case guard suite was re-run specifically to confirm the shared `install.sh`/`uninstall.sh` edits did not regress the `PreToolUse` wiring.

Manual end-to-end against a scratch repo: `--dry-run` lists both new lines; `install.sh -y` produces the two matcher entries; a synthetic `{"cwd":...,"source":"startup"}` payload yields well-formed `additionalContext`; `"source":"clear"` yields silence + exit 0; `uninstall.sh -y` reports and removes the entry.

## Out of Scope (deliberate)

The shell `claude` wrapper that edits `~/.zshrc`/`~/.bashrc` is **not** implemented, per the curated scope. Nothing in this PR touches any file outside the target repo. Design constraints #1 (alias parsing), #2 (recursion), #3 (subcommand gating), #5 (fish detection), #6 (rc idempotency markers + rc uninstall) and #7 (rc backup) apply only to that wrapper and remain unaddressed; they belong to a follow-up issue if a human still wants it.

No `CHANGELOG.md` entry: this repo adds those in the `Release vX.Y.Z` commit, not in feature PRs (verified against 7caf5fc and 568daa5).

Closes #32

### DIFF
--- a/hooks/repo/session-start-handoff.sh
+++ b/hooks/repo/session-start-handoff.sh
@@ -1,0 +1,165 @@
+#!/usr/bin/env bash
+# session-start-handoff.sh - SessionStart hook that surfaces a pending
+# /repo:handoff note at the start of a Claude Code session.
+#
+# Part of Repo Skills (https://github.com/rjwalters/repo), installed by
+# install.sh into .claude/skills/repo/hooks/ and wired into the consumer repo's
+# .claude/settings.json under hooks.SessionStart (matchers "startup" and
+# "resume").
+#
+# WHY: /repo:handoff writes .claude/handoff.md and adds a pointer line to the
+# agent's memory index. That makes the note findable, but nothing announces it,
+# and the command's one-shot contract ("read it, then delete it") means a note
+# that never gets absorbed lingers silently. handoff.md calls a stale note "a
+# trap of its own". This hook makes both facts visible at session start.
+#
+# =============================================================================
+# STABLE INTERFACE (the contract install.sh/uninstall.sh and the tests rely on)
+# =============================================================================
+#
+# Input:  JSON on stdin (Claude Code SessionStart payload). Fields consumed:
+#           .cwd    - the session's working directory
+#           .source - "startup" | "resume" | "clear" | "compact" | "fork"
+#         Anything else in the payload is ignored.
+#
+# Output: silence + exit 0  => nothing to surface;  otherwise EXACTLY one JSON
+#         object on stdout:
+#           { "hookSpecificOutput": { "hookEventName": "SessionStart",
+#               "additionalContext": "..." } }
+#         The "hookEventName" field is REQUIRED by Claude Code's schema -
+#         without it the additionalContext is silently discarded.
+#
+# Exit:   ALWAYS 0, on every path including internal error. SessionStart has no
+#         decision control (a hook cannot block session start), so the only
+#         thing a failure here can accomplish is noise in the transcript.
+#
+# Errors: this script MUST never exit non-zero and never emit invalid output.
+#         Any internal error is caught by the ERR trap, logged to
+#         <script-dir>/../logs/hook-errors.log, and resolves to silence (fail
+#         open) - mirroring guard-destructive.sh's convention.
+#
+# READ-ONLY: this hook never writes, deletes, or modifies .claude/handoff.md or
+#         any other file in the repo. Deleting the note after it is absorbed is
+#         the /repo:handoff one-shot contract's job, not this hook's. The only
+#         file it may ever write is its own diagnostic error log.
+#
+# SOURCE GATING: only "startup" and "resume" are acted on. install.sh wires
+#         exactly those two matchers, and this script re-checks .source anyway
+#         so a hand-edited settings.json that routes "clear"/"compact"/"fork"
+#         here still no-ops. A /clear is not a process relaunch and the note has
+#         not been touched, so repeating the banner there is pure noise.
+# =============================================================================
+
+# Age thresholds (hours). STALE_HOURS is the ">7d" line handoff.md warns about.
+RECENT_HOURS=48
+STALE_HOURS=168
+
+# Max section headers rendered. The real-world note that motivated this was
+# 7.5 KB - far too much to inject on every launch - but its headers convey the
+# shape in a handful of lines.
+MAX_HEADERS=9
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd 2>/dev/null || echo ".")"
+HOOK_ERROR_LOG="${SCRIPT_DIR}/../logs/hook-errors.log"
+
+# Log a diagnostic error message (best-effort, never fails the script).
+log_hook_error() {
+    local msg="$1"
+    mkdir -p "$(dirname "$HOOK_ERROR_LOG")" 2>/dev/null || true
+    echo "[$(date -u '+%Y-%m-%dT%H:%M:%SZ')] [session-start-handoff] $msg" >> "$HOOK_ERROR_LOG" 2>/dev/null || true
+}
+
+# Top-level error trap: on ANY unexpected error, emit nothing and exit 0. Note
+# this exits BEFORE any stdout is produced (the single printf below is the last
+# statement in the script), so the trap can never truncate a half-written JSON
+# object into malformed output.
+trap 'log_hook_error "Unexpected error on line ${LINENO}: ${BASH_COMMAND:-unknown} (exit=$?)"; exit 0' ERR
+
+# Read stdin safely - if cat fails, the ERR trap fires and we stay silent.
+INPUT=$(cat 2>/dev/null) || INPUT=""
+
+# jq is how we both parse the payload and build well-formed output. Without it
+# we cannot guarantee valid JSON, so stay silent rather than hand-roll one.
+if ! command -v jq &>/dev/null; then
+    log_hook_error "jq not found in PATH - emitting no context"
+    exit 0
+fi
+
+SOURCE=$(printf '%s' "$INPUT" | jq -r '.source // empty' 2>/dev/null) || SOURCE=""
+CWD=$(printf '%s' "$INPUT" | jq -r '.cwd // empty' 2>/dev/null) || CWD=""
+
+# Source gating. An absent/unparseable source means malformed input, which
+# fails open to silence rather than guessing.
+case "$SOURCE" in
+    startup|resume) ;;
+    *) exit 0 ;;
+esac
+
+# Resolve the repo root from cwd (handles worktrees). Fall back to cwd itself
+# when it is not a git repo, mirroring the original proposal's
+# `root=$(git rev-parse --show-toplevel) || root=$PWD`.
+[[ -n "$CWD" && -d "$CWD" ]] || exit 0
+REPO_ROOT=$(git -C "$CWD" rev-parse --show-toplevel 2>/dev/null) || REPO_ROOT=""
+[[ -n "$REPO_ROOT" ]] || REPO_ROOT="$CWD"
+
+NOTE="$REPO_ROOT/.claude/handoff.md"
+[[ -f "$NOTE" && -r "$NOTE" ]] || exit 0
+
+# Portable mtime: BSD/macOS `stat -f %m` vs GNU `stat -c %Y`. If neither works
+# we cannot compute age, so skip the banner entirely rather than emit a partial
+# one or error out.
+MTIME=$(stat -f %m "$NOTE" 2>/dev/null) || MTIME=""
+[[ -n "$MTIME" ]] || MTIME=$(stat -c %Y "$NOTE" 2>/dev/null) || MTIME=""
+[[ "$MTIME" =~ ^[0-9]+$ ]] || { log_hook_error "could not stat mtime of $NOTE - emitting no context"; exit 0; }
+
+NOW=$(date +%s 2>/dev/null) || NOW=""
+[[ "$NOW" =~ ^[0-9]+$ ]] || exit 0
+
+AGE_H=$(( (NOW - MTIME) / 3600 ))
+# Clock skew / a future mtime would otherwise produce a negative age. Written as
+# an `if` rather than `(( )) &&` so the arithmetic test's non-zero status in the
+# common case can never reach the ERR trap.
+if (( AGE_H < 0 )); then AGE_H=0; fi
+
+if   (( AGE_H < 1 ));             then AGE_LABEL="just now"
+elif (( AGE_H < RECENT_HOURS ));  then AGE_LABEL="${AGE_H}h old"
+else                                   AGE_LABEL="$(( AGE_H / 24 ))d old"
+fi
+
+# Headers only - never the file body. `|| true` because grep exits 1 when the
+# note has no headers at all (a valid, non-error state) and head closing the
+# pipe early can surface as a non-zero status.
+HEADERS=$(grep -E '^#{1,2} ' "$NOTE" 2>/dev/null | head -n "$MAX_HEADERS" | sed 's/^#* *//' || true)
+
+CONTEXT="A /repo:handoff note is pending in this repository.
+
+  Path: $NOTE
+  Age:  $AGE_LABEL"
+
+if (( AGE_H >= STALE_HOURS )); then
+    CONTEXT="$CONTEXT
+  STALE: this note is older than 7 days. A handoff describes one moment; verify
+         every claim in it against the repo before acting on it."
+fi
+
+if [[ -n "$HEADERS" ]]; then
+    CONTEXT="$CONTEXT
+
+Sections:
+$(printf '%s\n' "$HEADERS" | sed 's/^/  - /')"
+fi
+
+CONTEXT="$CONTEXT
+
+Read the note in full before doing anything else. Per the /repo:handoff one-shot
+contract, once you have absorbed it, delete both the note and its pointer line
+in the memory index - promote anything durable into real memory files or issues."
+
+OUTPUT=$(jq -cn --arg ctx "$CONTEXT" \
+    '{hookSpecificOutput: {hookEventName: "SessionStart", additionalContext: $ctx}}' 2>/dev/null) || OUTPUT=""
+
+# Never emit anything if jq could not build well-formed JSON.
+[[ -n "$OUTPUT" ]] || { log_hook_error "jq failed to build output JSON - emitting no context"; exit 0; }
+
+printf '%s\n' "$OUTPUT"
+exit 0
--- a/hooks/repo/tests/run.sh
+++ b/hooks/repo/tests/run.sh
@@ -1,10 +1,12 @@
 #!/usr/bin/env bash
-# Test harness for guard-destructive.sh.
+# Test harness for the Repo Skills hooks — the entry point `pnpm test` runs.
 #
 # Pure bash — no external test framework required (Repo Skills ships no test
-# runner, and bats is not assumed to be installed). Each case pipes a Claude
-# Code PreToolUse JSON payload ({"tool_input":{"command":...},"cwd":...}) to the
-# hook and asserts the resulting permissionDecision (deny / ask / allow).
+# runner, and bats is not assumed to be installed). Each guard case pipes a
+# Claude Code PreToolUse JSON payload ({"tool_input":{"command":...},"cwd":...})
+# to guard-destructive.sh and asserts the resulting permissionDecision
+# (deny / ask / allow). The SessionStart handoff hook's suite is delegated to at
+# the end of this file (see test-session-start-handoff.sh).
 #
 # Usage: ./hooks/repo/tests/run.sh
 # Exit status: 0 if all cases pass, 1 otherwise.
@@ -159,6 +161,26 @@ rm -rf "$CFG_REPO" "$LOOM_REPO"
 echo "-- edge: cwd absent / non-git --"
 expect deny  "rm -rf / with empty cwd"         "" "rm -rf /"
 expect allow "ls with nonexistent cwd"         "/nonexistent/path/xyz" "ls"
+
+# The SessionStart handoff hook ships its own suite rather than inline cases:
+# its payload shape, assertions, and install/uninstall wiring checks share
+# nothing with the guard's expect() helper above. Delegate to it and fold the
+# verdict in as a single case so `pnpm test` covers both hooks.
+echo
+echo "-- session-start-handoff.sh (delegated suite) --"
+SS_TEST="$TESTS_DIR/test-session-start-handoff.sh"
+if [[ ! -f "$SS_TEST" ]]; then
+    FAIL=$((FAIL + 1))
+    printf '  FAIL %-52s -> not found at %s\n' "test-session-start-handoff.sh" "$SS_TEST"
+elif SS_OUT="$(bash "$SS_TEST" 2>&1)"; then
+    PASS=$((PASS + 1))
+    printf '  ok   %-52s -> %s\n' "test-session-start-handoff.sh" \
+        "$(printf '%s\n' "$SS_OUT" | awk '/^  Total:/ {print $2 " cases pass"}')"
+else
+    FAIL=$((FAIL + 1))
+    printf '  FAIL %-52s -> suite failed; output follows\n' "test-session-start-handoff.sh"
+    printf '%s\n' "$SS_OUT" | sed 's/^/    /'
+fi
 
 echo
 echo "==============================="
--- a/hooks/repo/tests/test-session-start-handoff.sh
+++ b/hooks/repo/tests/test-session-start-handoff.sh
@@ -1,0 +1,425 @@
+#!/usr/bin/env bash
+# Test suite for hooks/repo/session-start-handoff.sh (and the install.sh /
+# uninstall.sh settings.json wiring that installs it).
+#
+# Usage: ./hooks/repo/tests/test-session-start-handoff.sh
+# Exit code 0 = all tests pass, 1 = failures detected.
+#
+# Structured like test-guard-destructive.sh next door: pure bash, no test
+# framework, PASS/FAIL counters and a summary block. Each hook case pipes a
+# synthetic Claude Code SessionStart payload ({"cwd":...,"source":...}) to the
+# hook and asserts on stdout + exit status.
+#
+# The contract under test (see the hook's header):
+#   - fires only for source "startup"/"resume"
+#   - emits hookSpecificOutput.hookEventName == "SessionStart" with
+#     additionalContext summarizing .claude/handoff.md (path, age, staleness,
+#     section headers only)
+#   - never exits non-zero, never emits malformed JSON, never writes the note
+
+set -uo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+REPO_ROOT="$(cd "$SCRIPT_DIR/../../.." && pwd)"
+HOOK="$REPO_ROOT/hooks/repo/session-start-handoff.sh"
+INSTALL_SH="$REPO_ROOT/install.sh"
+UNINSTALL_SH="$REPO_ROOT/uninstall.sh"
+
+PASS=0
+FAIL=0
+TOTAL=0
+
+RED='\033[0;31m'
+GREEN='\033[0;32m'
+NC='\033[0m'
+
+if [[ ! -f "$HOOK" ]]; then
+    echo "FATAL: hook not found at $HOOK" >&2
+    exit 1
+fi
+if ! command -v jq >/dev/null 2>&1; then
+    echo "FATAL: jq is required to run these tests" >&2
+    exit 1
+fi
+
+# Resolve to a PHYSICAL path: on macOS mktemp -d hands back /var/folders/... but
+# `git rev-parse --show-toplevel` reports /private/var/folders/..., and the hook
+# emits git's answer. Without `pwd -P` here every path assertion below would
+# compare the two spellings and fail.
+SCRATCH="$(cd "$(mktemp -d)" && pwd -P)"
+trap 'rm -rf "$SCRATCH"' EXIT
+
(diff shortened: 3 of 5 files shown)


rjwalters:
✅ **Approved** — all acceptance criteria from issue #32's *Curator Enhancement* section are met, and the three self-reported claims were independently verified rather than taken on trust.

Scope check first: the Curator explicitly ruled the shell `~/.zshrc` wrapper **out of scope** and decided "SessionStart hook only." This PR correctly implements *only* the hook. Not implementing the wrapper is the right call, not an omission.

---

## Verification 1 — the two-literal-matcher decision (`startup`, `resume`)

**Verdict: the wiring is correct and functional.** I verified this against the actual Claude Code hook dispatcher (v2.1.212 bundle) rather than against docs prose.

The event→query derivation confirms SessionStart matches on the source:

```js
case "SessionStart": a = n.source; break;
```

and the matcher comparator is:

```js
function h4_(e, t, r, n) {
  if (!t || t === "*") return true;
  if ((r ? /^[a-zA-Z0-9_|, -]+$/ : /^[a-zA-Z0-9_|]+$/).test(t))
    return t.split(r ? /[|,]/ : "|").map(s => s.trim()).filter(Boolean)
            .flatMap(s => QPn(o3(s), n)).includes(e);
  try { let i = new RegExp(t); ... }
}
```

With `r = p4_.has(hook_event_name)`, and `p4_` **includes `"SessionStart"`**. So `matcher: "startup"` → `["startup"].includes("startup")` → matches. Confirmed working.

One correction to the code comment, **non-blocking**: the rationale that *"alternation is documented for tool matchers but NOT for SessionStart"* is inaccurate as to the implementation — the split-on `|`/`,` branch applies to every event in `p4_`, so `"startup|resume"` would have worked identically. The two-literal-entry form is functionally equivalent and strictly no worse, so no change is required. Also confirmed the "matcher matches no tool" warning (`m4_`) gates on `f4_`, which does **not** contain `SessionStart`, so no spurious warning is emitted. And since a session has exactly one `source`, only one of the two groups can ever fire — no double-invocation.

## Verification 2 — relocating `HOOK_INSTALL_REL` / `HOOK_CMD` / `SETTINGS_JSON` above `--dry-run`

**Verdict: safe.** Traced every definition and use:

- `TARGET` is finalized and canonicalized at `install.sh:145` (`TARGET="$(cd "$TARGET" && pwd)"`) and is never reassigned afterward — the only `TARGET=` sites are 116/140/144/145, all well before the new definition at line 202.
- `HOOK_INSTALL_REL`/`HOOK_CMD` are pure literals with no dependency on anything defined between the old and new positions.
- Sole consumers are `merge_settings_hook()` (line 265) and `merge_settings_sessionstart_hook()` (line 323), both far downstream.

Moving a definition *earlier* can only break if it depends on something defined in between; nothing does. The PreToolUse guard path is unaffected — proven empirically by the 444-case guard suite still passing green.

## Verification 3 — the test-count discrepancy (56 vs 444)

**Verdict: pre-existing and correctly characterized; not a regression, and the runner is not silently under-reporting.**

`run.sh` and `test-guard-destructive.sh` are two distinct suites. `test-guard-destructive.sh:18` documents this on `origin/main`, before this PR: *"The quick smoke suite lives next door in run.sh; this file is the full regression suite."* `run.sh` on `main` contains ~55 inline smoke cases and has never invoked the 444-case regression suite. This PR did not change that.

Independently run:

| Command | Result |
|---|---|
| `./hooks/repo/tests/run.sh` | `PASS: 56  FAIL: 0`, exit 0 |
| `./hooks/repo/tests/test-session-start-handoff.sh` | `Total: 79  Passed: 79`, exit 0 |
| `./hooks/repo/tests/test-guard-destructive.sh` | `Total: 444  Passed: 444`, exit 0 |

All three match the reported numbers exactly.

I also **fault-injected** to confirm the tests aren't vacuous — replacing `session-start-handoff.sh` with a bare `exit 0` stub produced `PASS: 55  FAIL: 1` and `run.sh` exited **1**. The delegation propagates failure correctly (`[[ "$FAIL" -eq 0 ]]` as the final statement). Working tree restored.

---

## Independent end-to-end verification

Built a scratch repo pre-seeded with a hand-authored `SessionStart` `clear` hook, a foreign command **inside** what would become our `startup` group, and unrelated `permissions` content.

- **`--dry-run`** — lists both the script path and the `SessionStart→startup resume` wiring line; `cksum` on `settings.json` unchanged, writes nothing.
- **Install #1** — appended our command to the existing `startup` group (foreign command preserved *ahead* of ours), created a new `resume` group, left the foreign `clear` group and `permissions` untouched.
- **Install #2 (idempotency)** — `ℹ SessionStart handoff hook already wired … (no change)`; still exactly 2 of our entries across 3 groups. No duplication.
- **Uninstall** — summary lists the entry; removes only our 2 entries; **both** foreign hooks and `permissions` survive; the `startup` group is correctly *not* pruned because a foreign command remains in it.

**Hook runtime** (all exit 0):

| Input | Result |
|---|---|
| `startup`, fresh note | Valid single-line JSON; `hookEventName: "SessionStart"`; path + `Age: just now`; headers only |
| `resume`, mtime -8d | `STALE: this note is older than 7 days…` present |
| `clear` (note present) | Silent, exit 0 |
| `not json {{{` / empty stdin / `[1,2,3]` | Silent, exit 0, no crash |
| note absent | Silent, exit 0 |

Headers-only rendering confirmed: body prose was **excluded**, and a `###` heading was correctly **not** treated as a section (`^#{1,2} ` is right). Read-only confirmed via `cksum` + mtime unchanged across a run.

**Nothing outside the repo is touched.** Grepped the diff for `$HOME`, `~/.`, `zshrc`, `bashrc`, `/etc/`, `/usr/` — the only hits are `/usr/local/bin/mine.sh` *string literals used as test fixtures* for the coexistence assertions. The error log path (`<script-dir>/../logs/hook-errors.log`) resolves inside `.claude/skills/repo/`, matching `guard-destructive.sh:76` exactly.

`shellcheck -S warning` clean on all four shell files. `pnpm check:ci` passes (it is a no-op stub in this repo). `mergeStateStatus: CLEAN`, `MERGEABLE`. No GitHub Actions workflows exist in this repo, so `gh pr checks` reporting no checks is expected rather than a missing signal.

---

## Craft notes

The fail-open discipline in `session-start-handoff.sh` is genuinely careful, not cargo-culted. Two details stood out:

- `if (( AGE_H < 0 )); then AGE_H=0; fi` written as an `if` rather than `(( … )) &&`, with a comment explaining that the arithmetic test's non-zero status in the common case would otherwise reach the `ERR` trap. That is exactly the bug most authors ship.
- The single `printf` is the last statement in the file, so the `ERR` trap provably cannot truncate a half-written JSON object into malformed output — and the comment says so.

## Non-blocking follow-ups (no action needed on this PR)

1. **`pnpm test` still skips the 444-case guard regression suite.** Pre-existing and out of scope here — but this PR just introduced a clean delegation mechanism in `run.sh`. Applying the same delegation to `test-guard-destructive.sh` would make `pnpm test` genuinely complete. Worth a small follow-up issue.
2. The alternation rationale in the `install.sh` comment (verification 1 above) overstates the constraint. Harmless; correct if reworded whenever that block is next touched.

Great work — the PR description's self-assessment held up under every check I ran.

PR #40, opened 2026-07-27
test(hooks): fold the 444-case guard-destructive suite into `pnpm test`

rjwalters:
## Summary

`pnpm test` runs `hooks/repo/tests/run.sh`, which reported **56 cases**. The 444-case `hooks/repo/tests/test-guard-destructive.sh` was never invoked by it — so the repo's headline test command exercised ~12% of the guard's regression coverage. With **no CI in this repo** (no `.github/workflows/`; `check:ci` / `check:all` are `echo` stubs), `pnpm test` is the only automated gate, and the guard is the hook that decides whether destructive Bash commands get blocked.

Per the Curator's decided scope (Option A, full merge, no split), `run.sh` now delegates to `test-guard-destructive.sh` and folds **all 444** cases into its totals and exit code. **No `pnpm test:full` or second test command was introduced** — `package.json` is unchanged.

**Final combined count: `PASS: 500   FAIL: 0`** (55 `run.sh` inline smoke cases + 444 from `test-guard-destructive.sh` + 1 for the `test-session-start-handoff.sh` delegation, which stays folded as a single case as the issue anticipated).

**Measured runtime: ~53s** (three consecutive green runs: 53s, 54s, 54s). This is *below* the ~64s curation baseline — the two suites overlap on warm filesystem/`jq` cache rather than being strictly additive; `test-guard-destructive.sh` alone measures 50s standalone on this machine.

## Changes

Only `hooks/repo/tests/run.sh` is modified (+88/-2).

- **New delegation block** for `test-guard-destructive.sh`, modelled on the existing `test-session-start-handoff.sh` block proved in #34. It parses the suite's own `Passed:` / `Failed:` summary lines and adds the **real counts** to `$PASS` / `$FAIL` (rather than collapsing to one case), so the aggregate genuinely reflects 500 cases.
- **Failure surfacing**: on failure it prints the per-suite tally, the delegated exit code, and the individual `FAIL:` lines from the child suite.
- **Fail-safe parsing**: if the summary block is missing or unparseable (e.g. the child suite dies early under its own `set -e`), the block records `1 fail` rather than silently folding in zero. A non-zero child exit with a zero failure count is also forced to `1 fail`.
- **Per-suite breakdown** printed above the aggregate `PASS:` / `FAIL:` line, satisfying the "not one opaque total" criterion.
- **Shared helpers** `strip_ansi` / `suite_count` / `record_suite` added near the top so future suites can be added by copying a block and changing two names — the file header now documents that extension point explicitly.

[...]

PR #41, opened 2026-07-27
fix(branches): correct the permanent-loss check's `--not` toggle and make it content-aware

### LINKED ISSUE #39: branches.md: the permanent-loss check is malformed — `--not` toggles, so it re-includes main
## Context

`commands/repo/branches.md` defines the permanent-loss check — the guard that stands between a `--prune` run and deleted work — using this form:

```bash
git log --oneline <branch> --not --remotes --not main
```

`--not` is a **toggle**, not a per-argument negation. The first `--not` starts excluding refs; the second one flips the sense *back to positive*, so `main` is included as a positive ref rather than excluded. The command therefore asks "commits on `<branch>`, excluding remotes, **plus** all of `main`" — not "commits that exist nowhere else."

### Repro (observed this session)

Run against a squash-merged branch (`feature/issue-32`, whose only real work is `4415d45`, already in `main`):

```
$ git log --oneline feature/issue-32 --not --remotes --not main
abae8dc docs: document the SessionStart handoff hook; ignore Loom sweep-checkpoint + locks state
4415d45 feat(hooks): surface a pending /repo:handoff note via a SessionStart hook
```

`abae8dc` exists **only on `main`** (confirmed via `git branch --all --contains abae8dc` → only `main`) — it was committed to `main` after `feature/issue-32` branched off and has never been on that branch. Its presence in the output is proof the exclusion is not doing what the doc intends.

The correct form returns only the branch-only commit, with no `main`-only commits leaking in:

```
$ git log --oneline main..feature/issue-32
4415d45 feat(hooks): surface a pending /repo:handoff note via a SessionStart hook
```

Reproduced a second time against `feature/issue-25` — same pattern: the malformed form leaks `abae8dc` (a `main`-only commit) into the output, while `git log --oneline main..feature/issue-25` correctly returns only `e9185b0`.

### Why this matters more than the wrong output

Earlier in the same session, before `git fetch --prune` ran, the malformed command returned **empty** for both merged branches — because the still-present `origin/feature/*` tracking refs happened to cover the branch tips. Empty reads as "no unique commits — safe to delete." The verdict was correct, but by accident: it came from stale tracking refs, not from the check working.

That is the dangerous failure mode. A loss check that returns the right answer for the wrong reason will keep returning the right answer until the day the refs line up differently — and that day it deletes work.

### The squash-merge subtlety (load-bearing — do not "fix" this away)

`git branch --merged main` does **not** list a squash-merged branch, because a squash-merge creates a brand-new commit on `main` whose parent is `main`'s prior tip — the branch's original commits never become ancestors of `main`. Confirmed in this repo: `git branch --merged main` lists only `main` itself, even though both `feature/issue-32` and `feature/issue-25` are fully-merged, closed-out branches.

**A naive "fix" that swaps the check to `--merged` would classify every squash-merged branch as UNSAFE and silently stop pruning anything** — trading one silent failure (false SAFE) for another (false UNSAFE, defeating the feature entirely without any error).

The correct check must reason about **content containment**, not SHA ancestry:
- `git diff main...<branch>` is empty (three-dot: diff of `<branch>` against the merge-base with `main`), or
- the branch's PR (`gh pr list --head <branch> --state merged`) is merged.

SHA-ancestry checks (`--merged`, `main..<branch>` alone) answer "are these exact commits on `main`," which is false for squash-merges. Content-containment checks answer "does `main` already have this work," which is the question `--prune` actually needs answered.

## Audit: is the same idiom used elsewhere?

```
$ grep -rn -- "--not" commands/repo/
commands/repo/branches.md:127:git log --oneline <branch> --not --remotes --not <default>
commands/repo/release.md:245:gh release create ... --notes-file <(...)   # false positive: "--notes-file", not "--not"
```

`reset.md` does **not** duplicate the idiom — it delegates to `branches.md` by reference:
```
commands/repo/reset.md:56-59:  Run the full [[branches]] classification ... [[branches]]'
                                **permanent-loss check** applies — a branch with commits found ...
```
So `reset.md` is fixed automatically once `branches.md` is corrected; no separate edit needed there, but the AC below requires confirming that delegation still reads correctly after the fix (no stale wording implying the check works differently than what `branches.md` now says).

`commands/repo/branches.md:127` is the **only** site with the malformed `--not`-toggle idiom in `commands/repo/`.

## Existing test coverage (checked before writing the AC below)

This repo's only test harness (`pnpm test` → `./hooks/repo/tests/run.sh`) exercises **hooks** (`hooks/repo/*.sh`) — executable shell scripts invoked via PreToolUse JSON payloads piped through `jq`/`bash`. It does not touch `commands/repo/*.md` at all, and there is no markdown-doctest / prose-extraction tooling anywhere in the repo (`.loom/scripts/tests/`, `hooks/repo/tests/`) that runs snippets out of skill files. `commands/repo/branches.md` is prose an agent reads and acts on, not code that gets sourced or executed by any existing suite — so a traditional "run the command file and assert output" regression test is not expressible here.

The lightest thing that *would* actually catch a regression: the permanent-loss check is a single, self-contained git invocation, independent of the surrounding agent/prose logic. It can be tested directly, using the same scratch-repo pattern `hooks/repo/tests/*.sh` already uses (`mktemp -d`, `git init`, build fixture branches, assert on command output) — just targeting the git command from `branches.md` instead of a hook script. See AC below.

## Acceptance Criteria

- [ ] **Fix the idiom.** Replace `git log --oneline <branch> --not --remotes --not <default>` in `commands/repo/branches.md` (currently line 127, step 5 "If `--prune` flag is set") with a form that actually excludes both `<default>` and the remotes, e.g.:
  ```bash
  git log --oneline <branch> --not <default> --remotes
  ```
  (single `--not`, both exclusions after it — do NOT reintroduce a second `--not`), or the equivalent explicit two-ref form `git log --oneline <branch> --not <default> origin/<default>` if `--remotes` is considered too broad. Verify the fix in this repo before closing: it must NOT show `abae8dc` for `feature/issue-32` or `feature/issue-25`.

- [ ] **Make the check content-aware, not just SHA-ancestry-aware.** The permanent-loss check must classify a branch as safe-to-delete only when its content is fully contained in `<default>` OR its PR is merged — not merely when its raw SHAs are unreachable from `<default>`. Update `branches.md` step 5 to run both:
  1. `git log --oneline <branch> --not <default> --remotes` (fixed ancestry check, catches genuinely unpushed/unmerged commits), AND
  2. for branches where (1) is non-empty, a content check before concluding UNSAFE: `git diff --quiet main...<branch>` (exit 0 = no content difference) or a merged PR lookup (`gh pr list --head <branch> --state merged --json number --jq length`) — either passing means SAFE despite non-empty step-1 output (this is the squash-merge case).
  - Explicitly document in the skill file: **do not** "fix" this by switching to `git branch --merged <default>` alone — that under-counts squash-merged branches as unsafe and would silently stop pruning them (see squash-merge subtlety above). This sentence (or equivalent) must appear in the file so a future editor doesn't reintroduce that regression.

- [ ] **Specify and enforce the failure direction.** Add an explicit line to step 5 (near the existing "Empty → ... / Non-empty → ..." bullets) stating: when the check is ambiguous, errors (e.g. `gh` unavailable/rate-limited, `git diff` fails, network error on a remote lookup), or any command in the check exits non-zero unexpectedly, the branch MUST be classified as UNKNOWN/KEEP — never as SAFE. This must hold even under `--prune`.

- [ ] **Confirm `reset.md`'s delegation still reads correctly.** No separate `--not` occurrence exists there today (confirmed by audit above), but re-read `commands/repo/reset.md` lines ~56-59 after the `branches.md` edit and confirm the delegating language ("the loss check applies") doesn't imply stale semantics. No changes expected, but note in the PR if any are made.

- [ ] **Add a regression test**, since no existing harness executes `commands/repo/*.md` (confirmed above — `pnpm test` only covers `hooks/repo/*.sh`). Add a new script, e.g. `commands/repo/tests/test-branches-loss-check.sh`, following the scratch-repo convention already used in `hooks/repo/tests/test-guard-destructive.sh` (`mktemp -d`, `git init -q`, build fixtures, assert on `git`/`gh`-command output, `PASS`/`FAIL` counters, non-zero exit on failure). It must build two fixtures against a synthetic `main`:
  1. **Squash-merged branch**: create a branch, commit some content, squash-merge it into `main` as a *new* commit (so the branch's original commit SHAs are NOT ancestors of `main`), then run the corrected check → must classify **SAFE**.
  2. **Unpushed-commit branch**: create a branch with a commit whose content is genuinely absent from `main` (no squash, no cherry-pick, nothing merged), then run the corrected check → must classify **UNSAFE**, and confirm the branch is never included in a "safe to delete" set.
  - Wire the new script into `pnpm test` (extend `hooks/repo/tests/run.sh`'s delegation pattern, or add a second line to the `test` script in `package.json` — either is acceptable, pick whichever keeps `pnpm test` a single entry point).

## Affected Files

- `commands/repo/branches.md` (line 127 and surrounding step-5 prose — the fix)
- `commands/repo/reset.md` (read-only verification of delegating language; no fix expected)
- `commands/repo/tests/test-branches-loss-check.sh` (new — regression test)
- `hooks/repo/tests/run.sh` and/or `package.json` (wire the new test into `pnpm test`)

rjwalters:
## Summary

`commands/repo/branches.md` step 5 — the guard standing between a `--prune` run and permanently deleted branches — was written as:

```bash
git log --oneline <branch> --not --remotes --not <default>
```

`--not` is a **toggle**, not a per-argument negation. The second `--not` flips `<default>` back to a *positive* ref, so the command actually asked for "commits on `<branch>`, excluding remotes, **plus** all of `<default>`". The exclusion silently stopped working; the check returned empty only when unrelated refs happened to cover the branch tip — the right answer for the wrong reason, which holds until the day the refs line up differently.

Fixing the toggle alone is not enough. A squash-merge replays the branch as one brand-new commit whose parent is `<default>`'s prior tip, so the branch's original SHAs are *never* ancestors of `<default>` and always appear in the ancestry check even when every line of the work is merged. The check therefore had to become **content-aware**.

## Changes

- **`commands/repo/branches.md`** — step 5 rewritten as three explicit stages:
  - **5a ancestry** — single `--not`, both exclusions after it, plus an explanation of the toggle semantics and a "never add a second `--not`" rule so the bug cannot be reintroduced by a well-meaning edit.
  - **5b content containment** — `git merge-tree --write-tree <default> <branch>` compared against `<default>^{tree}` ("would merging this change anything?"), falling back to `git diff --quiet` and then a merged-PR lookup. Includes an explicit prohibition on "simplifying" this to `git branch --merged`.
  - **5c failure direction** — any error, unparseable output, or unavailable command classifies **UNKNOWN / KEEP, never SAFE**, under `--prune` exactly as without it.
- **`commands/repo/tests/test-branches-loss-check.sh`** (new) — first test to cover `commands/repo/` at all.
- **`hooks/repo/tests/run.sh`** — one additive delegation block appended (kept minimal; concurrent PRs are touching this file).

## A correction to the acceptance criteria

The AC proposed `git diff --quiet main...<branch>` as the content check. **This does not work.** For `git diff`, three-dot means `merge-base(main, branch)..branch` — the branch's *own* changes — which is non-empty for a squash-merged branch and would have misclassified it UNSAFE, i.e. the exact false-UNSAFE regression the AC set out to prevent. Verified against the fixture:

```
git diff --quiet main...feature/squashed   -> exit 1 (NON-empty)
git merge-tree --write-tree main feature/squashed == main^{tree}   -> true
```

`merge-tree` is used instead. It also survives `main` advancing past the merge point, which the two-dot form does not.

| Fix the malformed `--not` idiom | Done | Fixture asserts the malformed form leaks a `main`-only commit and the fixed form does not; a doc-drift test asserts no `git log` line in `branches.md` carries a second `--not` |
| Make the check content-aware, not SHA-ancestry-aware | Done | 5b uses `merge-tree`/`diff`/merged-PR. Squash-merged fixture classifies SAFE with **no** forge lookup at all |
| Document "do not switch to `git branch --merged`" | Done | Present in `branches.md` 5b and asserted by the doc-drift block |
| Specify and enforce the failure direction | Done | 5c added; tests cover forge-unavailable and unknown-ref paths, both yielding KEEP |
| Confirm `reset.md` delegation still reads correctly | Done, no change | It delegates by reference ("a branch with commits found nowhere else ... is never removed automatically"), which holds under the new semantics. A test asserts it still delegates and has not grown its own copy of the idiom |
| Add a regression test with both fixtures | Done | Squash-merged -> SAFE; genuinely-unpushed -> never in the safe-to-delete set |
| Wire into `pnpm test` | Done | Delegated from `run.sh`; `pnpm test` remains one entry point |

## Test Plan

```
pnpm test                                  -> PASS: 532  FAIL: 0   (exit 0)
  run.sh inline smoke cases                -> 55 cases pass
  test-guard-destructive.sh                -> 444 cases pass
  test-session-start-handoff.sh            -> 79 cases pass (folded as 1)
  test-install-claude-md-markers.sh        -> 49 cases pass (folded as 1)
  test-branches-loss-check.sh              -> 31 cases pass (folded at full weight)
hooks/repo/tests/test-guard-destructive.sh -> Total: 444  Passed: 444  Failed: 0  (exit 0)
```

The new suite was **mutation-tested** to confirm it is load-bearing rather than merely green:

1. Reintroducing the double `--not` in `branches.md` -> 2 failures.
2. Replacing 5b's content check with the naive `git branch --merged` -> 3 failures, precisely the regression the AC warns about:
   ```
   FAIL: squash-merged -> SAFE (without any PR lookup)   expected [SAFE], got [UNSAFE]
   FAIL: squash-merged branch IS offered for deletion    missing [feature/squashed]
   ```
3. Making 5c fail *open* (errors -> SAFE) -> 4 failures.

Fixture note: the `main`-only commit is deliberately left **unpushed**. Had it been pushed, `--remotes` would mask the leak and the malformed form would look correct — that accidental masking is the original "empty for the wrong reason" failure mode, so the fixture pins it down rather than reproducing it by luck.

Closes #39

### DIFF
--- a/commands/repo/branches.md
+++ b/commands/repo/branches.md
@@ -120,21 +120,76 @@ STALE WORKTREES (0):
 worktree is removed until it passes — irreversible removal of work that exists
 nowhere else is never acceptable, `--prune` or not.
 
+The check runs in two stages, plus one fixed rule about which way it fails.
+
+#### 5a. Ancestry — does the branch carry commits that exist nowhere else?
+
 For every branch about to be deleted, list commits that live *only* on that
 branch — not reachable from the default branch and not present on any remote:
 
 ```bash
-git log --oneline <branch> --not --remotes --not <default>
+git log --oneline <branch> --not <default> --remotes
 ```
 
-- **Empty** → the work is preserved elsewhere (merged or pushed); safe to delete.
-- **Non-empty** → those commits would be **permanently lost**. Do NOT auto-delete
-  even under `--prune`. Reclassify the branch as UNKNOWN, show the count and the
-  commit subjects, and require an explicit per-branch confirmation.
+- **Empty** → the branch carries no commits of its own; safe to delete.
+- **Non-empty** → this proves nothing yet. Go to 5b before concluding.
 
-(A merged PR whose remote branch was already deleted may show commits here after
-a squash-merge — that's the safe direction to err: surface it and let the user
-confirm the work is truly in the PR.)
+`--not` is a **toggle**, not a per-argument negation: it flips the sense of every
+ref that follows it, up to the next `--not`. Both exclusions must therefore sit
+after a **single** `--not`. Writing it as
+`<branch> --not --remotes --not <default>` flips the sense back to positive and
+folds all of `<default>`'s own commits into the output, so the exclusion silently
+stops working — the check then reports commits the branch never had, and returns
+empty only when unrelated refs happen to cover the tip. **Never add a second
+`--not`.** If `--remotes` is too broad for the repo, spell the exclusions out
+instead: `git log --oneline <branch> --not <default> origin/<default>`.
+
+#### 5b. Content — does `<default>` already contain this work?
+
+A non-empty 5a result does **not** mean the work would be lost. A squash-merge
+replays the branch's changes as one brand-new commit whose parent is
+`<default>`'s prior tip, so the branch's original commits never become ancestors
+of `<default>` and *always* appear in 5a — even when every line of the work is
+already merged. Decide with **content containment**, not SHA ancestry:
+
+```bash
+# Would merging the branch into <default> change anything at all?
+[ "$(git merge-tree --write-tree <default> <branch>)" = "$(git rev-parse '<default>^{tree}')" ]
+```
+
+Equal trees → `<default>` already holds the branch's content → safe to delete.
+If `git merge-tree` is unsupported (git < 2.38) or reports a conflict, fall back
+to the exact-match form `git diff --quiet <default> <branch>`: exit 0 (identical
+trees) is also proof of containment, while a non-zero exit proves nothing either
+way. Failing both, ask the forge whether the branch's PR was merged:
+
+```bash
+gh pr list --head <branch> --state merged --json number --jq length
+```
+
+A count `>= 1` means the work landed through that PR; safe to delete.
+
+If none of these establish containment, the 5a commits would be **permanently
+lost**. Do NOT auto-delete even under `--prune`. Reclassify the branch as
+UNKNOWN, show the count and the commit subjects, and require an explicit
+per-branch confirmation.
+
+**Do NOT "simplify" 5b to `git branch --merged <default>`.** That lists only
+branches whose commits are literal ancestors of `<default>`, which is never true
+after a squash-merge — every squash-merged branch would be classified unsafe and
+`--prune` would silently stop pruning anything, defeating the feature with no
+error to notice. `--merged` answers "are these exact commits on `<default>`";
+pruning needs "does `<default>` already have this work". Different questions.
+
+#### 5c. Failure direction — ambiguity and errors always mean KEEP
+
+If any command in 5a or 5b exits non-zero unexpectedly, emits output that cannot
+be parsed, or cannot run at all — `gh` missing, unauthenticated, or rate-limited;
+no network; an unknown or ambiguous ref; `git merge-tree` unsupported — the
+branch is classified **UNKNOWN / KEEP**. **Never SAFE.** Ambiguity is never
+resolved in favour of deletion, and this holds under `--prune` exactly as it does
+without it. A branch wrongly kept costs one line of report noise; a branch
+wrongly deleted costs the work.
 
 For each worktree about to be removed, refuse if it has uncommitted changes —
 that work exists nowhere else:
@@ -151,7 +206,9 @@ Then delete the branches that passed the loss check:
 
 ```bash
 git branch -d <branch_name>           # -d (safe): refuses if not merged
-# escalate to -D only for a branch the loss check proved is pushed/merged
+# escalate to -D only for a branch the loss check proved is pushed, or whose
+# content 5b proved is already contained in <default> (the squash-merge case,
+# where -d refuses because the original SHAs are not ancestors of <default>)
 ```
 
 Report what was deleted, what was skipped for potential data loss, and what
--- a/commands/repo/tests/test-branches-loss-check.sh
+++ b/commands/repo/tests/test-branches-loss-check.sh
@@ -1,0 +1,375 @@
+#!/usr/bin/env bash
+# Test suite for the permanent-loss check documented in commands/repo/branches.md
+# step 5 — the guard that stands between `/repo:branches --prune` (and
+# `/repo:reset --prune`, which delegates to it) and permanently deleted work.
+#
+# Usage: ./commands/repo/tests/test-branches-loss-check.sh
+# Exit code 0 = all tests pass, 1 = failures detected.
+#
+# Structured like hooks/repo/tests/test-guard-destructive.sh and
+# test-session-start-handoff.sh: pure bash, no test framework, PASS/FAIL/TOTAL
+# counters and a summary block. `pnpm test` delegates to this file via
+# hooks/repo/tests/run.sh.
+#
+# WHY THIS FILE EXISTS (repo#39): branches.md is prose an agent reads, not code
+# any harness executes, so nothing in this repo could catch a regression in the
+# loss check. But the check is a small set of self-contained git invocations,
+# independent of the surrounding prose — so they can be run directly against
+# scratch-repo fixtures. loss_check() below is a faithful transcription of
+# branches.md step 5; the doc-drift block at the end asserts the file still says
+# what this suite implements.
+#
+# The contract under test:
+#   5a  git log --oneline <branch> --not <default> --remotes   (ONE --not)
+#   5b  content containment (merge-tree / diff / merged-PR), NOT SHA ancestry
+#   5c  any error or ambiguity classifies KEEP, never SAFE
+
+set -uo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+REPO_ROOT="$(cd "$SCRIPT_DIR/../../.." && pwd)"
+BRANCHES_MD="$REPO_ROOT/commands/repo/branches.md"
+RESET_MD="$REPO_ROOT/commands/repo/reset.md"
+
+PASS=0
+FAIL=0
+TOTAL=0
+
+RED='\033[0;31m'
+GREEN='\033[0;32m'
+NC='\033[0m'
+
+if [[ ! -f "$BRANCHES_MD" ]]; then
+    echo "FATAL: branches.md not found at $BRANCHES_MD" >&2
+    exit 1
+fi
+
+SCRATCH="$(cd "$(mktemp -d)" && pwd -P)"
+trap 'rm -rf "$SCRATCH"' EXIT
+
+# ---------------------------------------------------------------------------
+# Assertion helpers
+# ---------------------------------------------------------------------------
+
+ok() {   # <label>
+    TOTAL=$((TOTAL + 1)); PASS=$((PASS + 1))
+    printf "  ${GREEN}PASS${NC}: %s\n" "$1"
+}
+no() {   # <label> <detail>
+    TOTAL=$((TOTAL + 1)); FAIL=$((FAIL + 1))
+    printf "  ${RED}FAIL${NC}: %s\n" "$1"
+    [[ -n "${2:-}" ]] && printf "        %s\n" "$2"
+    return 0
+}
+assert_eq() {  # <label> <expected> <actual>
+    if [[ "$2" == "$3" ]]; then ok "$1"; else no "$1" "expected [$2], got [$3]"; fi
+}
+assert_contains() {  # <label> <haystack> <needle>
+    if [[ "$2" == *"$3"* ]]; then ok "$1"; else no "$1" "missing [$3]"; fi
+}
+assert_not_contains() {  # <label> <haystack> <needle>
+    if [[ "$2" != *"$3"* ]]; then ok "$1"; else no "$1" "unexpected [$3] present"; fi
+}
+# The only verdict that can destroy work is SAFE. UNSAFE and KEEP are both
+# "do not delete", so most assertions care about that boundary, not which of
+# the two non-deleting verdicts came back.
+assert_not_safe() {  # <label> <actual>
+    if [[ "$2" != "SAFE" ]]; then ok "$1"; else no "$1" "got SAFE — this branch would be deleted"; fi
+}
+
+# ---------------------------------------------------------------------------
+# The check under test — a direct transcription of branches.md step 5.
+# ---------------------------------------------------------------------------
+
+DEFAULT="main"
+REPO=""
+
+# Stands in for `gh pr list --head <branch> --state merged --json number --jq length`.
+# PR_LOOKUP_MODE: none -> "0" | merged -> "1" | fail -> non-zero exit (gh
+# missing / unauthenticated / rate-limited / offline).
+PR_LOOKUP_MODE="none"
+pr_merged_count() {  # <branch>
+    case "$PR_LOOKUP_MODE" in
+        merged) echo 1 ;;
+        fail)   return 7 ;;
+        *)      echo 0 ;;
+    esac
+}
+
+# loss_check <branch> -> SAFE | UNSAFE | KEEP
+#   SAFE   : work is preserved elsewhere; branch may be deleted under --prune
+#   UNSAFE : branch holds commits whose content exists nowhere else
+#   KEEP   : ambiguous or errored (5c) — must be treated exactly like UNSAFE
+loss_check() {
+    local branch="$1" unique mt tree n
+
+    # -- 5a: ancestry. ONE --not, both exclusions after it.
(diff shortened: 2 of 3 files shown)


rjwalters:
🔧 Rebased onto `main` (`06b802a`) and resolved the `hooks/repo/tests/run.sh` conflict.

**What conflicted.** This PR was branched at `abae8dc`; since then #42 added a delegation block for `test-install-claude-md-markers.sh`, and #40 rewrote the runner's machinery (`strip_ansi`, `suite_count`, `record_suite`) so delegated suites fold their **real** case counts into the totals and get a row in the per-suite breakdown. This PR's block was written in the older inline-awk shape, which collided with #42's block in the same region.

**How it was resolved.** The old inline-awk block was not preserved — it was rewritten in #40's `record_suite` shape, modelled on the `test-guard-destructive.sh` block (the extension point the file header points at). It now folds `BL_PASS`/`BL_FAIL` into the aggregate rather than counting as 1, guards the unparseable-summary case so a suite that dies early can't fold in as zero failures, and calls `record_suite` on every path including "not found". Appended after #42's block, which is left untouched. The header's "currently delegated" list was updated to match.

**Result** — the suite's 31 cases now fold into the total instead of collapsing to 1:

```
-- branches.md permanent-loss check (delegated suite) --
  ok   test-branches-loss-check.sh                          -> 31 cases pass

Per-suite breakdown
  run.sh inline smoke cases                55 pass     0 fail
  test-guard-destructive.sh               444 pass     0 fail  (full regression)
  test-session-start-handoff.sh            79 pass     0 fail  (folded into totals as 1 case)
  test-branches-loss-check.sh              31 pass     0 fail  (branches.md loss check)
PASS: 532   FAIL: 0
```

532 = main's 501 + 31, exactly as expected. `hooks/repo/tests/test-guard-destructive.sh` standalone still reports 444/444.

**No behaviour changed.** `commands/repo/branches.md` and `commands/repo/tests/test-branches-loss-check.sh` are byte-identical to their pre-rebase blobs — verified via `git hash-object`. In particular step 5b still uses `git merge-tree --write-tree <default> <branch>` compared against `<default>^{tree}` for content containment; it was **not** simplified to `git diff main...<branch>` or `git branch --merged`, both of which misclassify a squash-merged branch as UNSAFE and silently stop pruning.

Note: the Test Plan table in the PR description quotes the pre-rebase `PASS: 57`; the current number is `PASS: 532` for the reason above. Still awaiting first review — `loom:review-requested` unchanged.

rjwalters:
## Approved

Reviewed at `e389dd4` against base `06b802a`. Merge state `CLEAN`. This repo has no GitHub Actions (`gh pr checks` → "no checks reported"); as `hooks/repo/tests/run.sh` itself documents, `pnpm test` is the only automated gate, and it was run in full below.

### 1. The deviation from the acceptance criteria is correct — the AC was wrong

The AC prescribed `git diff --quiet main...<branch>` as the content check. The builder rejected it and used `git merge-tree --write-tree <default> <branch>` compared against `<default>^{tree}` (`branches.md:155-158`). I verified this independently on a fresh scratch repo: a branch squash-merged into `main`, with `main` then advanced by an unrelated commit.

```
git diff --quiet main...feat   -> exit 1   (would classify UNSAFE — WRONG)
git diff --quiet main..feat    -> exit 1   (would classify UNSAFE — WRONG)
git branch --merged main       -> main     (squash-merged branch absent — WRONG)
merge-tree --write-tree main feat == main^{tree}  -> EQUAL -> SAFE  (CORRECT)
```

For `git diff`, three-dot means `merge-base(A,B)..B` — the branch's *own* changes — which is exactly what a squash-merge leaves non-empty. Had the AC been followed literally, every squash-merged branch would classify UNSAFE and `--prune` would silently stop pruning: the same class of silent failure the issue set out to eliminate, just inverted. The deviation is not a shortcut, it is the fix. Documenting it in the PR body rather than quietly following the letter of the AC was the right call.

`merge-tree` also survives `main` advancing past the merge point (verified above with `main` three commits ahead, and again in fixture 5 of the test suite, where a `MAINONLY` commit is added after the squash). Where `main` has *modified* the merged content, `merge-tree` conflicts and 5b correctly falls through to the merged-PR lookup — and if that is unavailable, to KEEP. Every fallthrough errs toward keeping work.

The 5b fallback ordering is also sound: skipping `git diff --quiet` when `merge-tree` succeeds but the trees differ is safe, because a differing merge result implies a non-empty two-tree diff. `diff` is a strictly weaker containment proof, only reached when `merge-tree` cannot answer.

### 2. Prose and failure direction

- `grep -rn -- '--not' commands/` shows the malformed idiom gone repo-wide. Every remaining hit is the corrected single-`--not` form, prose explaining the toggle, or the `--notes-file` false positive in `release.md:245`.
- 5c (`branches.md:184-192`) is explicit and unambiguous: non-zero exit, unparseable output, missing/unauthenticated/rate-limited `gh`, unknown ref, unsupported `merge-tree` → **UNKNOWN / KEEP**, "**Never SAFE**", "this holds under `--prune` exactly as it does without it."
- The `-D` escalation comment (`branches.md:206-210`) was updated to explain why `-d` refuses on a squash-merged branch. Easy to overlook, and its absence would have been the next foot-gun.
- `reset.md:56-59` still delegates by reference and has not grown its own copy of the idiom. Its wording ("a branch with commits found nowhere else … is never removed automatically") reads correctly under the new semantics — it describes the guarantee, not the mechanism.

### 3. The tests bind — mutation-tested independently

`commands/repo/tests/test-branches-loss-check.sh` → **31/31 pass**. I re-ran all three mutations from scratch rather than taking the PR body's word:

| Mutation | Failures | Which assertions |
|---|---|---|
| 5b content check → naive `git branch --merged` | **3** | `squash-merged -> SAFE (without any PR lookup)`, `squash-merged branch IS offered for deletion`, `forge failure leaves squash-merged branch SAFE on content alone` |
| Double `--not` reintroduced in `branches.md` | **2** | `no 'git log' command in branches.md uses a second --not`, `branches.md documents the corrected 5a idiom` |
| 5c fails open (`KEEP` → `SAFE`) | **5** | all four degraded-mode assertions plus `unknown ref -> KEEP` |

The `--merged` mutation is the one that matters and it flips the squash-merged fixture exactly as it must. The suite is load-bearing, not decorative.

One observation: mutating the *executed* 5a transcription to the double-`--not` form trips only 1 assertion (`pushed branch -> SAFE`), because the toggle bug is pinned by the literal `MALFORMED`/`FIXED` comparison at lines 227-231 and by the doc-drift block rather than by the transcribed helper. That is the right emphasis here — `branches.md` is prose an agent reads, so the doc is the artifact under test — but worth knowing if the helper is ever extended.

Fixture design is careful. Leaving the `MAINONLY` commit deliberately unpushed is the detail that makes the suite meaningful: had it been pushed, `--remotes` would mask the leak and the malformed form would look correct. That is precisely the "right answer for the wrong reason" mode from the issue, and the fixture pins it rather than reproducing it by luck.

### 4. Rebase preserved the logic

Both files are byte-identical pre/post rebase (`1042b7e` → `e389dd4`):

```
commands/repo/branches.md                        816e6126b2a8853d6dbbf2280ad7470a16a949c5  (both)
commands/repo/tests/test-branches-loss-check.sh  21e4692ce1eb08cc67d7913311d9393181153a40  (both)
```

Only `hooks/repo/tests/run.sh`'s delegation block was rewritten into #40's `record_suite` shape. That block handles the failure modes correctly: missing file, unparseable summary (never folds in as zero failures), and non-zero exit with zero counted failures all force `FAIL >= 1`.

### 5. Post-merge totals

```
main   (06b802a):  PASS: 501   FAIL: 0
PR     (e389dd4):  PASS: 532   FAIL: 0     (exit 0)
```

Delta is exactly 31 — the new suite folds at full weight, not as 1. Per-suite breakdown carries `test-branches-loss-check.sh   31 pass   0 fail  (branches.md loss check)`. `test-guard-destructive.sh` still reports 444/444. `shellcheck -S warning` is clean on both the new test and `run.sh`.

### Non-blocking notes

1. **Stale totals in the PR body — fixed during review.** The Test Plan quoted the pre-rebase `PASS: 57`. I updated it to the current 532 with the full per-suite breakdown rather than blocking; a stale number in a description is not worth a round trip on a correct fix.
2. **Pre-existing, not this PR:** `test-install-claude-md-markers.sh` runs and reports 49 cases but never calls `record_suite`, so it is absent from the per-suite breakdown and folds into the totals as 1. Present on `main` at `06b802a` already. Worth a follow-up issue for consistency with the suite this PR wires in at full weight.
3. **`git merge-tree --write-tree` needs git >= 2.38.** `branches.md` documents the fallback for older git, but line 252 of the test asserts on `merge-tree` directly with no fallback, so the suite itself would fail on git < 2.38. No CI here and local git is 2.55, so this is theoretical — noting it in case the suite ever runs somewhere older.

Approving. The correctness argument holds, the deviation from the AC is justified and independently verified, and the tests demonstrably fail when the logic is broken.

Issue #37, opened 2026-07-27
docs(install.sh): SessionStart matcher comment claims alternation is unsupported — it is supported


rjwalters:
## Original Issue

`install.sh` wires the `SessionStart` hook (added in #34) as **two literal matcher entries** — one for `startup`, one for `resume` — rather than a single `"startup|resume"` matcher. The code is correct and works. The comment explaining *why* is factually wrong.

The comment claims alternation (`"startup|resume"`) is not supported for `SessionStart` matchers, and that a matcher which silently matched nothing would leave the hook inert with no error.

## What is actually true

During review of #34 the Judge checked this against the Claude Code binary itself (v2.1.212) rather than docs prose. The matcher comparator takes a split-on-`|`-or-`,` branch when the hook event is in a specific allow-set, and that allow-set **does include `SessionStart`**. So `"startup|resume"` would have worked fine.

The two-entry form is functionally identical and strictly no worse, which is why review did not block on it. But the comment now actively misinforms the next person who touches this block — it would talk them out of a form that is in fact supported, and the stated failure mode (silent inert matcher) is not real here.

## Fix

Reword the comment block in `install.sh` to state the real reason for the two-entry form: it is explicit and unambiguous, and it avoids depending on alternation semantics that differ between tool matchers and event-source matchers. Drop the claim that alternation is unsupported for `SessionStart`.

## Related

- #34 — introduced the comment

---

### Re-verified against current `main` (`9c0a445`)

`install.sh` changed substantially since this issue was filed — PR #42 (issue #38) refactored CLAUDE.md block surgery into a new shared helper `lib/claude-md-block.sh` (`install.sh` lines 57-87). That refactor is entirely separate from the SessionStart block; it does not touch, move, or reword the comment described here. **This issue is still fully valid.**

The false-claim comment is exactly where #37 describes it, currently at `install.sh:195-202`:

```
195  # The SessionStart handoff hook, same resolution. Claude Code's SessionStart
196  # matchers key on the session SOURCE ("startup" | "resume" | "clear" |
197  # "compact" | "fork"), not on a tool name. We wire the two literal sources the
198  # hook acts on rather than an alternation like "startup|resume": alternation is
199  # documented for tool matchers but NOT for SessionStart, and a matcher that
200  # silently matched nothing would leave the hook inert with no error. Two literal
201  # entries use only documented values. The hook re-checks .source itself, so this
202  # is belt-and-braces rather than a load-bearing filter.
```

[...]

PR #42, opened 2026-07-27
fix(install): anchor CLAUDE.md block rewrites to the marker string, not the line

### LINKED ISSUE #38: install.sh deletes an adjacent tool's CLAUDE.md marker block
## Context

Hit while running `/repo:all` in `rjwalters/vibesql`, updating Repo Skills 0.4.3 → 0.6.1 via `install.sh -y <repo>`.

That repo's `CLAUDE.md` had the Repo Skills and Loom marker blocks written back-to-back with **no newline between them**:

```
removals. Managed by `install.sh` — edit outside the markers only.
<!-- END REPO-SKILLS --><!-- BEGIN LOOM ORCHESTRATION -->
This repository uses [Loom](https://github.com/rjwalters/loom) for AI-powered development orchestration — ...
<!-- END LOOM ORCHESTRATION -->
```

The install reported `✓ Updated REPO-SKILLS block in CLAUDE.md` and exited 0, but the resulting diff was:

```diff
-<!-- END REPO-SKILLS --><!-- BEGIN LOOM ORCHESTRATION -->
-This repository uses [Loom](https://github.com/rjwalters/loom) for AI-powered development orchestration — ...
-<!-- END LOOM ORCHESTRATION -->
+<!-- END REPO-SKILLS -->
```

**Loom's entire orchestration block was deleted.** The marker-bounded replacement appears to operate on whole lines, so when `<!-- END REPO-SKILLS -->` shares a line with a following block's opening marker, everything from that line through the other tool's block is consumed.

Failure is silent: no warning, success message, exit 0. In a repo where CLAUDE.md is committed, this lands as unexplained deletion of another tool's config in whatever commit follows.

## Why the adjacency exists

Not hand-authored — it's what you get when two installers each append a marker block without a trailing newline, which is plausible in any repo running both Repo Skills and Loom.

## Verified fix

Separating the markers with a blank line makes the install idempotent — re-ran `install.sh -y` afterwards and Loom's block survived intact. So the data needed to do this safely is present; the replacement just needs to be bounded by the marker *string*, not the line containing it.

- [ ] Marker-bounded replacement is anchored to `<!-- END REPO-SKILLS -->` itself, not the whole line, so trailing content on that line is preserved.
- [ ] Installing over a CLAUDE.md whose END marker is adjacent to another `<!-- BEGIN ... -->` block leaves that block byte-identical.
- [ ] Regression test covering the adjacent-markers layout (both orderings: Repo Skills first, and Repo Skills second).
- [ ] If the layout can't be handled safely, `install.sh` warns loudly and declines to rewrite rather than silently deleting.

Related: #31 already covers orphaned-pointer reconciliation in `install.sh`; this is a different failure mode in the same rewrite path.

---

### Root cause — CONFIRMED in code, and reproduced

The reporter's diagnosis is correct, and the defect is present in **three** call sites, not one — all use `awk` with whole-line string equality (`$0 == begin` / `$0 == end`) to bound the skip region instead of matching the marker as a substring:

1. **`install.sh:512-516`** — the update path that produced the reported bug (`✓ Updated REPO-SKILLS block in CLAUDE.md`):
   ```awk
   $0 == begin { skip = 1; while ((getline line < block) > 0) print line; close(block); next }
   $0 == end   { skip = 0; next }
   !skip       { print }
   ```
2. **`install.sh:77-81`** (`reconcile_orphaned_block`) — same pattern, block-removal only (no re-insertion), triggered when the install destination is gitignored/dev-mode and a stale block is found.
3. **`uninstall.sh:138-142`** — same pattern, block-removal only, triggered on every `uninstall.sh` run against a CLAUDE.md that has the marker.

**Mechanism:** `$0 == end` requires the *entire line* to equal `<!-- END REPO-SKILLS -->` exactly. When the end marker is immediately followed on the same physical line by another tool's begin marker (e.g. `<!-- END REPO-SKILLS --><!-- BEGIN LOOM ORCHESTRATION -->`), that combined line does not equal either `begin` or `end` verbatim. `skip` was already set to `1` by the preceding `$0 == begin` match, and since the merged line fails the `$0 == end` test, `skip` never resets to `0`. Every subsequent line — including the entirety of the adjacent tool's block — is silently dropped by the `!skip { print }` rule, and (in the update path) nothing terminates the skip because there is no other line later in the file that equals `<!-- END REPO-SKILLS -->` verbatim. The skip runs to EOF.

**Reproduction** (ran locally against the exact adjacency from the report):

```bash
cat > CLAUDE.md <<'EOF'
Some intro content.

<!-- BEGIN REPO-SKILLS -->
OLD repo skills content line 1
OLD repo skills content line 2
removals. Managed by `install.sh` — edit outside the markers only.
<!-- END REPO-SKILLS --><!-- BEGIN LOOM ORCHESTRATION -->
This repository uses Loom for AI-powered development orchestration.
<!-- END LOOM ORCHESTRATION -->
EOF

awk -v begin='<!-- BEGIN REPO-SKILLS -->' -v end='<!-- END REPO-SKILLS -->' -v block=block.txt '
    $0 == begin { skip = 1; while ((getline line < block) > 0) print line; close(block); next }
    $0 == end   { skip = 0; next }
    !skip       { print }
  ' CLAUDE.md
```

Output:
```
Some intro content.

<!-- BEGIN REPO-SKILLS -->
NEW repo skills content v0.6.1
<!-- END REPO-SKILLS -->
```

The Loom block is gone — bit-for-bit the failure mode described in the issue.

### uninstall.sh shares the defect

Yes — confirmed. `uninstall.sh:136-145` uses the identical `$0 == begin` / `$0 == end` awk pattern with no `getline`/re-insertion (it's pure removal), so running `uninstall.sh` against the same adjacent-marker CLAUDE.md would also swallow everything after the merged line to EOF. The fix must be applied to **both** `install.sh` (two call sites: update path at ~512, and `reconcile_orphaned_block` at ~77) **and** `uninstall.sh` (~136).

### Implementation guidance

Options, in order of preference:

1. **Preferred: string-anchored replacement, not line-anchored.** Read the file as a single string (not line-by-line) and use marker-substring boundaries: everything up to and including the first occurrence of the begin-marker string, the replacement block, everything from the end-marker string onward. In `awk`, this means switching from `$0 == begin` line comparison to `index($0, begin)` and slicing the matched line at the marker boundary (preserving any prefix/suffix text sharing that line with the marker) rather than dropping the whole line. Equivalently, this can be done more robustly in `sed`/`perl` operating on the whole file as one buffer with `\0`-safe substring matching, since awk's per-line model makes "preserve trailing content past the marker on the same line" awkward to express correctly. A small helper function shared by `install.sh` and `uninstall.sh` (e.g. sourced from a common lib, or duplicated with a comment cross-referencing the other file) avoids re-introducing drift between the two call sites' logic.
2. **Defensive guard (recommended in addition to #1):** Before rewriting, detect the ambiguous case — an END marker line that is not exactly `MARKER_END` (i.e., has trailing non-whitespace content) or a BEGIN marker line that is not exactly `MARKER_BEGIN` (leading non-whitespace content). This is a strict superset of the adjacency case in the report and also catches other "shared line" variants. If ambiguous, refuse to rewrite, print a loud, explicit warning naming the file and the offending line, and exit non-zero (`install.sh`) / skip the block removal with a warning (`uninstall.sh`) rather than guessing. This matches the repo's stated principle ("apply safe fixes, gate destructive ones") — consuming another tool's content without confirmation is a destructive action and should never happen silently.
3. **Backup before rewrite (defense in depth):** Before overwriting `CLAUDE.md` in any of the three call sites, copy it to `CLAUDE.md.bak` (or a `mktemp` snapshot surfaced in the warning/error message) so a botched rewrite is recoverable without `git checkout` (useful for gitignored/dev-mode `CLAUDE.md`, or before the first commit). This is cheap insurance and should be considered independent of whether #1/#2 land — it does not replace the anchoring fix, since a byte-correct backup of a corrupted result still requires human recovery, but it makes that recovery trivial.

None of these are mutually exclusive; #1 fixes the root cause, #2 prevents recurrence for adjacency patterns not explicitly enumerated, #3 makes any residual failure cheaply recoverable.

### Acceptance criteria (supersedes/refines the "Suggested acceptance criteria" above)

- [ ] Marker-bounded replacement in `install.sh` (both the update path at ~512 and `reconcile_orphaned_block` at ~77) and `uninstall.sh` (~136) is anchored to the marker **string**, not the line containing it — trailing/leading content sharing a line with a marker is preserved byte-for-byte.
- [ ] Regression test reproduces the exact adjacency from this report — `<!-- END REPO-SKILLS --><!-- BEGIN LOOM ORCHESTRATION -->` on one physical line, no newline between — and asserts the Loom block (and any content after it) survives `install.sh` (fresh install over existing block) byte-identical.
- [ ] Same regression test (or a sibling case) covers the reverse ordering: `<!-- END LOOM ORCHESTRATION --><!-- BEGIN REPO-SKILLS -->`, i.e. Repo Skills' own begin marker adjacent to a preceding tool's end marker.
- [ ] Regression test covers `uninstall.sh` against the same adjacent-marker fixture: content after the REPO-SKILLS block must survive removal byte-identical.
- [ ] Regression test covers `reconcile_orphaned_block`'s removal path (`install.sh` dev-mode / gitignored-destination code path) against the same adjacent-marker fixture.
- [ ] If the ambiguous-adjacency case (or any case where marker/line boundaries can't be unambiguously resolved) is detected and the fix chooses to refuse rather than guess (Option 2 above), the tool prints a clear warning naming the file and declines to rewrite `CLAUDE.md`, and exits non-zero for `install.sh`. Silent success (exit 0, "Updated"/"Removed" message) must never occur when another tool's content would be altered or dropped.
- [ ] Existing non-adjacent cases (normal install, normal uninstall, orphan reconciliation with markers on their own lines) continue to pass unchanged — this is a correctness fix, not a behavior change for the common case.
- [ ] `npm test` (`./hooks/repo/tests/run.sh`) passes, and the new test file is either added to that runner or documented as a separate entry point if kept standalone (see Affected Files note below — there is currently no test harness for `install.sh`/`uninstall.sh` in this repo, so the builder needs to establish one).

### Affected Files

- `install.sh` — lines ~77-81 (`reconcile_orphaned_block`) and ~510-518 (CLAUDE.md update path): replace whole-line `$0 == begin` / `$0 == end` awk matching with marker-substring-anchored logic; add the ambiguous-adjacency guard.
- `uninstall.sh` — lines ~136-145: same fix, mirrored (removal-only, no re-insertion).
- New test file, e.g. `hooks/repo/tests/test-install-claude-md-markers.sh` (or `tests/install-claude-md-markers.sh` at repo root, matching whichever convention the builder finds cleaner) — regression tests for the adjacent-marker cases above. **Note for the builder:** there is currently no existing test suite that exercises `install.sh`/`uninstall.sh` directly anywhere in this repo (`npm test` → `./hooks/repo/tests/run.sh` only covers the PreToolUse hooks; `.loom/scripts/tests/*` cover Loom's own self-hosted `install-loom.sh`/`create-pr.sh`, a different tool). This issue's test needs a new file and, ideally, wiring into `hooks/repo/tests/run.sh` (or a new `npm test` script entry) so it runs in the same place other repo-level tests do.

### Test Plan

- [ ] Automated tests: new regression test file (see Affected Files) covering:
  - Adjacent markers, Repo Skills END immediately followed by another tool's BEGIN on one line, no newline — `install.sh` update path must preserve the other block byte-identical.
  - Adjacent markers, reverse ordering (other tool's END immediately followed by Repo Skills' BEGIN on one line) — `install.sh` fresh-append and update paths must not corrupt the preceding block.
  - Same adjacency fixture run through `uninstall.sh` — content after the removed block must survive byte-identical.
  - Same adjacency fixture run through `reconcile_orphaned_block`'s removal path (dev-mode / gitignored-destination install.sh code path).
  - Non-adjacent (markers on their own lines) cases for all three call sites continue to pass, confirming no regression to the common case.
  - If the guard-and-refuse approach (Option 2) is implemented: a case asserting `install.sh` exits non-zero with a clear warning naming the file, and does not modify `CLAUDE.md`, when the ambiguous layout is detected.
- [ ] Manual verification: run the exact reproduction from this issue's Curator Enhancement section (scratch `CLAUDE.md` with the reported adjacency) through `install.sh -y <target>` before and after the fix; confirm the Loom-like block survives.
- [ ] Edge cases: trailing whitespace after a marker on its own line (should still match cleanly — not itself ambiguous), multiple marker blocks from three or more tools chained on adjacent lines, and a marker string appearing as literal text inside a code fence elsewhere in `CLAUDE.md` (should not be treated as a real marker — verify existing behavior isn't further degraded by the fix, even if not in scope to fully solve here).

### Cross-references

- Related: #31 (closed) — covered orphaned-pointer reconciliation (`reconcile_orphaned_block`) for a different failure mode (stale version claims when the destination goes gitignored) in the same rewrite path. This issue is a distinct data-loss bug in the same function plus two siblings.

rjwalters:
## Summary

`install.sh` silently deleted an adjacent tool's CLAUDE.md marker block — it destroyed Loom's entire orchestration block in `rjwalters/vibesql` during a routine 0.4.3 → 0.6.1 upgrade, printed `✓ Updated REPO-SKILLS block in CLAUDE.md`, and exited 0.

All three rewrite sites bounded the block with awk whole-line equality (`$0 == begin` / `$0 == end`). When two installers each append without a trailing newline you get

```
<!-- END REPO-SKILLS --><!-- BEGIN LOOM ORCHESTRATION -->
```

on one physical line. That merged line never compares equal to the end marker, so `skip` never clears and every byte through EOF is dropped.

This replaces all three sites with marker-**string**-anchored slicing, refuses to rewrite layouts it cannot resolve, backs up before rewriting, and establishes the first test harness for `install.sh`/`uninstall.sh`.

## Changes

- **`lib/claude-md-block.sh` (new)** — shared, sourced-only helper. `claude_md_block_rewrite()` slices the file on the marker strings, so text sharing a physical line with a marker survives byte-for-byte. Removal still consumes the end marker's trailing newline when the begin marker started its own line, so whole-line removal (the historical behavior) is unchanged for the common case. `claude_md_block_validate()` gates it.
- **`install.sh`** — both sites (update path ~512, `reconcile_orphaned_block` ~77) now call the helper. On an unresolvable layout: loud warning naming the file and the reason, CLAUDE.md untouched, **exit non-zero**.
- **`uninstall.sh`** — same helper (removal mode); on an unresolvable layout it warns and skips the removal without ever printing "Removed".
- **Backup before rewrite** — every rewrite snapshots CLAUDE.md to a temp file first and reports the path (`ℹ Backed up CLAUDE.md to …`). Kept out of the consumer's working tree so no untracked `CLAUDE.md.bak` litter appears.
- **`hooks/repo/tests/test-install-claude-md-markers.sh` (new, 49 cases)** — drives the real installers against scratch git repos; conventions follow `test-guard-destructive.sh` / `test-session-start-handoff.sh` (pure bash, scratch dirs, PASS/FAIL counters).
- **`hooks/repo/tests/run.sh`** — one additive delegation block, matching the handoff-suite pattern, so `pnpm test` covers it.
- **`README.md`** — two lines in the repository-layout block.

| Replacement anchored to the marker string at all three sites; line-sharing content preserved byte-for-byte | ✅ | `lib/claude-md-block.sh` used by all three; `assert_bytes` diffs the neighbour block file-to-file (catches even a missing trailing newline) |
| Exact reported adjacency (`END REPO-SKILLS` + `BEGIN LOOM ORCHESTRATION`, one line) leaves the other block byte-identical | ✅ | `adjacent-after:` cases + manual repro: the resulting `git diff` in a scratch consumer repo touches only our block |
| Reverse ordering (`END LOOM` + `BEGIN REPO-SKILLS`) covered | ✅ | `adjacent-before:` cases (update path) and `append:` cases (fresh-append path) |
| `uninstall.sh` against the adjacency fixture: content after the block survives byte-identical | ✅ | `uninstall adjacent:` cases |
| `reconcile_orphaned_block` removal path against the adjacency fixture | ✅ | `reconcile adjacent:` cases (gitignored `.claude/` destination triggers the real code path) |
| Ambiguous layout → clear warning naming the file, declines to rewrite, non-zero exit for install.sh; never a silent success | ✅ | `guard unterminated` / `guard duplicate` / `guard uninstall` cases assert exit status, warning text, absence of the success message, and that CLAUDE.md is byte-unchanged |
| Non-adjacent cases unchanged (install, uninstall, reconcile) | ✅ | `non-adjacent:` cases pin the exact post-removal bytes, locking the historical whole-line behavior |
| Test wired into `pnpm test` | ✅ | `./hooks/repo/tests/run.sh` → `ok test-install-claude-md-markers.sh -> 49 cases pass` |

## Test Plan

```
$ ./hooks/repo/tests/run.sh
  ok   test-session-start-handoff.sh    -> 79 cases pass
  ok   test-install-claude-md-markers.sh -> 49 cases pass
  PASS: 57   FAIL: 0          (exit 0)

$ ./hooks/repo/tests/test-guard-destructive.sh
  Total: 444   Passed: 444   Failed: 0   (exit 0)
```

Negative control: with `install.sh`/`uninstall.sh` stashed back to their pre-fix state (new lib + suite retained), **18 of the 49 cases fail** — the suite genuinely reproduces the bug rather than merely describing it.

Manual verification of the exact scenario from the report (scratch git repo, committed CLAUDE.md with the reported adjacency, `install.sh -y`): the Loom block is untouched and the diff contains only the REPO-SKILLS version bump.

## Notes / out of scope

- A marker string appearing as literal prose inside a code fence is not distinguished from a real marker; it now trips the duplicate-marker guard and is **refused** rather than silently mangled. Fully solving fence-awareness was explicitly out of scope.
- `install.sh` comment wording unrelated to this fix is untouched (#37 owns that).
- Pre-existing shellcheck `SC2094`/`SC2016` findings in `install.sh` / `run.sh` are left alone; no new findings were introduced.

Closes #38

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,8 +81,10 @@ commands/repo/*.md           Command files installed to .claude/commands/repo/
 hooks/repo/guard-destructive.sh  PreToolUse guard hook installed to .claude/skills/repo/hooks/
 hooks/repo/tests/run.sh      Quick smoke suite for the guard hook (bash, no framework needed)
 hooks/repo/tests/test-guard-destructive.sh  Full guard regression suite (ported from Loom)
+hooks/repo/tests/test-install-claude-md-markers.sh  CLAUDE.md marker-block regression suite
 install.sh                   Installer
 uninstall.sh                 Uninstaller
+lib/claude-md-block.sh       Marker-bounded CLAUDE.md surgery shared by install.sh/uninstall.sh
 ```
 
 ## Adding a skill
--- a/hooks/repo/tests/run.sh
+++ b/hooks/repo/tests/run.sh
@@ -182,6 +182,26 @@ else
     printf '%s\n' "$SS_OUT" | sed 's/^/    /'
 fi
 
+# install.sh / uninstall.sh CLAUDE.md marker-block surgery (repo#38). Same
+# delegation shape as the handoff suite above: it drives the installers against
+# scratch git repos rather than piping hook payloads, so it shares nothing with
+# expect() and folds in as a single case.
+echo
+echo "-- install/uninstall CLAUDE.md markers (delegated suite) --"
+MD_TEST="$TESTS_DIR/test-install-claude-md-markers.sh"
+if [[ ! -f "$MD_TEST" ]]; then
+    FAIL=$((FAIL + 1))
+    printf '  FAIL %-52s -> not found at %s\n' "test-install-claude-md-markers.sh" "$MD_TEST"
+elif MD_OUT="$(bash "$MD_TEST" 2>&1)"; then
+    PASS=$((PASS + 1))
+    printf '  ok   %-52s -> %s\n' "test-install-claude-md-markers.sh" \
+        "$(printf '%s\n' "$MD_OUT" | awk '/^  Total:/ {print $2 " cases pass"}')"
+else
+    FAIL=$((FAIL + 1))
+    printf '  FAIL %-52s -> suite failed; output follows\n' "test-install-claude-md-markers.sh"
+    printf '%s\n' "$MD_OUT" | sed 's/^/    /'
+fi
+
 echo
 echo "==============================="
 echo "PASS: $PASS   FAIL: $FAIL"
--- a/hooks/repo/tests/test-install-claude-md-markers.sh
+++ b/hooks/repo/tests/test-install-claude-md-markers.sh
@@ -1,0 +1,455 @@
+#!/usr/bin/env bash
+# Regression suite for marker-bounded CLAUDE.md surgery in install.sh /
+# uninstall.sh (lib/claude-md-block.sh).
+#
+# Usage: ./hooks/repo/tests/test-install-claude-md-markers.sh
+# Exit code 0 = all tests pass, 1 = failures detected.
+#
+# Structured like test-guard-destructive.sh / test-session-start-handoff.sh next
+# door: pure bash, no test framework, PASS/FAIL counters and a summary block.
+#
+# The bug under test (repo#38): all three call sites bounded the REPO-SKILLS
+# block with awk whole-line equality, so a CLAUDE.md written as
+#
+#     <!-- END REPO-SKILLS --><!-- BEGIN LOOM ORCHESTRATION -->
+#
+# (no newline between two tools' blocks — what you get when two installers each
+# append without a trailing newline) made the end marker never compare equal,
+# leaving the skip flag set and silently dropping every byte to EOF. An upgrade
+# deleted Loom's entire orchestration block, printed "✓ Updated", and exited 0.
+#
+# The contract asserted below:
+#   - both adjacency orderings survive byte-identical, at all three call sites
+#     (install update path, reconcile_orphaned_block, uninstall)
+#   - the non-adjacent common case is unchanged
+#   - an unresolvable marker layout is refused loudly, leaves CLAUDE.md
+#     untouched, and exits non-zero from install.sh
+
+set -uo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+REPO_ROOT="$(cd "$SCRIPT_DIR/../../.." && pwd)"
+INSTALL_SH="$REPO_ROOT/install.sh"
+UNINSTALL_SH="$REPO_ROOT/uninstall.sh"
+VERSION="$(cat "$REPO_ROOT/VERSION" 2>/dev/null || echo unknown)"
+
+PASS=0
+FAIL=0
+TOTAL=0
+
+RED='\033[0;31m'
+GREEN='\033[0;32m'
+NC='\033[0m'
+
+for f in "$INSTALL_SH" "$UNINSTALL_SH" "$REPO_ROOT/lib/claude-md-block.sh"; do
+    if [[ ! -f "$f" ]]; then
+        echo "FATAL: required script not found at $f" >&2
+        exit 1
+    fi
+done
+
+SCRATCH="$(cd "$(mktemp -d)" && pwd -P)"
+trap 'rm -rf "$SCRATCH"' EXIT
+
+# ---------------------------------------------------------------------------
+# Assertion helpers
+# ---------------------------------------------------------------------------
+
+ok() {   # <label>
+    TOTAL=$((TOTAL + 1)); PASS=$((PASS + 1))
+    printf "  ${GREEN}PASS${NC}: %s\n" "$1"
+}
+no() {   # <label> <detail>
+    TOTAL=$((TOTAL + 1)); FAIL=$((FAIL + 1))
+    printf "  ${RED}FAIL${NC}: %s\n" "$1"
+    [[ -n "${2:-}" ]] && printf "%s\n" "$2" | sed 's/^/        /'
+    return 0
+}
+assert_eq() {  # <label> <expected> <actual>
+    if [[ "$2" == "$3" ]]; then ok "$1"; else no "$1" "expected [$2], got [$3]"; fi
+}
+assert_contains() {  # <label> <haystack> <needle>
+    if [[ "$2" == *"$3"* ]]; then ok "$1"; else no "$1" "missing [$3]"; fi
+}
+assert_not_contains() {  # <label> <haystack> <needle>
+    if [[ "$2" != *"$3"* ]]; then ok "$1"; else no "$1" "unexpectedly present: [$3]"; fi
+}
+# Byte-exact file comparison. Deliberately file-to-file rather than
+# string-to-file: `$(...)` strips trailing newlines, and whether the rewritten
+# CLAUDE.md still ends in one is exactly the kind of byte the fix must preserve.
+assert_bytes() {  # <label> <actual-file> <expected-file>
+    if diff -u "$3" "$2" >/dev/null 2>&1; then
+        ok "$1"
+    else
+        no "$1" "$(diff -u "$3" "$2" | head -25)"
+    fi
+}
+
+# ---------------------------------------------------------------------------
+# Fixture helpers
+# ---------------------------------------------------------------------------
+
+# Slurp a file for substring assertions. Note that `$(slurp f)` still drops
+# trailing newlines — that is fine for assert_contains/assert_eq, where both
+# sides are stripped alike, but never use it for byte-exactness (assert_bytes
+# compares files directly for exactly that reason).
+slurp() {  # <file>
+    local c
+    c="$(cat "$1"; printf 'X')"
+    printf '%s' "${c%X}"
+}
+
+# Everything from the first occurrence of <needle> to EOF (needle included).
+# Substring-based, not line-based, because in the adjacency fixtures the needle
+# does not start its physical line. Redirect to a file and compare with
+# assert_bytes; the sentinel keeps the trailing newline intact.
+tail_from() {  # <file> <needle> -> stdout
+    local c=""
+    c="$(cat "$1"; printf 'X')"; c="${c%X}"
+    printf '%s%s' "$2" "${c#*"$2"}"
+}
+
+# Everything from the start of the file through <needle> (needle included).
+head_through() {  # <file> <needle> -> stdout
+    local c=""
+    c="$(cat "$1"; printf 'X')"; c="${c%X}"
+    printf '%s%s' "${c%%"$2"*}" "$2"
+}
+
+new_target() {  # <name> -> prints the target path
+    local t="$SCRATCH/$1"
+    mkdir -p "$t"
+    git init -q "$t" 2>/dev/null
+    printf '%s' "$t"
+}
+
+# uninstall.sh bails early unless an install is present; these tests care only
+# about the CLAUDE.md surgery, so fake the directory it looks for.
+fake_installed() {  # <target>
+    mkdir -p "$1/.claude/skills/repo"
+}
+
+# ===========================================================================
+echo "install.sh / uninstall.sh CLAUDE.md marker suite"
+echo "================================================"
+
+# ---------------------------------------------------------------------------
+echo ""
+echo "-- install.sh update path: our END adjacent to another tool's BEGIN (repo#38) --"
+
+T1="$(new_target adjacent-after)"
+cat > "$T1/CLAUDE.md" <<'EOF'
+Some intro content.
+
+<!-- BEGIN REPO-SKILLS -->
+OLD repo skills content line 1
+removals. Managed by `install.sh` — edit outside the markers only.
+<!-- END REPO-SKILLS --><!-- BEGIN LOOM ORCHESTRATION -->
+This repository uses Loom for AI-powered development orchestration.
+<!-- END LOOM ORCHESTRATION -->
+
+Trailing prose that must survive.
+EOF
+cat > "$SCRATCH/expected-loom-tail" <<'EOF'
+<!-- BEGIN LOOM ORCHESTRATION -->
+This repository uses Loom for AI-powered development orchestration.
+<!-- END LOOM ORCHESTRATION -->
+
+Trailing prose that must survive.
+EOF
+
+OUT1="$(bash "$INSTALL_SH" -y "$T1" 2>&1)"; RC1=$?
+BODY1="$(slurp "$T1/CLAUDE.md")"
+
+assert_eq        "adjacent-after: install exits 0"                "0" "$RC1"
+tail_from "$T1/CLAUDE.md" '<!-- BEGIN LOOM ORCHESTRATION -->' > "$SCRATCH/actual-loom-tail"
+assert_bytes     "adjacent-after: Loom block + tail byte-identical" \
+                 "$SCRATCH/actual-loom-tail" "$SCRATCH/expected-loom-tail"
+assert_contains  "adjacent-after: content before the block survives" "$BODY1" "Some intro content."
+assert_contains  "adjacent-after: block was refreshed"            "$BODY1" "Repo Skills](https://github.com/rjwalters/repo) v$VERSION"
+assert_not_contains "adjacent-after: stale block body is gone"    "$BODY1" "OLD repo skills content line 1"
+assert_contains  "adjacent-after: adjacency preserved verbatim"   "$BODY1" '<!-- END REPO-SKILLS --><!-- BEGIN LOOM ORCHESTRATION -->'
+assert_contains  "adjacent-after: reports the update"             "$OUT1" "Updated REPO-SKILLS block in CLAUDE.md"
+assert_contains  "adjacent-after: reports a backup snapshot"      "$OUT1" "Backed up CLAUDE.md to "
+
+# Re-running must be a no-op on the neighbour's block (idempotence).
+BEFORE_RERUN="$(slurp "$T1/CLAUDE.md")"
+bash "$INSTALL_SH" -y "$T1" >/dev/null 2>&1
+assert_eq "adjacent-after: re-install is idempotent" "$BEFORE_RERUN" "$(slurp "$T1/CLAUDE.md")"
+
+# ---------------------------------------------------------------------------
+echo ""
+echo "-- install.sh update path: another tool's END adjacent to our BEGIN --"
+
+T2="$(new_target adjacent-before)"
+cat > "$T2/CLAUDE.md" <<'EOF'
+Preamble that must survive.
+
+<!-- BEGIN LOOM ORCHESTRATION -->
+This repository uses Loom for AI-powered development orchestration.
+<!-- END LOOM ORCHESTRATION --><!-- BEGIN REPO-SKILLS -->
+OLD repo skills content line 1
+<!-- END REPO-SKILLS -->
+
+Trailing prose that must survive.
+EOF
+# No trailing newline: head_through() stops at the marker itself.
+printf '%s' 'Preamble that must survive.
+
+<!-- BEGIN LOOM ORCHESTRATION -->
+This repository uses Loom for AI-powered development orchestration.
+<!-- END LOOM ORCHESTRATION -->' > "$SCRATCH/expected-loom-head"
+
+OUT2="$(bash "$INSTALL_SH" -y "$T2" 2>&1)"; RC2=$?
+BODY2="$(slurp "$T2/CLAUDE.md")"
+
+assert_eq    "adjacent-before: install exits 0" "0" "$RC2"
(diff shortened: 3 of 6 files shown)


rjwalters:
## ✅ Approved

Reviewed against issue #38's Curator acceptance criteria. Every builder claim was re-verified independently rather than taken on trust — including a negative control against the pre-fix installers and a from-scratch reproduction of the reported vibesql scenario.

### 1. All three defect sites route through the shared helper

`lib/claude-md-block.sh` is sourced by both scripts (`install.sh:60`, `uninstall.sh:23`) and all three legacy `$0 == begin` / `$0 == end` awk blocks are gone:

| Site | Before | After |
|---|---|---|
| `install.sh` `reconcile_orphaned_block` (~77) | whole-line awk, removal-only | `claude_md_block_rewrite "$TARGET/CLAUDE.md" "$MARKER_BEGIN" "$MARKER_END"` |
| `install.sh` update path (~513) | whole-line awk + `getline` re-insert | `claude_md_block_rewrite ... "$BLOCK_FILE"` |
| `uninstall.sh` (~146) | whole-line awk, removal-only | `claude_md_block_rewrite ... ` (warn + skip on refusal, non-fatal) |

`grep -n '\$0 == begin' install.sh uninstall.sh` returns nothing on this branch. No site was missed.

### 2. Negative control — the suite genuinely reproduces the bug

Rebuilt the branch tree into a scratch dir, then overwrote `install.sh`/`uninstall.sh` with `git show main:` (pre-fix) versions and ran the *new* suite unchanged:

```
=== confirming pre-fix awk present ===
negctl/uninstall.sh:139:    $0 == begin { skip = 1; next }
negctl/install.sh:78:      $0 == begin { skip = 1; next }
negctl/install.sh:513:    $0 == begin { skip = 1; while ((getline line < block) > 0) print line; ... }

  Total:  49
  Passed: 31
  Failed: 18

TESTS FAILED
```

**18/49 fail against the buggy code — the builder's number is exact.** The failing cases:

```
FAIL: adjacent-after: Loom block + tail byte-identical
FAIL: adjacent-after: adjacency preserved verbatim
FAIL: adjacent-after: reports a backup snapshot
FAIL: adjacent-before: block was refreshed
FAIL: adjacent-before: stale block body is gone
FAIL: uninstall adjacent: everything but our block survives byte-identical
FAIL: reconcile adjacent: neighbour + tail survive byte-identical
FAIL: guard unterminated: install exits non-zero
FAIL: guard unterminated: warning names the file
FAIL: guard unterminated: explains the refusal
FAIL: guard unterminated: never claims success
FAIL: guard unterminated: CLAUDE.md untouched
FAIL: guard duplicate: install exits non-zero
FAIL: guard duplicate: reports the occurrence count
FAIL: guard duplicate: CLAUDE.md untouched
FAIL: guard uninstall: warning names the file
FAIL: guard uninstall: never claims removal
FAIL: guard uninstall: CLAUDE.md untouched
```

All three call sites are represented (`adjacent-after` = update path, `reconcile adjacent`, `uninstall adjacent`). This is a regression suite, not a description of one.

### 3. End-to-end reproduction of the reported vibesql scenario

Built a scratch git repo with the exact adjacency (`printf '%s' '<!-- END REPO-SKILLS -->'` with no newline, then the Loom block appended).

**Pre-fix baseline — the reported data loss, verbatim:**

```
✓ Updated REPO-SKILLS block in CLAUDE.md
exit=0

-<!-- END REPO-SKILLS --><!-- BEGIN LOOM ORCHESTRATION -->
-This repository uses [Loom](https://github.com/rjwalters/loom) for AI-powered development orchestration — ...
-<!-- END LOOM ORCHESTRATION -->
-
-Tail prose after both blocks.
+<!-- END REPO-SKILLS -->
```

**With this PR — only the REPO-SKILLS block changes:**

```
 <!-- BEGIN REPO-SKILLS -->
-This repository has [Repo Skills](https://github.com/rjwalters/repo) v0.4.3 installed —
+This repository has [Repo Skills](https://github.com/rjwalters/repo) v0.6.1 installed —
+general repository hygiene and environment commands invoked as `/repo:<command>`. Run
+... (4 more pointer lines)
 removals. Managed by `install.sh` — edit outside the markers only.
 <!-- END REPO-SKILLS --><!-- BEGIN LOOM ORCHESTRATION -->
 This repository uses [Loom](https://github.com/rjwalters/loom) for AI-powered development orchestration — ...
```

md5 of the region from `<!-- BEGIN LOOM ORCHESTRATION -->` to EOF: `f41c8858...` before and after — **byte-identical**. The adjacency itself is preserved verbatim.

**`uninstall.sh` on the same fixture** — verified byte-exactly in Python against `git show HEAD:CLAUDE.md`:

```
loom tail byte-identical after uninstall: True
prefix before block byte-identical: True
```

The only structural change is the Loom BEGIN marker relocating to column 0 of its own line (the REPO-SKILLS span it shared the line with is gone) — correct, and no Loom bytes are lost.

### 4. Guard-and-refuse scoping is correct

`claude_md_block_validate` refuses on exactly three conditions: begin-count ≠ 1, end-count ≠ 1, end-before-begin. **The reported adjacency satisfies all three checks and is rewritten, not refused** — confirmed by the end-to-end run above and by the `adjacent-after` / `adjacent-before` cases. A normal non-adjacent install is likewise unaffected (`non-adjacent` case group + differential test below). No existing consumer is broken.

Refusal never prints success: `install.sh` funnels through `error()` (`exit 1`), which fires *before* both `success "Updated REPO-SKILLS block"` and the final `success "Repo Skills v$VERSION installed"`. Pinned by `guard unterminated: never claims success` + `install exits non-zero` (observed rc=1) and `guard duplicate: install exits non-zero`. `uninstall.sh` correctly warns and skips without exiting non-zero (removal is best-effort there) and never prints `Removed REPO-SKILLS block`.

### 5. Newline handling preserves the historical case byte-exactly

Ran a differential harness sourcing `claude_md_block_rewrite` and comparing against the legacy awk removal on non-adjacent fixtures:

```
case 1 (block mid-file, blank lines around):  byte-identical to legacy awk
case 2 (block at start of file):              byte-identical to legacy awk
case 3 (block at EOF, trailing newline):      byte-identical to legacy awk
case 4 (block at EOF, no trailing newline):   byte-identical to legacy awk
case 5 (trailing whitespace after END marker): DIFFERS
```

Case 5 is the *intended* divergence: legacy awk failed `$0 == end` on `<!-- END REPO-SKILLS -->   ` and silently dropped everything to EOF (output `a\n`); the new code correctly yields `a\nb\n`. That is issue #38's own edge-case note ("trailing whitespace after a marker on its own line should still match cleanly — not itself ambiguous"), handled by the `rest_of_line` whitespace test at `lib/claude-md-block.sh:144`.

Both branches are pinned by byte-exact `diff` assertions in the suite: `uninstall non-adjacent: whole-line removal (historical behavior)` and `reconcile non-adjacent: whole-line removal (historical behavior)` for the common case, `uninstall adjacent` / `reconcile adjacent: neighbour + tail survive byte-identical` for the adjacent case.

### Test execution

| Check | Result |
|---|---|
| `bash hooks/repo/tests/test-install-claude-md-markers.sh` | **49/49 pass** |
| `pnpm test` (`hooks/repo/tests/run.sh`) | **PASS: 57  FAIL: 0** (new suite folds in as 1 delegated case) |
| `bash hooks/repo/tests/test-guard-destructive.sh` (pre-existing) | **444/444 pass**, unchanged |
| `pnpm check:ci` | no-op stub (repo has no configured CI) |
| `shellcheck -x` on all changed shell files | info-level only: `SC2059` in the new test file (matches sibling suites' style), `SC2094` pre-existing in `install.sh` |

### Safety checks

- **No writes outside the target repo.** The suite confines everything to `mktemp -d` scratch git repos; the source worktree was byte-clean after the full run.
- **Backups go to `mktemp`, not `CLAUDE.md.bak`.** Observed: `ℹ Backed up CLAUDE.md to /var/folders/.../CLAUDE.md.backup.WKZvHy before rewriting`. Consumer repo after install/uninstall contained no `.bak` litter.
- **`lib/` availability**: `install.sh`/`uninstall.sh` source from `SOURCE_ROOT`, and both are only ever invoked from a clone of this repo (README + `update-tools.md` document no curl-pipe path), so the new dependency is always satisfied.
- **`Closes #38`** present in the PR body (line 60). Merge state `CLEAN`.

### ⚠️ Merge-conflict warning on `hooks/repo/tests/run.sh` (reported, not fixed here)

This PR's `run.sh` edit is genuinely minimal and purely additive — **+20 lines, 0 deletions**, a single self-contained delegation block copied from the adjacent session-start block. It touches no existing line. That is the right shape.

It will still conflict, because both siblings in this wave touch the same region:

- **vs #41** — near-certain textual conflict. #41's hunk header is `@@ -182,6 +182,26 @@`, byte-identical to this PR's insertion anchor (immediately after the session-start `fi`, before the final summary). Two additions at the same point with the same context. **Resolution is mechanical: keep both blocks.**
- **vs #40** — likely conflict. #40 rewrites the delegation machinery on both sides of this insertion point: it adds `record_suite` calls inside the session-start `if/elif/else` directly above, and replaces the trailing summary block directly below. This PR's block lands between two modified hunks.
- **Semantic follow-up if #40 lands first**: #40 introduces `strip_ansi` / `suite_count` / `record_suite` so delegated suites fold their *real* counts into totals. This PR's block uses the older inline-`awk` shape and contributes `1` to PASS rather than `49`, and won't appear in #40's per-suite breakdown. Whoever rebases should convert it. Not a defect in this PR — it was written against current `main`.

No conflict risk on any other file; `lib/claude-md-block.sh` and the test file are new, and the `install.sh`/`uninstall.sh`/`README.md` edits are unique to this PR.

### Non-blocking notes

1. **Widened refusal surface.** A consumer `CLAUDE.md` that quotes `<!-- BEGIN REPO-SKILLS -->` as literal prose or inside a code fence now counts as a duplicate marker and hard-fails the install with exit 1. Previously it would silently mangle or no-op. The loud failure is the better behavior and matches the issue's stated principle, but it is a behavior change for such repos — worth a line in the release notes when this ships.
2. **Partial install on refusal.** `reconcile_orphaned_block`'s `error()` aborts after skills/commands are already installed. The message says so explicitly ("skills and commands were installed"), so this is handled, just noting the state is intentionally partial.
3. **Coverage gap.** `claude_md_block_validate`'s out-of-order branch (`'END' appears before 'BEGIN'`) has no test — the guard cases cover missing-end and duplicated-begin only. Cheap to add; not worth a round-trip.

Root cause is fixed at all three sites, the fix is proven by a reproducing negative control, and the reported production scenario is verified byte-exact end-to-end. Approving.

PR #47, opened 2026-07-28
docs(install): correct the SessionStart matcher rationale

### LINKED ISSUE #37: docs(install.sh): SessionStart matcher comment claims alternation is unsupported — it is supported

rjwalters:
## What

`install.sh` wires the SessionStart handoff hook as two literal matcher entries (`startup`, `resume`). The comment explaining that choice was factually wrong:

> alternation is documented for tool matchers but NOT for SessionStart, and a matcher that silently matched nothing would leave the hook inert with no error

Alternation **is** supported — Claude Code's matcher comparator takes the split-on-`|` branch for SessionStart the same way it does for tool matchers, so `"startup|resume"` would have worked. This was established by binary inspection during review of PR #34.

The wiring itself is correct and unchanged. Only the rationale was wrong — and a comment that would talk a future maintainer out of a form that actually works is worse than no comment at all.

## Scope

Comment text only. `SESSIONSTART_SOURCES=(startup resume)` and the jq merge loop are untouched.

## Verification

- **Byte-identical output**: installed into scratch repos from `main` and from this branch; emitted `.claude/settings.json` is identical on both sides (`md5 af1d91a169fe1afae924a2d200061482`). Emitted matchers remain `"startup"`, `"resume"`.
- `./hooks/repo/tests/run.sh` → **PASS: 532 FAIL: 0** (includes the 444-case guard suite and the 31-case branches suite)
- `shellcheck -S warning install.sh` → clean

Closes #37

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -195,11 +195,12 @@ HOOK_CMD="\${CLAUDE_PROJECT_DIR}/${HOOK_INSTALL_REL}"
 # The SessionStart handoff hook, same resolution. Claude Code's SessionStart
 # matchers key on the session SOURCE ("startup" | "resume" | "clear" |
 # "compact" | "fork"), not on a tool name. We wire the two literal sources the
-# hook acts on rather than an alternation like "startup|resume": alternation is
-# documented for tool matchers but NOT for SessionStart, and a matcher that
-# silently matched nothing would leave the hook inert with no error. Two literal
-# entries use only documented values. The hook re-checks .source itself, so this
-# is belt-and-braces rather than a load-bearing filter.
+# hook acts on rather than an alternation like "startup|resume". Alternation
+# would work — SessionStart matchers are split on "|" the same way tool matchers
+# are — but two literal entries state the intent without depending on that
+# shared behavior, and each entry reads unambiguously in a consumer's
+# settings.json. The hook re-checks .source itself, so this is belt-and-braces
+# rather than a load-bearing filter.
 SESSIONSTART_HOOK_INSTALL_REL=".claude/skills/repo/hooks/session-start-handoff.sh"
 SESSIONSTART_HOOK_CMD="\${CLAUDE_PROJECT_DIR}/${SESSIONSTART_HOOK_INSTALL_REL}"
 SESSIONSTART_SOURCES=(startup resume)


rjwalters:
## Judge review — APPROVED

Independently re-verified all claims in this PR rather than taking the description at face value.

**1. Is the new comment true?** Yes, and I confirmed it directly from the Claude Code v2.1.212 binary (the installed cask version matches), not from docs prose.

- `strings`-extracted the hook matcher comparator: `h4_(e,t,r,n)`, called as `h4_(a, k.matcher, l, c)` where `l = p4_.has(n.hook_event_name)`.
- `p4_` is a `Set` literal containing, among others, `"PreToolUse"`, `"PostToolUse"`, **`"SessionStart"`**, `"SessionEnd"`, `"Setup"`, etc. — confirmed the exact set membership by string-searching the binary for the `p4_=new Set([...])` literal.
- Inside `h4_`, the split step is `t.split(r ? /[|,]/ : "|")` — i.e. it splits on `"|"` in *both* branches; the `r` flag (true for events in `p4_`, which includes `SessionStart`) only extends this to also accept `,` and permits spaces/hyphens in the validation regex. So `SessionStart` matchers take literally the same regex-validate → split-on-`|` → `.includes()` path that tool matchers (`PreToolUse`, etc.) take.
- This directly falsifies the *old* comment's claim ("documented for tool matchers but NOT for SessionStart") and directly confirms the *new* comment's claim ("Alternation would work — SessionStart matchers are split on `|` the same way tool matchers are").

Confidence: high — this is a first-party read of the shipped comparator function and the event allow-set, not inference from behavior or docs.

**2. Zero behavioral change.** Built both revisions in separate worktrees, installed each into a fresh scratch git repo with `bash install.sh -y <target>`, and diffed the emitted `.claude/settings.json`:

```
MD5 (pre)  = af1d91a169fe1afae924a2d200061482
MD5 (post) = af1d91a169fe1afae924a2d200061482
diff → no output (identical)
```

Matches the md5 claimed in the PR description exactly.

**3. Scope discipline.** `git diff main...feature/issue-37 -- install.sh` shows only the comment block changed (6 insertions, 5 deletions, single hunk at the SessionStart comment). `SESSIONSTART_SOURCES=(startup resume)` and the jq wiring loop are byte-for-byte untouched. `+6/-5`, one file — matches the stated diff stat.

**4. Tests.** `bash hooks/repo/tests/run.sh` → `PASS: 532   FAIL: 0` (includes the 444-case guard-destructive suite and the 79-case session-start-handoff suite). `shellcheck -S warning install.sh` → clean.

**5. Comment quality.** The replacement is a genuine improvement: it states the verified-true fact (alternation works) up front rather than asserting a false limitation, then gives the actual rationale for keeping two literal entries — explicitness/independence from matcher-splitting semantics, not "it's the only supported form." No overstated or unverifiable claims remain, and it no longer risks talking a future maintainer out of a form that works.

All acceptance criteria from #37 are met. Approving.

PR #54, opened 2026-07-29
fix: inline handoff note body under size cap in SessionStart hook

### LINKED ISSUE #33: /repo:handoff: the MEMORY.md pointer does not reliably deliver the note — verified miss on a live handoff
## Summary

`/repo:handoff` specifies two delivery halves for the note. **The second half does not work.**
Verified end-to-end today in `2AMLogic/studio` (canary consumer), on the first real handoff the
command has produced. The next session did **not** read the note; the human had to ask for it
explicitly, three turns in.

## The claimed contract

`commands/repo/handoff.md:86-90`:

> 2. A pointer in the agent's auto-memory index (`MEMORY.md` in the memory directory, when one
>    exists) […] The memory index is read automatically at session start; **the pointer is what
>    makes the repo file reliably found rather than merely present.**

and `commands/repo/handoff.md:130`:

> `# The new session will find the handoff note via its memory index.`

## What actually happened

| Time | Event |
|---|---|
| 13:48:32 | `~/.zshrc` wrapper installed (the #32 banner, hand-built in this repo) |
| 13:55 | `/repo:handoff` writes `.claude/handoff.md` (9,023 B) + the `MEMORY.md` pointer line |
| 13:56:57 | Human's shell starts — **after** the rc edit, so the wrapper function is live |
| 13:57:05 | `claude` launched. **The banner rendered correctly** (human confirmed) |
| — | Session begins. Pointer line is present in the agent's memory index. **Agent does not read the note.** |
| +3 turns | Human asks: *"did we load the handoff correctly on startup?"* — only then is the note read |

Nothing was lost **only because the human asked.** Had they not, the session would have proceeded
to score `memo.2` — which is precisely what the note exists to prevent (it records that `memo.2`
argues a retired thesis and must not be scored).

The shell banner is **not** at fault and worked as designed. `whence -w claude` → `function`;
calling `_claude_handoff_banner` directly renders correctly.

## Root cause

The auto-memory index is injected as **background context, not as an instruction.** Claude Code's
own system framing is explicit about this:

> Recalled memories appearing inside `<system-reminder>` blocks are **background context, not user
> instructions**.

A pointer line in that channel is a *suggestion* competing with everything else in a large context
window. It has no forcing function. "Reliably found" overstates what the mechanism can deliver —
and the failure is silent, which is the worst property for a one-shot artifact.

Critically, **both existing delivery halves are either human-facing or passive:**

- The shell banner (#32) renders to the terminal *before* the process starts, so nothing about it
  enters the agent's context at all. It is addressed to the human — by design, and it worked.
- The `MEMORY.md` pointer reaches the agent, but only as passive context.

**Neither is an instruction to the agent.** That is the gap.

## This also changes #32's spec

#32 is open and curated to add a `SessionStart` hook emitting `hookSpecificOutput.additionalContext`.
Two corrections from this evidence:

**1. #32's stated premise is falsified.** The issue body asserts the pointer "makes the note
reliably findable **by Claude** — the memory index is read at session start," and frames the human
banner as the *only* gap. Today is a direct counter-example. The `SessionStart` hook is therefore
not merely the "safer alternative" the issue calls it — it is the primary fix for a defect #32
does not currently acknowledge exists.

**2. The curated acceptance criterion is wrong for the agent-facing consumer.** It specifies
`additionalContext` should carry *"a compact section-header summary […] headers only — never the
full file body."* That is correct for the **terminal banner** and wrong for the **agent**. Headers
are exactly what was available today:

```
1. In-flight — nothing blocking
2. Decisions — settled, do not relitigate
3. Traps discovered empirically — these cost real time
```

Those convey shape but nothing actionable. The load-bearing content was *inside* §1 — "`memo.2` is
complete and UNSCORED […] do **not** spend a review cycle scoring it." A headers-only payload would
not have prevented the failure. **Two consumers, two verbosities:** headers for the terminal, full
body for the agent.

## Proposed fix, ranked by forcing strength

1. **Positional prompt** (strongest). `claude "<directive>"` — the CLI accepts a positional
   `[prompt]` that becomes the first user turn. This is the only channel with a real forcing
   function, because it arrives as a *user instruction*. Costs the human's first turn, so it should
   **prepend rather than replace** when the human supplies their own prompt.
2. **`--append-system-prompt`**. Does not consume a turn and carries system-level authority.
   Weaker than a user turn, stronger than context.
3. **`SessionStart` hook `additionalContext`** (what #32 currently specifies). Repo-scoped and
   requires no shell edit — but it is the *same passive-context class* as the pointer that just
   failed. Better than a pointer (inline content, not a reference), yet still not an instruction.

Options 1 and 2 are shell-wrapper territory — outside the repo, so #32's opt-in/default-N gating
correctly applies. Option 3 is repo-scoped and ungated.

**Recommendation:** ship (3) as the repo-scoped baseline that works with no shell edit, and offer
(1) in the optional shell wrapper for the strong guarantee. They compose. Whatever ships, the
agent-facing payload must be **the note body, not headers**.

Separately, `commands/repo/handoff.md:89-90` and `:130` should stop asserting the pointer is
sufficient until a mechanism actually makes it so.

---

Filed from `2AMLogic/studio` (canary consumer), where `/repo:handoff` was run end-to-end for the
first time on 2026-07-27. Related: #32.

---

### Status update — the landscape changed minutes after filing

Verified against `origin/main` (e0ee40b..2a77c74):

- **PR #34 merged 2026-07-27T21:19Z** (~17 min after this issue was filed) and shipped **option 3**: a `SessionStart` hook (`hooks/repo/session-start-handoff.sh`), installed by `install.sh` and wired under matchers `startup` + `resume`. **#32 is now closed as completed** — this issue's "#32 is open and curated" framing is stale; the spec corrections argued above now apply to the *shipped hook*, not to #32's pending scope.
- The shipped payload is **headers-only** (`MAX_HEADERS=9`; `session-start-handoff.sh:129` — "Headers only — never the file body") — the exact design this issue argues is insufficient. However, the hook **also injects an imperative directive** ("Read the note in full before doing anything else…", `session-start-handoff.sh:152-156`). That is materially stronger than the failed `MEMORY.md` pointer (an inline instruction rather than a passive reference), but it travels in the same `additionalContext` passive-context channel; whether it suffices is empirically untested.
- The **doc overstatements remain**, verified on `origin/main`: `commands/repo/handoff.md:90` ("the pointer is what makes the repo file reliably found rather than merely present") and `:130` ("The new session will find the handoff note via its memory index"). The post-#34 docs commit (1de4b1b) touched only README/SKILL.md; neither corrected these lines.

### Revised scope

Narrowed to what remains after PR #34:

1. **Doc correction** — `handoff.md` must stop asserting the pointer is sufficient, credit the SessionStart hook as the announcing mechanism, and demote the pointer to best-effort backup.
2. **Agent-facing payload policy** — decide and implement whether the hook inlines the note body (this issue's core claim: headers convey shape but nothing actionable).
3. **Options 1–2 (shell wrapper: positional prompt / `--append-system-prompt`)** — explicitly **out of scope** here. PR #34 deliberately deferred the wrapper; if still wanted, file separately.

### Implementation options — payload policy

**Option A — keep headers + directive (shipped); docs only.** Zero behavior risk; the body arrives when the directive is followed. But it does not satisfy this issue's core claim, and if the directive is ignored the failure stays silent — the property this issue calls "the worst" for a one-shot artifact.

**Option B — always inline the full body.** Strongest repo-scoped delivery. Unbounded injection at every startup/resume; the motivating note was 9,023 B and nothing caps a future pathological note.

**Option C — size-capped body (recommended).** If the note is ≤ a cap (suggest `MAX_BODY_BYTES=10240`, which covers the 9 KB motivating note), inline the **full body** in `additionalContext`; above the cap, fall back to the current headers summary plus a loud "too large to inline — read it now" line. Keep the read-in-full / one-shot-delete directive in **both** branches. Rationale: `/repo:handoff`'s exclusion discipline is designed to keep notes short, so the common case inlines; a once-per-launch injection of ≤10 KB is cheap relative to a silently lost handoff; the cap bounds the pathological case. This directly implements "two consumers, two verbosities" — the terminal banner (#32's wrapper) stays headers, the agent gets the body.

### Acceptance Criteria

- [ ] `commands/repo/handoff.md` no longer claims the memory-index pointer makes the note "reliably found" — both cited passages (`:86-90`, `:130`) corrected to name the SessionStart hook as the announcing mechanism, pointer as best-effort.
- [ ] `hooks/repo/session-start-handoff.sh` inlines the full note body when the note is ≤ cap, and falls back to headers + explicit oversize warning above cap; the read-in-full/one-shot directive is present in both branches.
- [ ] The hook's stable interface is preserved: silence + exit 0 when nothing to surface, always exit 0, exactly one JSON object on stdout, read-only, source gating (`startup`/`resume` only).
- [ ] `skills/repo/SKILL.md:143-144` and the README hook section no longer say "headers only, never the full body" — updated to describe the capped-body behavior.
- [ ] `pnpm test` passes with new cases covering the payload policy.

### Affected Files

- `commands/repo/handoff.md` — correct the two overstating passages (~:86-90, :130)
- `hooks/repo/session-start-handoff.sh` — payload policy (cap constant, body-inline branch, fallback branch)
- `hooks/repo/tests/test-session-start-handoff.sh` — new cases (see Test Plan)
- `skills/repo/SKILL.md` — hook description (~:143-144)
- `README.md` — hook section description

### Test Plan

- [ ] Automated tests: extend `hooks/repo/tests/test-session-start-handoff.sh` — body inlined for a note under the cap; headers fallback for a note over the cap; boundary at exactly the cap; directive present in both branches. Run via `pnpm test`.
- [ ] Manual verification: write a small `.claude/handoff.md`, run `echo '{"source":"startup","cwd":"'$PWD'"}' | hooks/repo/session-start-handoff.sh`, confirm the note body appears in `additionalContext`; repeat with an oversized note and confirm the headers fallback + warning.
- [ ] Edge cases: note with no headers, stale (>7 d) note, non-git cwd, missing jq — existing 79-case suite covers these; confirm still green.

### Dependencies

None. #32 is closed and PR #34 is merged — this issue builds on the shipped hook.

### Duplicate / related-work check

`check-duplicate.sh --include-merged-prs --issue 33`: no `RELATED_OPEN_WORK`; the only `DUPLICATE_FOUND` hit was this issue itself (100% self-match). File-reference enumeration in the original body (`handoff.md:86-90`, `:130`) re-verified accurate on `origin/main`.

rjwalters:
## Summary

The `session-start-handoff.sh` SessionStart hook previously surfaced only a headers-only outline of a pending `/repo:handoff` note. As issue #33 documented from a live miss, headers convey shape but nothing actionable — the load-bearing content lives in the note body — so a handoff could slip past a new session silently. This PR makes the hook inline the note body when it is small, corrects the docs that overstated the memory-index pointer, and extends the test suite.

## Changes

- **`hooks/repo/session-start-handoff.sh`** — new `MAX_BODY_BYTES=10240` cap. When the note is `<=` the cap, the **full body** is inlined between `----- BEGIN/END HANDOFF NOTE -----` markers. Above the cap, it falls back to the existing capped (`MAX_HEADERS=9`) header outline plus an explicit `OVERSIZE` warning. The read-in-full / one-shot-delete directive is appended in **both** branches. Stable interface preserved: silence + exit 0 when nothing to surface, always exit 0, exactly one JSON object on stdout, read-only, `startup`/`resume` source gating.
- **`commands/repo/handoff.md`** — the two overstating passages (`:86-90`, `:130`) no longer claim the pointer makes the note "reliably found." The pointer is now described as a best-effort backup (passive background context), and the SessionStart hook is named as the mechanism that actively announces the note.
- **`skills/repo/SKILL.md`** — hook description no longer says "headers only, never the full body"; now describes the capped-body behavior.
- **`README.md`** — hook section updated to describe full-body inline under the cap, header outline + oversize warning above it.
- **`hooks/repo/tests/test-session-start-handoff.sh`** — new payload-policy cases (body inlined under cap, headers fallback over cap, exact-cap boundary at 10240 and one byte over, directive present in both branches); reworked the former "headers-only" cases now that small notes inline in full.

- [x] `handoff.md` no longer claims the pointer makes the note "reliably found" — both cited passages corrected; SessionStart hook named as announcer, pointer as best-effort backup.
- [x] Hook inlines the full body at/under the cap and falls back to headers + oversize warning above it; read-in-full/one-shot directive present in both branches.
- [x] Stable interface preserved (silence + exit 0, always exit 0, one JSON object, read-only, `startup`/`resume` gating) — existing interface tests still green.
- [x] `SKILL.md` and README no longer say "headers only, never the full body" — both describe capped-body behavior.
- [x] Handoff test suite passes with new payload-policy coverage.

## Test Plan

- `pnpm test` → the delegated `session-start-handoff.sh` suite passes **99/99** (up from 79; new payload-policy cases added).
- Pre-existing note: `test-guard-destructive.sh` reports 8 `forceScope` failures on the base branch. That file is **not** touched by this PR (`git diff --name-only` confirms) — the failures are pre-existing and unrelated.

Closes #33

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ Every category is configurable per repo (`REPO_GUARD_*`/`REPO_*` env vars or `gu
 
 ## Handoff notes at session start
 
-Installing Repo Skills also wires a **SessionStart hook** (`session-start-handoff.sh`). When `/repo:handoff` has left a note at `.claude/handoff.md`, the hook surfaces it as session context on startup and resume — path, age, a section outline, and a staleness warning once the note passes seven days — so a rolled session opens already knowing what the last one left behind.
+Installing Repo Skills also wires a **SessionStart hook** (`session-start-handoff.sh`). When `/repo:handoff` has left a note at `.claude/handoff.md`, the hook surfaces it as session context on startup and resume — path, age, a staleness warning once the note passes seven days, and the note itself: the full body inlined when it is small (at or under a 10 KB cap), or a header outline plus an oversize warning when it is larger — so a rolled session opens already knowing what the last one left behind.
 
 It is strictly read-only: it never writes or deletes the note (absorbing it and removing it is `/repo:handoff`'s own one-shot contract), it stays silent when no note exists, and it fails open, so a hook error can never block session start. It deliberately does not fire on `/clear`, which is not a process relaunch.
 
--- a/commands/repo/handoff.md
+++ b/commands/repo/handoff.md
@@ -86,8 +86,14 @@ reset actually did — any earlier and it is speculative.
 2. A pointer in the agent's auto-memory index (`MEMORY.md` in the memory
    directory, when one exists): a single line —
    `- Handoff note at .claude/handoff.md — READ FIRST, then delete note + this line.`
-   The memory index is read automatically at session start; the pointer is
-   what makes the repo file reliably found rather than merely present.
+   The memory index is read automatically at session start, but it arrives as
+   passive background context, not an instruction — so the pointer is a
+   *best-effort backup*, not a guarantee the note is read (a live handoff has
+   been observed to slip past it). The mechanism that actively announces the
+   note is the `session-start-handoff.sh` **SessionStart hook** (installed by
+   Repo Skills), which surfaces the note as session context on startup and
+   resume — the full body inlined when the note is small, a header outline plus
+   an oversize warning when it is large.
 
 **What goes in — only what is not recoverable from the repo:**
 
@@ -127,7 +133,8 @@ to. End by printing an exact, copy-pasteable block for the human, e.g.:
 claude update
 #   3. Relaunch in this repo:
 cd <repo-root> && claude
-# The new session will find the handoff note via its memory index.
+# The SessionStart hook announces the handoff note to the new session
+# (the memory-index pointer is a best-effort backup).
 ```
 
 Then stop. The ritual is complete when the note is durable and the
--- a/hooks/repo/session-start-handoff.sh
+++ b/hooks/repo/session-start-handoff.sh
@@ -54,9 +54,15 @@
 RECENT_HOURS=48
 STALE_HOURS=168
 
-# Max section headers rendered. The real-world note that motivated this was
-# 7.5 KB - far too much to inject on every launch - but its headers convey the
-# shape in a handful of lines.
+# Payload sizing (issue #33). When the note is small enough (<= MAX_BODY_BYTES)
+# it is cheap to inline verbatim on every launch, and the full body is what
+# carries the actionable content - a headers-only outline conveys shape but
+# nothing to act on. Above that cap we fall back to a headers-only outline
+# (capped at MAX_HEADERS) plus a loud oversize warning, bounding the pathological
+# case. The 10 KB cap comfortably covers the ~9 KB real-world note that motivated
+# the hook, and /repo:handoff's exclusion discipline is designed to keep notes
+# short, so the common case inlines.
+MAX_BODY_BYTES=10240
 MAX_HEADERS=9
 
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd 2>/dev/null || echo ".")"
@@ -126,11 +132,6 @@ elif (( AGE_H < RECENT_HOURS ));  then AGE_LABEL="${AGE_H}h old"
 else                                   AGE_LABEL="$(( AGE_H / 24 ))d old"
 fi
 
-# Headers only - never the file body. `|| true` because grep exits 1 when the
-# note has no headers at all (a valid, non-error state) and head closing the
-# pipe early can surface as a non-zero status.
-HEADERS=$(grep -E '^#{1,2} ' "$NOTE" 2>/dev/null | head -n "$MAX_HEADERS" | sed 's/^#* *//' || true)
-
 CONTEXT="A /repo:handoff note is pending in this repository.
 
   Path: $NOTE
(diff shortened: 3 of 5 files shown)


rjwalters:
Reclaiming stale loom:reviewing claim (age > 30m, no follow-up comment) — a prior Judge's parent sweep likely died mid-review.

rjwalters:
✅ **Approved!**

Evaluation summary:

- **Issue linking**: PR body uses `Closes #33` — correct auto-close syntax.
- **Correctness**: The payload policy is implemented cleanly — full body inlined at/under `MAX_BODY_BYTES=10240` (`<=`, matching the tested boundary), headers-only outline + explicit `OVERSIZE` warning above it, and the read-in-full/one-shot directive appended in both branches. Arbitrary note content is safe: the body flows into `jq --arg`, so JSON encoding is handled correctly, and read failures stay on the fail-open path (`|| BODY=""`, jq-failure guard emits nothing + exit 0). Stable interface (silence + exit 0, one JSON object, read-only, startup/resume gating) preserved.
- **Tests**: New payload-policy coverage is thorough — under-cap inline with markers, over-cap fallback, exact-cap boundary at 10240 bytes and one-byte-over flip, directive in both branches, and the former headers-only cases correctly reworked. Suite passes 99/99 locally.
- **Docs**: `handoff.md`, `SKILL.md`, and README all consistently updated; the memory-index pointer is now correctly described as best-effort backup with the SessionStart hook as the active announcer.

Actions taken during evaluation:
- 🔀 Rebased the branch onto latest main (branch predated `ci.yml` from #55, so `gh pr checks` reported "no checks reported") and pushed. CI then ran and **passed** (run 30482914209: success). Merge state is clean.
- Pre-existing note confirmed: the 7–8 local `test-guard-destructive.sh` failures reproduce identically on current main (macOS-local environment quirk; CI on main is green) and are unrelated — this PR does not touch that file.

## Test Execution

**Test plan from PR description:**
1. `pnpm test` → delegated `session-start-handoff.sh` suite — ✅ Executed: 99/99 pass (re-verified after rebase)
2. Pre-existing `test-guard-destructive.sh` failures unrelated — ✅ Verified: reproduce on base branch; CI green on main and on this PR

Issue #46, opened 2026-07-28
test: branches loss-check suite fails on git < 2.38 and never covers the documented merge-tree fallback


rjwalters:
## Context

PR #41 (issue #39) fixed the permanent-loss check in `commands/repo/branches.md` and made it content-aware, using:

```bash
[ "$(git merge-tree --write-tree <default> <branch>)" = "$(git rev-parse '<default>^{tree}')" ]
```

`git merge-tree --write-tree` requires **git >= 2.38** (Oct 2022). The prose handles this correctly — `commands/repo/branches.md:161` explicitly documents the degraded path:

> If `git merge-tree` is unsupported (git < 2.38) or reports a conflict, fall back …

and `:188` lists "`git merge-tree` unsupported" among the degraded-mode conditions that must resolve to KEEP.

**The test suite does not honor that contract.** `commands/repo/tests/test-branches-loss-check.sh:252`:

```bash
MT=$(git -C "$REPO" merge-tree --write-tree main feature/squashed 2>/dev/null)
assert_eq "5b: merging the squash-merged branch changes nothing" \
    "$(git -C "$REPO" rev-parse 'main^{tree}')" "$MT"
```

On git < 2.38 the command fails, `2>/dev/null` swallows the error, `MT` is empty, and the assertion fails with a confusing tree-hash-vs-empty-string mismatch. So:

1. `pnpm test` **fails outright** on git < 2.38 — on the exact versions the documented fallback exists to support.
2. The fallback path itself has **no coverage at all**. Nothing verifies that an environment without `merge-tree` degrades to KEEP rather than to SAFE — and "degrades to SAFE" is precisely the silent-data-loss failure #39 was filed to prevent.

Item 2 is the one that matters. A loss check whose degraded path is untested is a loss check with an untested failure mode.

## Not currently observable

Local git is 2.55.0, so the suite passes here. There is no CI in this repo (no `.github/workflows/`; `check:ci` is an `echo` stub), so nothing exercises another git version. This was surfaced by the Judge during review of PR #41 and deliberately not blocked on, since it is theoretical for the current machine.

## Suggested acceptance criteria

- [ ] `test-branches-loss-check.sh` detects `merge-tree --write-tree` support up front (e.g. probe once on a scratch repo, or compare `git --version` against 2.38).
- [ ] On unsupported git, the `merge-tree`-dependent assertions **skip with a clear message** rather than failing — `pnpm test` must stay green on older git.
[...]

Issue #36, opened 2026-07-27
test: `pnpm test` skips the 444-case guard-destructive regression suite


rjwalters:
## Problem

`pnpm test` runs `hooks/repo/tests/run.sh`, which reports **56 cases**. The guard-destructive regression suite — `hooks/repo/tests/test-guard-destructive.sh` — reports **444 cases** and is **never invoked by `pnpm test`**.

So the repo's headline test command exercises roughly **12%** of the guard-destructive coverage. The 444-case suite only runs if someone remembers to invoke it by hand.

This is pre-existing, not a regression: `test-guard-destructive.sh:18` already documents the split ("The quick smoke suite lives next door in run.sh; this file is the full regression suite"), and `run.sh` on `main` has never called it. It was surfaced by the Judge during review of #34 and deliberately not blocked on there, since #34 neither caused nor worsened it.

## Why it is worth fixing now

#34 added a clean delegation mechanism to `run.sh` — it now shells out to `test-session-start-handoff.sh` and propagates that suite's failure into its own exit code. The Judge fault-injected this during review (replaced the hook with a bare `exit 0` stub) and confirmed `run.sh` correctly went to `PASS: 55 FAIL: 1`, exit 1. So the pattern is proven.

Extending the same delegation to `test-guard-destructive.sh` should be a small change, and it closes the gap for the single most safety-critical script in the repo — the hook that decides whether a destructive Bash command gets blocked.

## Notes (original)

Decide deliberately whether the 500 combined cases belong in one command or two. The current state is the worst of both — it *looks* like one command covers everything, and it doesn't. Either outcome (one full suite, or two clearly-named commands) is better than the status quo.

## Related

- #34 — added the `run.sh` delegation pattern this issue extends
- `hooks/repo/tests/test-guard-destructive.sh:18` — documents the existing split

---

### Verified numbers (measured 2026-07-27, this machine)

| Suite | Cases | Wall time |
|---|---|---|
| `hooks/repo/tests/run.sh` (`pnpm test` today) | 56 (includes the delegated `test-session-start-handoff.sh` suite folded in as 1 case, 79 sub-cases) | ~8.9s |
| `hooks/repo/tests/test-guard-destructive.sh` (standalone) | 444 | ~54.7s |
[...]

PR #55, opened 2026-07-29
ci: add GitHub Actions workflow running the full test suite

rjwalters:
## Summary

Nothing ran the ~532-case suite automatically: there was no `.github/workflows/`
directory, and `pnpm test` fired only when a human remembered to run it, so
`gh pr checks <n>` reported "no checks reported" and PRs merged unverified. On top
of that, `package.json`'s `check:ci` / `check:all` were `echo ... && exit 0` stubs
— a script named like a CI gate that always passes, which the `.loom` role docs
reference as this repo's documented CI command.

This PR adds a GitHub Actions workflow that runs `pnpm test` on every pull request
and on pushes to `main`, and repoints the two stub scripts at the real suite.

## Changes

- **`.github/workflows/ci.yml`** (new): a `test` job on `ubuntu-latest` that
  checks out the repo, ensures `jq` is present (the harness hard-fails without it),
  reports the runner git version (the loss-check suite needs git >= 2.38, per #46;
  `ubuntu-latest` ships >= 2.43), installs pnpm + Node, and runs `pnpm test`. A
  failing assertion turns the check red. Triggers on `pull_request` and `push` to
  `main`; a `concurrency` group cancels superseded runs.
- **`package.json`**: repoint `check:ci` and `check:all` from their
  `echo ... && exit 0` stubs to `pnpm test`, so the CI command the role docs
  reference is a genuine gate. Deleting them was the other option allowed by the
  acceptance criteria, but the `.loom/roles/*.md` and `.loom/docs/build-gate.md`
  files assume `pnpm check:ci` exists, so wiring it to the suite avoids a docs
  regression across those files.

## Acceptance Criteria Verification

- [x] A workflow runs `pnpm test` on PRs and on pushes to the default branch —
  `.github/workflows/ci.yml` triggers on `pull_request` and `push: branches: [main]`.
- [x] The workflow fails when the suite fails — verified by fault injection: a
  broken assertion makes `run.sh` exit `1` (the passing suite exits `0`), and the
  `Run test suite` step has no `continue-on-error`, so a nonzero exit fails the job.
- [x] Runner git is >= 2.38 — `ubuntu-latest` ships >= 2.43; the workflow logs
  `git --version` so a future runner-image regression below the floor is visible.
- [x] `check:ci` / `check:all` do something real — both now run `pnpm test`; no
  script exits 0 unconditionally while named like a gate.
- [x] Suite-run split documented — the workflow runs the full ~60s suite on both
[...]

PR #56, opened 2026-07-29
test(run): fold the two delegated suites in at real case counts

### LINKED ISSUE #44: test: two delegated suites still count as 1 case each — headline says 501, breakdown sums to 578
## Context

PR #40 (issue #36) folded the 444-case `test-guard-destructive.sh` suite into `pnpm test` and introduced the machinery to do it properly: `strip_ansi`, `suite_count`, and `record_suite` in `hooks/repo/tests/run.sh`. Delegated suites wired through `record_suite` contribute their **real** case counts to the aggregate and get their own row in the per-suite breakdown table.

Two delegations predate that machinery and still use the older inline-`awk` shape. Each contributes **1** to the headline total regardless of how many cases it actually runs:

| Suite | Real cases | Currently contributes | In breakdown table? |
|---|---|---|---|
| `test-session-start-handoff.sh` | 79 | 1 | yes, but annotated "folded into totals as 1 case" |
| `test-install-claude-md-markers.sh` | 49 | 1 | **no** — absent entirely |

Observed on `main` at `06b802a`: headline is `PASS: 501`, while the breakdown column sums to 578. Those two numbers should agree.

`test-install-claude-md-markers.sh` is the worse of the two: it landed in PR #42 after #40 was cut, so it has no `record_suite` call at all and does not appear in the breakdown. Its `49 cases pass` line still prints in the streaming output above the table, so it is easy to believe it is accounted for when it is not.

## This is a reporting defect, not a safety hole

Verified during review of #40: both old-shape blocks still do `FAIL=$((FAIL+1))` on a non-zero child exit, so a regression in either suite **does** fail `pnpm test`. Nothing is silently passing. The problem is that the headline number understates coverage by 126 cases and the breakdown table is internally inconsistent — which is corrosive for a repo whose only automated gate is this one command (there is no CI; `check:ci` is an `echo` stub).

It is also the same class of problem #36 was filed about: a test count that looks like it means one thing and means another.

- [ ] `test-session-start-handoff.sh` and `test-install-claude-md-markers.sh` are converted to the `record_suite` + real-count shape introduced by #40.
- [ ] The headline aggregate rises from 501 to ~628 with no change in what is actually executed — confirm the delta is exactly the two suites' real counts (79 + 49 = 128, less the 2 they currently contribute).
- [ ] The per-suite breakdown column **sums to the headline aggregate**. Add an assertion or a self-check so this cannot silently drift again.
- [ ] The "folded into totals as 1 case" annotation is removed once it is no longer true.
- [ ] Verify by fault injection, not by reading code: break one assertion in each converted suite, confirm `pnpm test` reports the failing case by name and exits non-zero, then restore and confirm a clean tree.
- [ ] No suite loses the ability to run standalone.

## Notes

The `record_suite` helpers make this close to a copy-a-block change — #40's file header documents the extension point explicitly. The value is that after this, the number `pnpm test` prints is the number of cases that ran.

## Related

- #36 / PR #40 — introduced the machinery and the decision to keep one command, no `test:full` split
- #38 / PR #42 — added the marker suite that is currently uncounted

---

### Updated arithmetic — the counts have moved since filing

This issue was filed against `06b802a` (headline 501, breakdown 578). PR #41 has since merged `test-branches-loss-check.sh` (31 cases, properly wired through `record_suite`), so on current `main` (`3f46e49`) the picture is:

| Component | Real cases | Contributes to headline | In breakdown table? |
|---|---|---|---|
| `run.sh` inline smoke cases | 55 | 55 | yes |
| `test-guard-destructive.sh` | 444 | 444 | yes |
| `test-branches-loss-check.sh` | 31 | 31 | yes |
| `test-session-start-handoff.sh` | 79 | **1** | yes, annotated "folded into totals as 1 case" |
| `test-install-claude-md-markers.sh` | 49 | **1** | **no** — absent entirely |

Headline today: **532**. Breakdown column sums to **609**. The defect is unchanged; only the numbers moved.

**Corrected acceptance-criterion numbers**: the headline should rise from **532 to 658** (532 − 2 + 79 + 49), a delta of exactly 126. Both suite counts were re-verified on 2026-07-29 by running each standalone: 79/79 and 49/49 pass.

### Implementation guidance

Both target suites already print the `Total:/Passed:/Failed:` trailing summary that `suite_count` parses (verified on `origin/main` — `test-session-start-handoff.sh` ~line 415, `test-install-claude-md-markers.sh` ~line 445), so **neither suite file needs changes**. The conversion is the documented copy-a-block change inside `run.sh`:

1. Model both blocks on the `test-branches-loss-check.sh` block at the bottom of `run.sh` (or the guard-destructive block — same shape): capture output **and** exit status, extract `suite_count Passed` / `suite_count Failed`, guard the non-numeric-parse case (`PASS=0; FAIL=1` + dump output tail — never fold an unparseable summary in as zero failures), fold real counts into `PASS`/`FAIL`, and call `record_suite` with the real numbers.
2. Remove the "folded into totals as 1 case" annotation from the handoff block's `record_suite` call; add a `record_suite` call to the markers block (it currently has none).
3. For the "cannot silently drift again" criterion: accumulate machine-readable per-suite pass/fail sums alongside `SUITE_LINES` (or extend `record_suite` to track running sums), then after printing the breakdown assert the sums equal the final `$PASS`/`$FAIL` and fail the run if not. Keep it self-contained in `run.sh`. Design note (interaction with #46): assert over the **pass and fail columns**, not row totals, so a future skip column (proposed in #46 for the loss-check suite) can be added without weakening or rewriting this check.

### Affected Files

- `hooks/repo/tests/run.sh` — convert the two old-shape delegation blocks ("session-start-handoff.sh (delegated suite)" and "install/uninstall CLAUDE.md markers (delegated suite)") to the real-count + `record_suite` shape; add the breakdown-sums-to-headline self-check; update the file-header comment if it still describes the two suites as folding in as one case.
- `hooks/repo/tests/test-session-start-handoff.sh` — **no changes** (summary format already parseable; listed for scope clarity only).
- `hooks/repo/tests/test-install-claude-md-markers.sh` — **no changes** (same).

### Test Plan

- [ ] Manual verification: `pnpm test` on a clean tree → headline `PASS: 658   FAIL: 0`; per-suite breakdown shows **five** rows including `test-install-claude-md-markers.sh`; pass column sums to 658; no "folded into totals as 1 case" annotation remains.
- [ ] Fault injection (per AC — do not verify by reading code): break one assertion in `test-session-start-handoff.sh`, confirm `pnpm test` exits non-zero and reports the failing case by name with real counts in headline and breakdown; restore. Repeat for `test-install-claude-md-markers.sh`.
- [ ] Self-check verification: temporarily corrupt one `record_suite` count (or the accumulation), confirm the new drift assertion fails the run; restore and confirm a clean tree passes.
- [ ] Standalone: `bash hooks/repo/tests/test-session-start-handoff.sh` and `bash hooks/repo/tests/test-install-claude-md-markers.sh` still run and pass on their own.
- [ ] Automated tests: N/A beyond the harness itself — this change *is* the test harness.

### Dependencies

None. #46 and #48 are related open work, not prerequisites (see curator comment for reconciliation).

rjwalters:
## Summary

Two delegated test suites — `test-session-start-handoff.sh` (79 cases) and `test-install-claude-md-markers.sh` (49 cases) — used the old inline-`awk` delegation shape and each contributed only **1** to the `pnpm test` headline PASS/FAIL total, regardless of how many cases they actually ran. The markers suite had no `record_suite` call at all, so it was absent from the per-suite breakdown entirely. The result: headline `PASS: 532` while the breakdown column summed to `609` — the same "a count that looks like one thing and means another" defect #36 was filed about.

This is a reporting fix only. Both old-shape blocks already did `FAIL+1` on a non-zero child exit, so regressions in either suite already failed `pnpm test`; nothing was silently passing. Nothing about what is executed changes.

## Changes

- `hooks/repo/tests/run.sh`:
  - Convert the `session-start-handoff.sh` and `install/uninstall CLAUDE.md markers` delegation blocks to the `record_suite` + real-count shape already used by `test-guard-destructive.sh` / `test-branches-loss-check.sh` — capture output **and** exit status, parse `Passed`/`Failed` with the non-numeric-parse guard (`PASS=0; FAIL=1` + output tail, never fold an unparseable summary in as zero failures), fold real counts into `PASS`/`FAIL`, and record a breakdown row.
  - Add a **drift self-check**: `record_suite` now accumulates the pass/fail *columns*, and after the breakdown prints, the run asserts those sums equal the headline aggregate and aborts non-zero if not. Asserting over columns (not row totals) leaves room for a future skip column (#46) without weakening the check.
  - Remove the stale "folded into totals as 1 case" annotation and refresh the file-header comment.

- [x] Both suites converted to the `record_suite` + real-count shape.
- [x] Headline rises from **532 to 658** (532 − 2 + 79 + 49 = 658, delta +126), no change to what is executed.
- [x] Per-suite breakdown column **sums to the headline** — enforced by a self-check that fails the run on drift (verified: corrupting one `record_suite` count to 999 aborts with `FATAL: breakdown does not sum to headline`, exit 1).
- [x] "folded into totals as 1 case" annotation removed.
- [x] Fault injection (not code reading): a deliberate failing case injected into each converted suite is reported **by name** in headline and breakdown with real counts, and `pnpm test` exits non-zero — `FAIL: FAULT-INJECTION-PROBE-HANDOFF` / `FAIL: FAULT-INJECTION-PROBE-MARKERS`, headline `PASS: 658 FAIL: 2`; both reverted afterward.
- [x] Neither suite lost the ability to run standalone (both pass on their own).

## Test Plan

- [x] Clean `pnpm test` → headline `PASS: 658  FAIL: 0`; breakdown shows **five** rows including `test-install-claude-md-markers.sh`; columns sum to 658; no stale annotation.
- [x] Fault injection in each converted suite → failing case named, non-zero exit; restored to a clean tree.
- [x] Drift self-check → corrupting one `record_suite` count fails the run; clean tree passes.
- [x] Standalone: `test-session-start-handoff.sh` and `test-install-claude-md-markers.sh` both pass on their own.

Closes #44

### DIFF
--- a/hooks/repo/tests/run.sh
+++ b/hooks/repo/tests/run.sh
@@ -8,10 +8,13 @@
 # (deny / ask / allow).
 #
 # Sibling suites are NOT inlined here — they are delegated to at the end of this
-# file, each in its own block that folds the suite's verdict into this runner's
-# PASS/FAIL totals and records a line in the per-suite breakdown. To add another
-# suite, copy one of those blocks; nothing else needs to change. Currently
-# delegated: test-guard-destructive.sh (the full guard regression suite),
+# file, each in its own block that folds the suite's REAL per-case PASS/FAIL
+# counts into this runner's totals and records a line in the per-suite
+# breakdown. To add another suite, copy one of those blocks; nothing else needs
+# to change. A self-check after the breakdown asserts the breakdown columns sum
+# to the headline aggregate, so a miswired block can never make the headline
+# diverge from the breakdown (repo#44). Currently delegated:
+# test-guard-destructive.sh (the full guard regression suite),
 # test-session-start-handoff.sh, test-install-claude-md-markers.sh, and
 # commands/repo/tests/test-branches-loss-check.sh.
 #
@@ -98,12 +101,20 @@ strip_ansi() { printf '%s\n' "$1" | sed $'s/\033\\[[0-9;]*m//g'; }
 suite_count() { strip_ansi "$2" | awk -v f="$1:" '$1 == f { print $2; exit }'; }
 
 # Per-suite breakdown lines, printed above the final aggregate totals so the
-# result is never one opaque number.
+# result is never one opaque number. The pass/fail columns are also accumulated
+# into machine-readable running sums so the breakdown can be asserted to sum to
+# the headline aggregate — see the self-check after the breakdown is printed.
 # record_suite <label> <pass> <fail> [note]
 SUITE_LINES=()
+SUITE_PASS_SUM=0
+SUITE_FAIL_SUM=0
 record_suite() {
     SUITE_LINES+=("$(printf '  %-38s %4s pass  %4s fail%s' \
         "$1" "$2" "$3" "${4:+  ($4)}")")
+    # Sum the pass and fail COLUMNS (not row totals) so a future skip column can
+    # be added to the breakdown without weakening this invariant (repo#46).
+    [[ "$2" =~ ^[0-9]+$ ]] && SUITE_PASS_SUM=$((SUITE_PASS_SUM + $2))
+    [[ "$3" =~ ^[0-9]+$ ]] && SUITE_FAIL_SUM=$((SUITE_FAIL_SUM + $3))
 }
 
 echo "guard-destructive.sh test suite"
@@ -241,8 +252,9 @@ fi
 
 # The SessionStart handoff hook ships its own suite rather than inline cases:
 # its payload shape, assertions, and install/uninstall wiring checks share
-# nothing with the guard's expect() helper above. Delegate to it and fold the
-# verdict in as a single case so `pnpm test` covers both hooks.
+# nothing with the guard's expect() helper above. Delegate to it and fold its
+# real PASS/FAIL counts — not a single collapsed case — into this runner's
+# totals, so `pnpm test` genuinely reports every case it runs (repo#44).
 echo
 echo "-- session-start-handoff.sh (delegated suite) --"
 SS_TEST="$TESTS_DIR/test-session-start-handoff.sh"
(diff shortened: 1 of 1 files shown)


rjwalters:
✅ **Approved!**

Verified against issue #44's acceptance criteria by execution in the `issue-44` worktree, not by reading code:

**Clean run** — headline `PASS: 658   FAIL: 0`, exit 0. Breakdown now shows five rows including `test-install-claude-md-markers.sh`, and the columns sum to the headline: 55 + 444 + 79 + 49 + 31 = 658. Delta from prior 532 is exactly +126 (79 + 49 − 2), matching the curator's corrected arithmetic.

**Drift self-check** — corrupting one `record_suite` count to 999 aborts the run with `FATAL: breakdown does not sum to headline (harness accounting bug)`, exit 1. Asserting over the pass/fail *columns* (not row totals) is the right call — it leaves room for a future skip column (#46) without weakening the invariant.

**Fault injection** (per AC, not code reading) — injected a deliberate failing case into `test-install-claude-md-markers.sh`; the harness reported it by name (`FAIL: FAULT-INJECTION-PROBE-MARKERS`), folded real counts (49 pass, 1 fail), surfaced `PASS: 658  FAIL: 1`, and exited non-zero. Restored to a clean tree afterward.

**Standalone** — `test-session-start-handoff.sh` (79/79) and `test-install-claude-md-markers.sh` (49/49) both pass on their own.

Code quality is solid: the non-numeric-parse guard (`PASS=0; FAIL=1` + output tail rather than folding an unparseable summary in as zero failures) is a genuine improvement over the old inline-`awk` shape, the "folded into totals as 1 case" annotation is gone, and the file-header comment now documents the column-sum invariant as the extension-point contract. Scope is tight — a single file (`run.sh`, +81/−28), no suite files touched, exactly as the issue scoped it.

Merge state is CLEAN / MERGEABLE. Main has moved since this branched (#55 added `.github/workflows/ci.yml` + repointed `package.json`), but those are disjoint files — no conflict, and disjoint-file drift is not a blocker. No CI checks are reported on this branch (expected: it predates #55's workflow); the repo's real gate is `pnpm test`, which passes here.

`Closes #44` present. Nice work.

PR #58, opened 2026-07-29
fix(guard-destructive): redact dangerous strings quoted as echo/printf data

### LINKED ISSUE #53: guard-destructive: dangerous strings inside quoted literals still block (echo/heredoc data, guard self-tests)
## Context
Surfaced 2026-07-29 in rjwalters/loom while verifying the guard chain after a hook-file anomaly: running a guard self-test of the form

    echo '{"tool_name":"Bash","tool_input":{"command":"rm -rf /"}}' | .loom/hooks/guard-destructive.sh

was itself BLOCKED by the PreToolUse guard ("Command matches dangerous pattern: rm[[:space:]]+...") before it could run. The dangerous text appears only as a single-quoted literal argument to `echo` — data, not an executable command. The guard's own documentation claims "quote-aware parsing and literal-text redaction keep prose, commit messages, and issue bodies from tripping it"; that redaction evidently does not cover this shape (single-quoted JSON payload piped to a script).

**Live second repro while filing THIS issue:** the `gh issue create` invocation carrying this body inline (heredoc) was itself blocked by the same pattern — the danger string appears here only as documentation text. Filed successfully via `--body-file` (the guard scans the command string, not file contents), which is the same workaround loom uses for its curl-pipe FP (loom#4040).

Impact is modest but real: guard self-tests, writing test fixtures for the guard, issue/PR bodies quoting dangerous commands as examples, and log greps can all be blocked.

- [ ] A dangerous pattern appearing entirely inside a single- or double-quoted argument to a non-shell-invoking command (echo, printf, grep pattern args, heredoc bodies fed to gh) does not trigger a block.
- [ ] Piping into a guard script for testing works (or a documented GUARD_SELFTEST escape hatch exists).
- [ ] Regression cases added to the guard test suite for both repro shapes above (quoted-literal pipe, heredoc issue body).

---

### Problem Statement

Confirmed live in this repo (`hooks/repo/guard-destructive.sh`, the canonical Repo Skills destructive-command guard — `.claude/skills/repo/hooks/guard-destructive.sh` is a symlink to it, and Loom's `.loom/hooks/guard-destructive.sh` dispatches to it):

```
echo '{"tool_name":"Bash","tool_input":{"command":"rm -rf /"}}' | .loom/hooks/guard-destructive.sh
# -> BLOCKED: Command matches dangerous pattern: rm[[:space:]]+-[a-zA-Z]*[rf][a-zA-Z]*[[:space:]]+/([^[:alnum:]._~/-]|$)
```

Root cause: the catastrophic `ALWAYS_BLOCK_PATTERNS` scan runs against `COMMAND_NO_LITERAL_TEXT`, which is produced by `strip_literal_text()` (`hooks/repo/guard-destructive.sh:1038-1090`). That function redacts a quoted span **only** when it immediately follows one of a fixed, narrow set of text-carrying flags: `--message`, `--body`, `--notes`, `--title`, `--comment`, `-m` (see the `re =` regex at line 1048, and the gating `if` at lines 1236-1241/1284-1288 that only even attempts redaction when one of those flag substrings is present in the raw command). A dangerous string quoted as a plain argument to `echo`/`printf`, or embedded in a heredoc body piped into another command, is **not** preceded by any of those flags, so it survives into `COMMAND_NO_LITERAL_TEXT` verbatim and trips the raw substring/regex scan in `ALWAYS_BLOCK_PATTERNS` (`hooks/repo/guard-destructive.sh:1149-1226`), even though nothing in the command actually executes the dangerous payload.

This is the same class of false positive that #3679/#3756 fixed for `--body`/`-m`/etc., but `strip_literal_text()`'s flag allowlist was never extended to generic data-sink commands (`echo`, `printf`) or to heredoc bodies fed to `gh`/other CLIs. The README's claim ("quote-aware parsing and literal-text redaction keep prose, commit messages, and issue bodies from tripping it" — `README.md:34`) is broader than what the code actually does today.

There is currently **no** `GUARD_SELFTEST`-style escape hatch in the script (verified: no `SELFTEST` string anywhere in `hooks/repo/guard-destructive.sh`), so piping a test payload into the guard for its own self-tests has no documented workaround other than `--body-file`-style indirection (which doesn't help for the `echo | guard-destructive.sh` shape at all, since there's no flag to redact).

### Suggested acceptance criteria (refined)

- [ ] A dangerous pattern that appears **only** inside a single- or double-quoted argument to a non-executing "data sink" command (`echo`, `printf`, and equivalently a heredoc body handed to a script/CLI as its stdin or as a `--body`-style value) does not trigger a catastrophic `deny`.
- [ ] The existing safety floor is unchanged: a dangerous command reachable via command substitution or backticks inside a quoted span (e.g. `echo "$(<dangerous command>)"`, `bash -c '<dangerous command>'`) is still hard-denied — mirror `strip_literal_text()`'s existing `$(`/backtick escape-hatch check (lines 1078-1084) and the qsplit() precedent at lines 852-925 that keeps separators/commands active whenever command substitution is present.
- [ ] Piping a self-test payload into the guard script works, either via the broadened redaction above or via a documented, narrowly-scoped escape hatch (e.g. a `REPO_GUARD_SELFTEST`/legacy `LOOM_GUARD_SELFTEST` env var, following the existing `REPO_*`-wins-over-`LOOM_*` toggle convention used throughout this file). Document the chosen mechanism in the script's "STABLE INTERFACE" header comment block (lines 16-57).
- [ ] Regression cases added to `hooks/repo/tests/test-guard-destructive.sh` for both repro shapes: (a) `echo '<json containing a dangerous command as data>' | guard-destructive.sh`-style piping, and (b) a multi-line/heredoc quoted literal containing dangerous text used as documentation/example prose (mirroring the #3898 multi-line redaction regression already covered by `strip_literal_text()`'s `END`-block buffering).
- [ ] No regression in guard coverage: the full existing suite (`hooks/repo/tests/test-guard-destructive.sh`, and the smoke suite in `hooks/repo/tests/run.sh` if present) still passes, and no genuinely dangerous directly-executed command (not inside quoted data) loses its deny/ask decision.

### Implementation Guidance

This file's own commentary (segment parsing, `qsplit()`, `strip_literal_text()`) repeatedly emphasizes "never widen a deny into an allow" and prefers **structural, first-token-anchored** admission over blanket content redaction (see the read-only fast path's rationale at lines 194-231). Apply the same philosophy here rather than a blanket "redact every quoted string" rule, which would risk suppressing genuine `bash -c '<payload>'`/`sh -c` detections.

Options to weigh (not prescriptive — Architect/Builder should pick and document the choice):

1. **Extend `strip_literal_text()`'s command-word anchor, not just its flag anchor.** Reuse the existing `_QSPLIT_AWK` quote-aware segmentation (lines 876-925, already shared by `parse_force_ops`/`lifecycle_or_cloud_reason`/`extract_rm_targets`) to identify segments whose **command word** is a known non-executing data sink (`echo`, `printf`) and redact their quoted argument(s) — but only when the quoted span itself carries no `$(`/backtick (same escape-hatch check already in `strip_literal_text()`). This is deliberately narrower than "any quoted string anywhere," and structurally excludes `bash -c`/`sh -c`/`eval`/`xargs` wrappers the same way `fastpath_builtin_admits()` does (command-word-anchored, never a wrapper).
2. **Heredoc handling** likely needs separate treatment from the flag-based/echo-based redaction above, since a heredoc body (`<<'EOF' ... EOF`) piped to `gh issue create`/`gh pr create` etc. isn't a quoted span in the same sense — consider whether the existing `--body`/`--title`/etc. flag detection already covers the common `gh ... --body "$(cat <<'EOF' ...)"` shape, or whether raw heredoc bodies (no `--body` flag on the same line) need their own redaction pass.
3. **Self-test escape hatch**: a narrowly-scoped `REPO_GUARD_SELFTEST`/`LOOM_GUARD_SELFTEST` env toggle (mirroring the existing `REPO_*`/`LOOM_*` env-override convention, e.g. `REPO_GUARD_READONLY_FASTPATH` at lines 411-419) that a test harness sets when deliberately piping a dangerous-looking payload into the guard for self-testing, documented in the STABLE INTERFACE block. This is the smallest, lowest-risk fix for the "guard self-test" half of the repro and does not by itself address the general echo/heredoc-as-documentation false positive — likely both this and option 1 are needed to satisfy all suggested acceptance criteria.
4. Do **not** hand-edit `.loom/hooks/guard-destructive-generic.sh` — its own header explicitly says "DO NOT hand-edit generic pattern behavior here — send fixes upstream to Repo Skills" (it is a vendored copy re-synced at release time). The fix belongs solely in `hooks/repo/guard-destructive.sh`; `.claude/skills/repo/hooks/guard-destructive.sh` is a symlink to it in this repo (no separate edit needed here), but note that consumer repos get a **copy**, so this is a source-of-truth fix, not a copy-everywhere fix.
5. Consider whether `README.md:34`'s claim about literal-text redaction should be narrowed/clarified once the actual scope of the fix is known (optional polish, not required for acceptance).

### Affected Files

- `hooks/repo/guard-destructive.sh` - core fix: broaden `strip_literal_text()` (or add a parallel redaction pass) to cover quoted arguments to non-executing data-sink commands (echo/printf) and/or heredoc bodies, and/or add a documented self-test escape hatch; update the STABLE INTERFACE header comment block (lines 16-57) to document any new toggle.
- `hooks/repo/tests/test-guard-destructive.sh` - add regression tests for both repro shapes (echo-piped JSON payload, heredoc/quoted documentation text containing a dangerous command), and confirm no existing catastrophic-deny test regresses.

### Test Plan

- [ ] Manual verification: run `echo '{"tool_name":"Bash","tool_input":{"command":"rm -rf /"}}' | hooks/repo/guard-destructive.sh` (and the equivalent `.loom/hooks/guard-destructive.sh` dispatcher path) and confirm it now allows (silent, exit 0, no BLOCKED output) instead of denying.
- [ ] Manual verification: confirm a genuinely dangerous directly-executed command with no quoting at all, and a smuggled form like `echo "$(rm -rf /)"`, both still deny — no safety-floor regression.
- [ ] Automated tests: extend `hooks/repo/tests/test-guard-destructive.sh` with cases for (a) the echo-piped JSON self-test shape and (b) a heredoc/quoted-literal documentation shape containing dangerous text; run the full suite (`./hooks/repo/tests/test-guard-destructive.sh`) and confirm 0 failures.
- [ ] Edge cases: verify `bash -c 'rm -rf /'` and `sh -c 'rm -rf /'` still hard-deny (the `-c` flag is deliberately excluded from redaction per the existing `strip_literal_text()` comment at lines 1027-1028) and are not accidentally swept up by any new echo/printf-specific redaction.

rjwalters:
## Summary

`echo`/`printf` print their arguments; they never execute them, so a dangerous command quoted as a literal argument to a data sink is inert **data** — not an executable command. But `strip_literal_text()`'s redaction only covered quoted values of a fixed flag allowlist (`--body`/`-m`/`--title`/`--notes`/`--comment`). A payload quoted as an `echo`/`printf` argument survived verbatim into the raw `ALWAYS_BLOCK`/`ASK` scans and false-blocked. This is the meta false-positive that blocked a guard self-test (`echo '{"…":"<danger>"}' | guard-destructive.sh`) and blocked filing the issue's own heredoc body.

## Changes

- **`strip_datasink_literals()`** (new): a quote-aware, single-pass, **command-word-anchored** lexer that redacts the quoted argument(s) of a simple command whose first token is exactly `echo` or `printf` (behind an optional bare `sudo`/`env` wrapper). Multi-line spans are redacted same-length (byte offsets preserved). A wrapper that actually executes its argument (`bash -c`, `sh -c`, `eval`, `xargs`) is never a data sink.
- **`command_has_shell_segment()`** (new): gates the redaction. When any quote-aware pipeline segment's command word is a shell binary (`sh`/`bash`/`dash`/…), the data could be executed (data piped into a shell), so redaction is skipped and the raw catastrophic scan still blocks the payload. `guard-destructive.sh` (basename is not a bare shell word) is deliberately not matched, so the guard's own self-test pipe redacts.
- Wired into **both** the catastrophic (`COMMAND_NO_LITERAL_TEXT`) and ask (`COMMAND_ASK_SCAN`) working copies, gated on `echo`/`printf` presence to stay off the hot path. Neither feeds a scan that could widen a deny.

I deliberately chose the **structural redaction** over an allow-all `GUARD_SELFTEST` env escape hatch: the broadened redaction already satisfies the "piping a self-test payload works" criterion (the outer guard now allows piping a self-test payload into `guard-destructive.sh`), and an env toggle that disables the guard is itself a bypass risk.

- [x] Dangerous pattern inside a single-/double-quoted `echo`/`printf` argument no longer triggers a catastrophic deny (verified: single/double-quoted, `printf`, multi-line prose, and the exact JSON-piped self-test repro all allow).
- [x] Safety floor unchanged: `$(…)`/backtick smuggling still denies; `bash -c`/`sh -c` payloads still deny; a data sink piped into a shell still denies (shell-segment gate); a real command in a separate segment still denies.
- [x] Piping a self-test payload into the guard works via the broadened redaction.
- [x] Regression cases added for both repro shapes (echo-piped JSON payload; multi-line documentation literal), plus the ask tier and the full safety floor — 17 new cases.
- [x] No coverage regression: full suite green.

## Test Plan

- [x] `hooks/repo/tests/test-guard-destructive.sh`: **461 pass / 0 fail** (was 444; +17).
- [x] `hooks/repo/tests/run.sh`: **549 pass / 0 fail** (was 532; +17).
- [x] `bash -n` clean on both files.

### Noted out-of-scope limitation

A multi-line quoted literal whose **interior line begins** with a recursive-force root delete still denies, because `extract_rm_targets()` reads the raw command per-line. This is a pre-existing limitation shared identically with `--body` (the #3898 tests place prose before the danger for the same reason); it is a separate, deeper issue and out of scope here.

Closes #53

### DIFF
--- a/hooks/repo/guard-destructive.sh
+++ b/hooks/repo/guard-destructive.sh
@@ -1089,6 +1089,142 @@ strip_literal_text() {
     }'
 }
 
+# Redact the quoted argument(s) of a non-executing "data sink" command word
+# (echo, printf) so a dangerous-looking string that appears ONLY as quoted DATA
+# handed to echo/printf no longer trips the raw ALWAYS_BLOCK_PATTERNS scan
+# (catastrophic tier) or the ASK_PATTERNS scan (ask tier) (#53). echo/printf
+# print their arguments verbatim; they never EXECUTE them, so a quoted argument
+# is inert text — exactly like a --body/-m value — yet strip_literal_text()'s
+# flag allowlist never covered it. This is the meta false-positive that blocked
+# a guard self-test (`echo '{"…":"<dangerous cmd>"}' | guard-destructive.sh`)
+# and blocked filing this very issue's heredoc body.
+#
+# Command-word anchored (mirrors fastpath_builtin_admits() and the segment
+# parsers): the quoted args are redacted ONLY for a simple command whose FIRST
+# token is exactly `echo` or `printf` (optionally behind a bare `sudo`/`env`
+# wrapper). A wrapper that actually EXECUTES its argument — `bash -c '<payload>'`,
+# `sh -c`, `eval`, `xargs` — is never a data sink and is never redacted here.
+#
+# Safety floor, identical to strip_literal_text()/qsplit():
+#   - A quoted span is redacted ONLY when it carries no command substitution /
+#     backtick opener (`$(` or a backtick), so a smuggled `echo "$(<payload>)"`
+#     keeps its payload intact and still hard-denies.
+#   - The `echo '<payload>' | sh` shape (data PIPED into a shell that WOULD
+#     execute it) is handled by the command_has_shell_segment() gate at the call
+#     site, which skips this redaction entirely whenever any pipeline segment's
+#     command word is a shell — so the raw scan still sees and blocks the payload.
+#
+# Single-pass quote-aware lexer. It deliberately does NOT reuse qsplit(), whose
+# `\n`-per-separator contract would conflate a real newline inside a multi-line
+# quoted span with a shell separator; here a multi-line span is redacted as one
+# inert unit (`.` never matches a newline, so each line stays SAME-LENGTH and the
+# surrounding byte offsets are preserved). Best-effort like strip_literal_text():
+# an unterminated quote copies the remainder verbatim (never redacts), and the
+# result feeds only the NARROWING scans, so the worst case is a raw substring
+# surviving (a false block) — never a catastrophic block being skipped.
+strip_datasink_literals() {
+    printf '%s' "$1" | awk '
+    BEGIN {
+        SQ = sprintf("%c", 39)   # single quote
+        DQ = sprintf("%c", 34)   # double quote
+        buf = ""
+    }
+    { buf = buf (NR > 1 ? "\n" : "") $0 }
+    END {
+        s = buf
+        n = length(s)
+        out = ""
+        i = 1
+        atcmd = 1     # at the start of a simple command (command-word position)
+        sink = 0      # inside an echo/printf data-sink command
+        while (i <= n) {
+            c = substr(s, i, 1)
+            # A shell separator resets to command-word position.
+            if (c == ";" || c == "&" || c == "|" || c == "\n") {
+                out = out c; i++; atcmd = 1; sink = 0; continue
+            }
+            # Leading whitespace is copied without leaving command-word position.
+            if (c == " " || c == "\t") { out = out c; i++; continue }
+            # Command-word position: read the first token and classify it.
+            if (atcmd) {
+                atcmd = 0
+                tok = ""
+                j = i
+                while (j <= n) {
+                    cc = substr(s, j, 1)
+                    if (cc == " " || cc == "\t" || cc == ";" || cc == "&" || cc == "|" || cc == "\n") break
+                    tok = tok cc
+                    j++
+                }
+                # A bare sudo/env wrapper: emit it and stay in command-word
+                # position so the NEXT token is classified as the command word.
+                if (tok == "sudo" || tok == "env") {
+                    out = out tok; i = j; atcmd = 1; continue
+                }
+                if (tok == "echo" || tok == "printf") { sink = 1 }
+                out = out tok; i = j; continue
+            }
+            # Mid-command: a quoted span is redacted only inside a data sink.
+            if (c == DQ || c == SQ) {
+                qc = c
+                ci = 0
+                for (j = i + 1; j <= n; j++) {
+                    if (substr(s, j, 1) == qc) { ci = j; break }
+                }
+                if (ci == 0) {
+                    # Unterminated quote: copy the rest verbatim, never redact.
+                    out = out substr(s, i); i = n + 1; continue
+                }
+                inner = substr(s, i + 1, ci - i - 1)
+                if (sink && index(inner, "$(") == 0 && index(inner, "`") == 0) {
+                    gsub(/./, "X", inner)   # . never matches \n: multi-line stays same-length
+                }
+                out = out qc inner qc; i = ci + 1; continue
+            }
+            out = out c; i++
+        }
+        printf "%s", out
+    }'
+}
+
+# Return 0 (success) if ANY quote-aware segment's command word is a shell binary
+# (sh/bash/dash/zsh/ksh/csh/tcsh/fish/pwsh, with or without a leading path).
+# GATES strip_datasink_literals(): when a shell could consume the command's data
+# (e.g. `echo '<payload>' | sh`), the data-sink redaction is skipped so the raw
+# catastrophic scan still sees — and blocks — the payload. Conservative by
+# construction: a shell ANYWHERE in the command disables the (narrowing)
+# redaction, so the worst case is a preserved false BLOCK, never a skipped one.
+# `guard-destructive.sh` (basename is not a bare shell word) is deliberately NOT
+# matched, so the guard's own `echo '<json>' | guard-destructive.sh` self-test
+# still redacts and no longer false-blocks (#53). Emits "yes"/"no".
+command_has_shell_segment() {
+    printf '%s' "$1" | awk "$_QSPLIT_AWK"'
+    { buf = buf (NR > 1 ? "\n" : "") $0 }
+    END {
+        found = 0
+        s = qsplit(buf)   # quote-aware segmentation (#3755); separators -> \n
+        n = split(s, segs, "\n")
+        for (i = 1; i <= n; i++) {
+            seg = segs[i]
+            sub(/^[ \t]+/, "", seg)
+            # Strip any run of leading VAR=val assignments, then a sudo/env wrapper,
+            # so the REAL command word is classified. The required trailing [ \t]+
+            # in the assignment pattern guarantees the loop makes progress.
+            while (match(seg, /^[A-Za-z_][A-Za-z0-9_]*=[^ \t]*[ \t]+/)) { seg = substr(seg, RLENGTH + 1) }
+            sub(/^sudo[ \t]+/, "", seg)
+            sub(/^env[ \t]+/, "", seg)
+            sub(/^[ \t]+/, "", seg)
+            m = split(seg, toks, /[ \t]+/)
+            if (m == 0) continue
+            w = toks[1]
+            sub(/.*\//, "", w)   # basename only
+            if (w == "sh" || w == "bash" || w == "dash" || w == "zsh" || \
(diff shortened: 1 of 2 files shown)


rjwalters:
## Judge review: APPROVE

Reviewed as security-sensitive guard logic. Checked out the PR head (f47aa32) in its worktree (`.loom/worktrees/issue-53`), ran both suites, and hand-probed the safety floor myself through the guard's JSON stdin interface. Not a single deny was widened into an allow.

### Suites (worktree, matching PR claims)
- `test-guard-destructive.sh`: **461 pass / 0 fail** (+17)
- `run.sh`: **549 pass / 0 fail** — the guard counts fold in via `suite_count`, so the new cases integrate with no brittle hardcoded self-check to break.
- `bash -n` clean on both files; `merge-tree` shows no conflicts (mergeStateStatus CLEAN, MERGEABLE).

### Safety floor — probes I personally ran (all correctly DENY)
- Command substitution and backtick smuggled inside an `echo` quoted arg — the `$(`/backtick guard keeps the span intact.
- `bash -c '<payload>'`, `sh -c '<payload>'`, `sudo bash -c '...'`, `env FOO=1 bash -c '...'` — wrappers that execute are never data sinks.
- `echo '<payload>' | sh`, `| bash`, `printf '...' | dash` — the `command_has_shell_segment()` gate skips redaction so the raw scan still blocks.
- Segment-scoped: `echo 'hi' && <danger>`, `; <danger>`, `|| <danger>`, `echo x | <danger>`, `echo 'ok' | tee f ; <danger>`, and a newline-separated danger line — echo redaction never bleeds across separators.
- Anti-spoof: `./echo '<danger>'` (path/alias trick) still denies — first-token match is exactly `echo`/`printf`; an unterminated quote spanning a separator copies verbatim and denies; escaped-quote form denies.
- Plain unquoted danger still denies.

### False-positive fix — probes (all correctly ALLOW)
- The exact live repro: `echo '{...}' | .loom/hooks/guard-destructive.sh` self-test.
- Single/double-quoted `echo` arg, `printf '%s' '<danger>'`, `sudo/env echo`, multi-line documentation prose, and a "git push --force origin main" phrase quoted as data.

### Note (not blocking)
A `$(which rm) -rf /` form allows — but this is a pre-existing command-substitution-as-command-name limitation: I confirmed it allows **identically on main**, and it is definitionally untouched by this PR (the new code is gated on `echo`/`printf` presence, which this command lacks). Out of scope, correctly excluded.

Scope is tight (only the two intended files, +255/-0), the source-of-truth file is edited (not the vendored generic copy), and the design is command-word-anchored / narrowing-only per the file's stated philosophy. Nice work on the conservative shell-segment gate. Approving.

PR #61, opened 2026-07-29
fix(install): gate CLAUDE.md pointer on commands/ ignore state alone

### LINKED ISSUE #51: install.sh: gitignore check misses negated patterns (!.claude/commands/) — CLAUDE.md pointer wrongly skipped
## Context
Observed installing Repo Skills 0.6.1 into rjwalters/claude-monitor (2026-07-28). That repo's `.gitignore` has:

```
.claude/*
!.claude/commands/
```

so `.claude/commands/repo/*.md` is **tracked** (and was committed there for versions 0.4.x–0.6.1). The installer nevertheless printed:

```
⚠ Install destination (.claude/commands, .claude/skills) is gitignored in <repo>;
⚠ skipping the CLAUDE.md pointer block (a committed pointer to uncommitted command
⚠ files is not what you want). The /repo:* commands still work in-session.
```

The stated rationale doesn't hold in this repo — the command files *are* committed, so a committed CLAUDE.md pointer would be exactly right. The gitignore probe presumably tests `.claude/` (or `.claude/commands` without the trailing-path nuance) and doesn't account for negation patterns.

- [ ] Use `git check-ignore` on a representative file path (e.g. `.claude/commands/repo/help.md`), which resolves negations correctly, rather than pattern-matching the directory
- [ ] Mixed state handled: commands tracked but skills gitignored → pointer block still written (it points at the commands)
- [ ] Regression test: repo with `.claude/*` + `!.claude/commands/` gets the pointer block installed

---

### Verified root cause (corrects the issue title's premise)

Reproduced locally with a scratch repo containing the reported `.gitignore`:

```
.claude/*
!.claude/commands/
```

`git check-ignore` **already resolves the negation correctly** — this is not a "negated patterns are missed" bug in the literal sense:

```
$ git check-ignore -q .claude/commands/repo ; echo $?
1   # NOT ignored — negation resolved correctly
$ git check-ignore -q .claude/skills/repo ; echo $?
0   # ignored — no negation for skills/
```

`install.sh`'s actual probe (`install.sh:487-490`) already uses `git check-ignore -q` (not directory-pattern matching) on two separate paths:

```bash
dest_is_gitignored() {
  git -C "$TARGET" check-ignore -q .claude/commands/repo 2>/dev/null \
    || git -C "$TARGET" check-ignore -q .claude/skills/repo 2>/dev/null
}
```

The real defect is the **`||` (OR) across the two independent probes**: `dest_is_gitignored` returns true — and the whole CLAUDE.md pointer block is skipped — if *either* `.claude/commands/repo` *or* `.claude/skills/repo` is ignored, even when only one of the two is actually ignored. In the reported repo, `commands` is tracked (not ignored) but `skills` has no negation and is still ignored by `.claude/*`, so the OR trips and the pointer is wrongly skipped. The existing code comment at `install.sh:483-486` states this OR is intentional ("skip if *either* is ignored, since a split state is itself broken") — that premise is what this issue disproves with a real-world repo (rjwalters/claude-monitor 0.6.1), where the split state (`commands/` tracked, `skills/` gitignored) is legitimate, not broken.

Confirmed with the exact reported gitignore:

```
$ git check-ignore -q .claude/commands/repo || git check-ignore -q .claude/skills/repo; echo $?
0   # dest_is_gitignored() == true -> pointer wrongly skipped
```

### Implementation guidance

`install.sh`'s CLAUDE.md pointer block (written at `install.sh:501-514`) references both `/repo:*` commands and `.claude/skills/repo/SKILL.md`, so there's a real question of which directory should gate the decision:

- **Option A (minimal, matches the issue's suggested ACs)**: Gate skip-vs-write on `.claude/commands/repo` alone, since that's what makes `/repo:*` commands work from a fresh clone (the primary functional path called out in the warning text itself: "The /repo:* commands still work in-session"). If `skills/` is separately gitignored, still write the pointer — optionally soften the SKILL.md reference in the block text, or leave it as-is since `.claude/skills/repo/SKILL.md` degrades gracefully (file just won't exist post-clone; only the link in CLAUDE.md is stale).
- **Option B**: Keep OR-based skip, but only for a *fully* gitignored destination (both commands and skills ignored); add a third "mixed state" branch that writes the pointer with a caveat note about `skills/` being machine-local.

Option A is simpler, directly satisfies the issue's suggested acceptance criteria, and matches the warning text's own framing (which focuses on `/repo:*` commands, not SKILL.md). Recommend Option A unless the SKILL.md linkage is considered load-bearing enough to warrant Option B's extra branch.

Also update the probe to use a representative *file* path (e.g. `.claude/commands/repo/help.md`) instead of the bare directory, per the issue's first suggested AC — behaviorally equivalent in the reproduction above but avoids relying on git's directory-negation edge cases for less common `.gitignore` shapes (e.g. a nested `.gitignore` inside `.claude/commands/`).

### Affected Files

- `install.sh` - `dest_is_gitignored()` (`install.sh:487-490`) and the surrounding skip/write branch (`install.sh:479-499`); update the OR logic per the implementation guidance above, and update the warning message text if the skip condition narrows to commands-only
- `hooks/repo/tests/test-install-claude-md-markers.sh` - add a regression test case using a `.claude/*` + `!.claude/commands/` gitignore (mixed state) verifying the CLAUDE.md REPO-SKILLS pointer block *is* written; existing gitignore coverage in this file (around line 339-377) only exercises the fully-gitignored (`.claude/`) case

### Test Plan

- [ ] Automated test: add a case to `hooks/repo/tests/test-install-claude-md-markers.sh` with `.gitignore` containing `.claude/*` + `!.claude/commands/`, run `install.sh -y <target>`, assert the REPO-SKILLS block IS written to CLAUDE.md (byte-comparison assertion, consistent with the existing style in that file)
- [ ] Automated test: keep/extend existing coverage for the fully-gitignored case (`.claude/` alone) to assert the block is still correctly skipped (no regression)
- [ ] Manual verification: reproduce locally per the "Verified root cause" repro steps above (`git init`, write the two-line gitignore, run `install.sh`) before and after the fix
- [ ] Run the full existing suite: `bash hooks/repo/tests/test-install-claude-md-markers.sh`

### Related open work

`check-duplicate.sh --include-merged-prs --issue 51` surfaced #52 under `RELATED_OPEN_WORK`. Disregarded — #52 (repo:remote scriptable provisioning) cross-references "#51" only inside its own curator's comment as an example of the duplicate-scorer's false positives (`#51/#35/#31/#29/#13` all score 100% similarity against #52 despite being unrelated `install.sh`/`guard-destructive` issues); it is not substantive related work for this issue's gitignore-probe scope.

`DUPLICATE_FOUND` results (#35, #31 closed, #29 closed, #13 closed) are the same scorer false-positive pattern — reviewed each; none overlap this issue's scope (#31, closed, addressed *orphaned-block reconciliation* when fully gitignored, a different code path (`reconcile_orphaned_block`) from this issue's *wrongly-skipped-pointer-on-mixed-state* defect).

rjwalters:
## Summary

`install.sh`'s `dest_is_gitignored()` OR'd two `git check-ignore` probes (`commands/` OR `skills/`), so a legitimate split state wrongly skipped the CLAUDE.md pointer block. In a repo with:

```
.claude/*
!.claude/commands/
```

`commands/` is tracked (negated back in) but `skills/` is still ignored by `.claude/*`. The OR tripped on `skills/` and skipped the pointer — even though the `/repo:*` commands the pointer advertises are committed, making the pointer correct. Observed installing Repo Skills 0.6.1 into rjwalters/claude-monitor.

Per the Curator's recommended Option A, the skip decision now gates on the **commands** destination alone (that is what makes `/repo:*` work from a fresh clone), and probes a representative *file* path (`.claude/commands/repo/help.md`) so git resolves `!` negations and nested `.gitignore` files exactly as it would for a real committed file. The `skills/repo/SKILL.md` reference in the block degrades gracefully when `skills/` is machine-local (only the link goes stale).

## Changes

- `install.sh` — `dest_is_gitignored()` now probes `.claude/commands/repo/help.md` alone instead of OR-ing the commands and skills directory paths; comment rewritten to explain the split-state rationale; skip warning text narrowed to `.claude/commands`.
- `hooks/repo/tests/test-install-claude-md-markers.sh` — new "gitignore gate" section: a mixed-state repo (`.claude/*` + `!.claude/commands/`) asserts the REPO-SKILLS block **is** written; a fully-gitignored (`.claude/`) repo asserts it is still **skipped** (no regression).

- [x] Uses `git check-ignore` on a representative file path (`.claude/commands/repo/help.md`) rather than pattern-matching the directory.
- [x] Mixed state handled: commands tracked but skills gitignored -> pointer block still written. Verified by the new automated test and a manual `git init` reproduction of the reported gitignore.
- [x] Regression test: repo with `.claude/*` + `!.claude/commands/` gets the pointer block installed.
- [x] Fully-gitignored case still correctly skips the block (no regression).

## Test Plan

- `bash hooks/repo/tests/test-install-claude-md-markers.sh` -> 59/59 pass (10 new cases).
- `CI=true pnpm test` -> 756/756 pass, 0 fail.
- Manual reproduction: `git init` a scratch repo with the two-line gitignore, `install.sh -y` -> "Appended REPO-SKILLS block to CLAUDE.md" and the block is present.

Closes #51

### DIFF
--- a/hooks/repo/tests/test-install-claude-md-markers.sh
+++ b/hooks/repo/tests/test-install-claude-md-markers.sh
@@ -378,6 +378,62 @@ assert_bytes "reconcile non-adjacent: whole-line removal (historical behavior)" 
 
 # ---------------------------------------------------------------------------
 echo ""
+echo "-- install.sh gitignore gate: the pointer block is skipped only when commands/ is ignored --"
+
+# Mixed state (repo#51): commands/ tracked via a negation, skills/ still ignored.
+# The pointer advertises the /repo:* commands, which ARE committed here, so the
+# block MUST be written. An earlier OR'd probe (commands OR skills) wrongly
+# skipped it because skills/ is ignored.
+T12="$(new_target gitignore-mixed-commands-tracked)"
+printf '%s\n' '.claude/*' '!.claude/commands/' > "$T12/.gitignore"
+cat > "$T12/CLAUDE.md" <<'EOF'
+Project notes.
+
+<!-- BEGIN LOOM ORCHESTRATION -->
+Loom content.
+<!-- END LOOM ORCHESTRATION -->
+EOF
+
+# Sanity-check the fixture reproduces the reported split state before asserting
+# on install behavior: commands/ negated back in, skills/ still ignored.
+assert_eq       "gitignore mixed: commands/ is NOT ignored (negation resolved)" \
+                "1" "$(git -C "$T12" check-ignore -q .claude/commands/repo/help.md; echo $?)"
+assert_eq       "gitignore mixed: skills/ IS ignored" \
+                "0" "$(git -C "$T12" check-ignore -q .claude/skills/repo; echo $?)"
+
+OUT12="$(bash "$INSTALL_SH" -y "$T12" 2>&1)"; RC12=$?
+BODY12="$(slurp "$T12/CLAUDE.md")"
+assert_eq       "gitignore mixed: install exits 0" "0" "$RC12"
+assert_not_contains "gitignore mixed: does NOT skip the pointer block" \
+                "$OUT12" "skipping the CLAUDE.md pointer block"
+assert_contains "gitignore mixed: reports the append"     "$OUT12" "Appended REPO-SKILLS block to CLAUDE.md"
+assert_contains "gitignore mixed: REPO-SKILLS block IS written" \
+                "$BODY12" "<!-- BEGIN REPO-SKILLS -->"
+
+# Fully gitignored (`.claude/`): commands/ is machine-local, so a committed
+# pointer would advertise uncommitted command files. The block MUST be skipped.
+T13="$(new_target gitignore-fully-ignored)"
+printf '%s\n' '.claude/' > "$T13/.gitignore"
+cat > "$T13/CLAUDE.md" <<'EOF'
+Project notes.
+
+<!-- BEGIN LOOM ORCHESTRATION -->
+Loom content.
+<!-- END LOOM ORCHESTRATION -->
+EOF
+BEFORE13="$(slurp "$T13/CLAUDE.md")"
+
+OUT13="$(bash "$INSTALL_SH" -y "$T13" 2>&1)"; RC13=$?
+BODY13="$(slurp "$T13/CLAUDE.md")"
+assert_eq       "gitignore fully: install exits 0" "0" "$RC13"
+assert_contains "gitignore fully: skips the pointer block" \
+                "$OUT13" "skipping the CLAUDE.md pointer block"
+assert_not_contains "gitignore fully: no REPO-SKILLS block written" \
+                "$BODY13" "<!-- BEGIN REPO-SKILLS -->"
+assert_eq       "gitignore fully: CLAUDE.md untouched" "$BEFORE13" "$BODY13"
+
+# ---------------------------------------------------------------------------
+echo ""
 echo "-- guard: unresolvable marker layouts are refused, never guessed --"
 
 T9="$(new_target guard-unterminated)"
--- a/install.sh
+++ b/install.sh
@@ -490,20 +490,32 @@ fi
 
 CLAUDE_MD="$TARGET/CLAUDE.md"
 
-# If the install destination is gitignored in the target, the command/skill
-# files we just wrote are effectively machine-local (they won't be committed),
-# so a tracked CLAUDE.md pointer would advertise /repo:* commands whose files
-# aren't in the repo. Mirror dev mode here: skip the CLAUDE.md block entirely.
-# Probe each destination separately (check-ignore -q accepts only one pathname)
-# and skip if *either* is ignored, since a split state is itself broken. The
-# probe exits non-zero outside a git repo or when nothing is ignored, which
-# correctly falls through to the write path.
+# If the commands destination is gitignored in the target, the command files we
+# just wrote are effectively machine-local (they won't be committed), so a
+# tracked CLAUDE.md pointer would advertise /repo:* commands whose files aren't
+# in the repo. Mirror dev mode here: skip the CLAUDE.md block entirely.
+#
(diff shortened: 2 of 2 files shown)


rjwalters:
✅ **Approved** — LGTM.

**Verified independently (in `.loom/worktrees/issue-51`):**
- Marker suite: `bash hooks/repo/tests/test-install-claude-md-markers.sh` → **59/59 pass** (10 new gitignore-gate cases).
- Full suite: `CI=true pnpm test` → **756/756 pass, 0 fail**.
- Manual reproduction of the reported `.claude/*` + `!.claude/commands/` split state (`git init` scratch repo): `check-ignore` returns rc=1 for `.claude/commands/repo/help.md` (not ignored) and rc=0 for `.claude/skills/repo` (ignored); `install.sh -y` prints "Appended REPO-SKILLS block to CLAUDE.md" and the block is present. The acceptance-criteria defect is fixed.
- No-regression: the fully-gitignored (`.claude/`) case still prints the skip warning and leaves CLAUDE.md untouched (no REPO-SKILLS block).
- CI: `pnpm test` check passes; merge state CLEAN.

**Assessment:**
- **Correctness** — the fix is exactly Curator's recommended Option A: the skip decision now gates on the commands destination alone (what actually makes `/repo:*` work from a fresh clone) and probes a representative *file* path (`.claude/commands/repo/help.md`), so git resolves `!` negations and nested `.gitignore` files as it would for a real committed file. The `skills/repo/SKILL.md` reference degrades gracefully when skills/ is machine-local (stale link only).
- **Scope** — tightly bounded to `install.sh` + the markers test, no collateral changes.
- **Quality** — the rewritten comment clearly documents the split-state rationale and the #51 root cause; the warning text was correctly narrowed to `.claude/commands`. New tests sanity-check the fixture's ignore state before asserting on install behavior, which makes them robust and self-documenting.

Closes #51.

PR #62, opened 2026-07-29
docs(changelog): announce CLAUDE.md marker refusal conditions from #42

### LINKED ISSUE #45: docs(release): announce the new CLAUDE.md marker refusal conditions from #42
## Context

PR #42 (issue #38) fixed a real data-loss bug: `install.sh` was silently deleting an adjacent tool's `CLAUDE.md` marker block when `<!-- END REPO-SKILLS -->` shared a physical line with the next block's `BEGIN`. The fix anchors block surgery to the marker **string** rather than the whole line, and adds a validation guard.

The guard is correct and desirable, but it **widens the set of inputs that now hard-fail an install**, and that is a user-visible behavior change that has not been announced anywhere.

## The change in behavior

`claude_md_block_validate` refuses to rewrite when the marker layout is unresolvable — specifically when the begin-marker count is not exactly 1, the end-marker count is not exactly 1, or the end marker appears before the begin marker. On refusal it warns naming the file and exits non-zero.

The consequence worth flagging: a consumer `CLAUDE.md` that mentions `<!-- BEGIN REPO-SKILLS -->` **as prose or inside a code fence** — documenting the convention, say, or pasting an example — now counts as a duplicate marker. That repo's install fails with exit 1 where it previously would have proceeded (mangling or no-op'ing).

Failing loudly is the right call and matches the repo's stated "apply safe fixes, gate destructive ones" principle. This issue is not asking to revert it. It is asking that people find out from a release note rather than from a broken install.

## Second-order detail

`reconcile_orphaned_block`'s refusal path calls `error()` **after** skills and commands have already been installed, so a refused install is a partial install. The message says so explicitly, but it is worth stating in the release note that the remedy is to fix the markers and re-run, not to assume nothing was written.

- [ ] CHANGELOG entry for the next release describes: (a) the data-loss fix, (b) the new refusal conditions, (c) that a `CLAUDE.md` quoting the markers in prose or a code fence will now be refused, and (d) that a refusal from `reconcile_orphaned_block` leaves a partial install.
- [ ] The remedy is stated plainly — correct the markers so exactly one BEGIN and one END exist in order, then re-run.
- [ ] Consider whether the refusal message itself should name the remedy inline, since that is where an affected user actually reads it.
- [ ] Decide whether code-fence-aware marker detection is worth doing (explicitly out of scope in #42). If yes, file it separately; if no, say so in the release note so the prose-quoting limitation is documented rather than latent.

## Also worth folding in — a test gap from the same review

`claude_md_block_validate`'s out-of-order branch (END appears before BEGIN) has **no test**. The guard suite covers missing-end and duplicated-begin only. Small gap, cheap to close, and it is one of the three conditions that trigger the new refusal this note is documenting.

## Related

- #38 / PR #42 — the fix that introduced the refusal

rjwalters:
## Summary

Adds an `## Unreleased` section to `CHANGELOG.md` announcing the user-visible behavior change introduced by PR #42 (issue #38). PR #42 fixed a real data-loss bug (whole-line marker surgery could delete an adjacent tool's marker block) by anchoring block surgery to the marker string and adding a validation guard — but the guard **widens the set of inputs that now hard-fail an install**, and that was undocumented. This PR documents it so people learn it from a release note rather than a broken install.

## Changes

- `CHANGELOG.md`: new `## Unreleased` section covering all four acceptance-criteria points:
  - **(a)** the data-loss fix (marker-string-anchored surgery + validation guard).
  - **(b)** the three new refusal conditions (BEGIN count ≠ 1, END count ≠ 1, END before BEGIN).
  - **(c)** the known limitation — a CLAUDE.md quoting the markers as prose or in a code fence now counts as a duplicate marker and is refused; fence-aware detection is intentionally **not** built (documented as a limitation with a workaround: alter the literal marker, e.g. a non-ASCII hyphen).
  - **(d)** a refusal from `reconcile_orphaned_block` leaves a **partial** install (skills/commands already written), with the remedy stated plainly: fix the markers to exactly one ordered BEGIN/END pair, then re-run.

## Scope note (this wave)

Per my dispatch instructions, this PR is **scoped to changelog/release docs only**. The issue's Curator-listed affected files also include `install.sh`, `uninstall.sh`, and `hooks/repo/tests/test-install-claude-md-markers.sh` — those are owned by other builders this wave, so they are intentionally **not** touched here:

- **`install.sh` / `uninstall.sh` inline-remedy wording** (AC item 3, at `install.sh:84-87`/`522-526` and `uninstall.sh:150`) — deferred; the release note already states the remedy so an affected user has a documented path even before the message wording lands.
- **T12 out-of-order guard test** (the END-before-BEGIN branch at `lib/claude-md-block.sh:101-104` has no test) — deferred to the builder owning the test file this wave. The existing marker suite (49 cases) still passes.

## Release-mechanism finding (for the release driver)

The Curator asked me to verify whether `/repo:release` folds an existing `## Unreleased` section into the next version heading. It does **not**: Phase 4 drafts a fresh entry from the `<prev-tag>..<tag>` git-log range and Phase 5 inserts it at the top of `CHANGELOG.md` — there is no logic that merges an `## Unreleased` block. **The release driver must fold this `## Unreleased` section into the next version heading by hand** so points (a)-(d) survive into the tagged release note. The Phase 1.5 completeness gate is **not** confused by the heading — it greps only for `^## v?[version]` tag headers, and `Unreleased` matches no tag, so it is neither flagged MISSING nor mismatched. Latest tag is v0.6.1 and the #42 fix is not yet tagged, so this note can still ship *with* the behavior change.

## Test Plan

- [x] `CI=true pnpm test` from the worktree — **PASS: 746, FAIL: 0** (includes `test-install-claude-md-markers.sh` 49/49).
- [x] `## Unreleased` entry covers all four points (a)-(d) and states the remedy.
- [x] Scope check: only `CHANGELOG.md` modified (`git diff --stat` = 1 file, 35 insertions).
- [x] Verified `/repo:release` Phase 1.5 gate greps tag version headers only (no `Unreleased` confusion).

Closes #45

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,40 @@
 # Changelog
 
+## Unreleased
+
+- **CLAUDE.md marker surgery is now anchored to the marker string, closing a
+  data-loss bug (#38, PR #42).** Previously `install.sh`/`uninstall.sh` bounded
+  the REPO-SKILLS block by whole physical lines, so when `<!-- END REPO-SKILLS
+  -->` shared one line with an adjacent tool's `<!-- BEGIN … -->`, block removal
+  could silently delete that neighbouring tool's marker block. Removal now
+  anchors to the marker **substring** and a validation guard
+  (`claude_md_block_validate`) gates every rewrite.
+- **New refusal conditions (user-visible behavior change).** The guard refuses
+  to rewrite — and, in `install.sh`, exits non-zero — when the marker layout is
+  unresolvable: the `<!-- BEGIN REPO-SKILLS -->` count is not exactly 1, the
+  `<!-- END REPO-SKILLS -->` count is not exactly 1, or the END marker appears
+  before the BEGIN marker. On refusal, CLAUDE.md is left byte-for-byte untouched
+  and the warning names the offending condition. This is deliberate: the repo
+  applies safe fixes automatically and gates destructive ones rather than
+  guessing at an ambiguous layout. **Installs that previously proceeded (mangling
+  or no-op'ing) against such a CLAUDE.md now fail loudly instead.**
+- **Known limitation — quoting the markers as prose or in a code fence now
+  triggers a refusal.** Marker counting is substring-based, so a consumer
+  CLAUDE.md that *mentions* `<!-- BEGIN REPO-SKILLS -->` (documenting the
+  convention, or pasting an example inside a code fence) counts as a duplicate
+  marker and its install fails with exit 1. Fence-aware detection is
+  intentionally **not** implemented — it would reintroduce the ambiguity-guessing
+  #42 removed. If you need to show the markers as an example, alter them so they
+  are not literal occurrences (e.g. a non-ASCII hyphen, `BEGIN REPO‑SKILLS`, or a
+  broken-up string).
+- **A refusal from `reconcile_orphaned_block` leaves a partial install.** That
+  refusal path (and the update path) fires *after* skills and commands have
+  already been installed, so a refused run is a partial install — the warning
+  already says "(skills and commands were installed)". The remedy is not to
+  assume nothing was written: **correct the markers so exactly one BEGIN and one
+  END exist, in that order, then re-run the installer** (which reconciles the
+  now-valid block). CLAUDE.md itself is never modified by a refused run.
+
 ## 0.6.1 (2026-07-27)
 
 - **`install.sh` now reconciles an orphaned CLAUDE.md pointer block (#31).**


rjwalters:
## Approved

Reviewed for factual accuracy, completeness against #45's acceptance criteria, and scope discipline. Verdict: **approve**.

### Factual accuracy (verified against PR #42's actual diff)
Every claim in the `## Unreleased` entry checks out against `lib/claude-md-block.sh`:
- **Three refusal conditions match the code exactly.** `claude_md_block_validate` returns 1 when `n_begin -ne 1`, when `n_end -ne 1`, and when the prefix before BEGIN already contains END (`'$end' appears before '$begin'`). The changelog's "BEGIN count is not exactly 1 / END count is not exactly 1 / END before BEGIN" is a precise restatement, not an approximation.
- **The prose/code-fence limitation is correctly grounded.** `_claude_md_count` counts substring occurrences via `awk index()` across the whole file (multiple per line included), so a CLAUDE.md that merely *mentions* a marker genuinely pushes the count to 2 and trips refusal with exit 1. The claim is accurate, and the "intentionally not fence-aware + alter-the-literal-marker" workaround is a fair characterization.
- **Partial-install claim confirmed.** `reconcile_orphaned_block` calls `error()` (non-zero) *after* skills/commands are written, and the actual message is "(skills and commands were installed)" — matching the changelog. The "CLAUDE.md is untouched" and "exits non-zero in install.sh" claims are both borne out by the warning/error lines.

### Completeness vs. acceptance criteria
All four core points (a)-(d) from #45's AC-1 are present, and AC-2 (remedy stated plainly: "exactly one BEGIN and one END, in that order, then re-run") is satisfied. AC-4's decision — whether to build fence-aware detection — is answered in the note itself (intentionally not implemented, limitation documented rather than left latent), which is exactly what the AC asked for.

### Scope discipline
Single-file diff (`CHANGELOG.md`, +35/-0). The deferred items — install.sh/uninstall.sh inline-remedy wording (AC-3) and the T12 out-of-order guard test — are **explicitly documented in the PR body as owned by other builders this wave**, not silently dropped. That is the correct handling: the release note already gives affected users a documented remedy path even before the inline message wording lands. The release-mechanism finding (that `/repo:release` does not fold an existing `## Unreleased` block and the driver must merge it by hand) is a useful, accurate hand-off note.

### Gate checks
- CI: `pnpm test` **pass** (746 cases, includes the 49-case marker suite).
- Merge state: **CLEAN**.
- Issue linking: `Closes #45` present.

Accurate, well-scoped, and CI-green. Approving.

PR #57, opened 2026-07-29
feat(remote): add headless repo-remote provisioning script

rjwalters:
## Summary

Extracts the `/repo:remote` provisioning contract from prose (`commands/repo/remote.md`) into an executable, non-interactive script — `scripts/repo/repo-remote.sh` — so loom's `fleet add-worker` can drive provisioning programmatically instead of a Claude session hand-transcribing 450 lines of prose (as happened in the 2026-07-28 pilot). The interactive skill now **delegates** to this single implementation rather than re-issuing cloud CLI calls, so the headless and interactive paths cannot drift (the third acceptance criterion).

The seam produced, consumed by loom: "a reachable Ubuntu box, repo-remote SSH alias written, instance id recorded" — emitted as machine-readable JSON so a caller can implement loom's "plan shown before money is spent" rule.

## Changes

- **`scripts/repo/repo-remote.sh` (new)** — the non-interactive implementation: two-layer config loading (`~/.config/repo/remote.env` shared → repo `.env` overrides), provider dispatch (aws/gcp), `up [--yes] [--json]` / `status` / `down [--yes] [--delete]` (both verb and remote.md-style `--status`/`--down` flag forms), `repo-remote=<name>` tagging, idle-shutdown guard via cloud-init user-data, `REPO_REMOTE_INSTANCE_ID` write-back to the repo `.env`, SSH-alias write, GPU-family detection + Deep Learning AMI selection + `VcpuLimitExceeded` quota remediation, and JSON output (instance id, public IP, SSH alias, estimated hourly cost).
- **`commands/repo/remote.md`** — documents the headless entry point and rewires the interactive steps + `--status`/`--down` to delegate to the shared script instead of issuing cloud CLI calls from prose.
- **`install.sh` / `uninstall.sh`** — ships the script to consumer repos at `.claude/skills/repo/scripts/repo-remote.sh` (`chmod +x`), mirroring the hook-install pattern; dry-run enumeration and uninstall listing updated.
- **`commands/repo/tests/test-repo-remote.sh` (new)** — 71-case suite (mock `aws` CLI + fixtures) covering the cost gate, config-layer precedence, JSON shape, GPU cost, reuse, write-back, quota remediation, and the up/status/down flow, plus a doc-drift assertion against `remote.md` and a packaging assertion against `install.sh`. `hooks/repo/tests/run.sh` delegates to it.
- **`README.md` / `skills/repo/SKILL.md`** — write-footprint, repository-layout, and `[[remote]]` description updated.

[...]

PR #64, opened 2026-07-29
test(branches): skip merge-tree assertions on git < 2.38 and cover the fallback

### LINKED ISSUE #46: test: branches loss-check suite fails on git < 2.38 and never covers the documented merge-tree fallback

rjwalters:
## Summary

The `test-branches-loss-check.sh` suite hard-failed on git < 2.38 and never covered the documented `merge-tree` fallback. The direct `git merge-tree --write-tree` probe swallowed the "unsupported" error under `2>/dev/null`, leaving an empty string that was then asserted against a tree hash — a confusing failure on exactly the git versions the fallback exists to support. Worse, the degraded path (`git diff --quiet` -> merged-PR lookup) had **zero** coverage, so the loss check's most dangerous failure mode — degrading to SAFE and silently deleting work (#39) — was untested.

This change makes the suite green on older git via a capability probe + skips, and adds git-version-independent positive coverage of the fallback chain that pins its failure direction (never SAFE unless containment is otherwise proven).

## Changes

- **Capability probe (not version parsing):** probe `merge-tree --write-tree` once on the fixture repo. The four assertions that need a real merge-tree result now `SKIP` with a reason instead of failing on git < 2.38.
- **Degraded-path coverage:** a new section forces the fallback chain via a subshell-scoped PATH shim (a `git` wrapper that fails only on `merge-tree`, execs real git otherwise) and asserts:
  - identical trees -> `SAFE` via `git diff --quiet` (legitimate degraded success);
  - main-advanced squash-merge, no PR -> `UNSAFE` (never SAFE);
  - same, forge lookup down -> `KEEP` (never SAFE);
  - a merged PR still clears the branch without merge-tree.
  A new `feature/identical` fixture (amended commit: main's tree, distinct SHA) drives the `diff --quiet` success arm. A hygiene assertion confirms the shim never leaks into later real-git cases.
- **Skip accounting:** a `Skipped: N` summary line in the suite; `run.sh` parses it and annotates the per-suite breakdown row (`— N skipped`) so a skipped run is visibly distinct from a full pass. Skips do not feed the pass/fail folding or the breakdown-sums-to-headline self-check (#44).
- **Docs:** one line in `branches.md` noting `--write-tree` requires git >= 2.38.

- [x] Suite detects `merge-tree --write-tree` support up front (probe on the fixture repo).
- [x] On unsupported git the merge-tree-dependent assertions skip with a clear message; `pnpm test` stays green (verified under a merge-tree-failing shim: suite exits 0, 4 skips).
- [x] Positive fallback coverage asserts the classification is KEEP/UNSAFE, never SAFE, when containment can't be proven.
- [x] Skip count surfaced in both the suite summary and the run.sh breakdown row.
- [x] `branches.md` states the minimum git version next to the command.

## Test Plan

- [x] Modern git (2.55.0): `pnpm test` -> 752 pass, 0 fail; branches suite 37 pass, **0 skipped**.
- [x] Simulated git < 2.38 (whole suite under a `merge-tree`-failing PATH shim): suite exits 0, 32 pass, **4 skipped**; run.sh breakdown row shows `branches.md loss check — 4 skipped`.
- [x] Fault injection: flipped the `no PR -> UNSAFE` fallback expectation to `SAFE`; suite went red naming the case (exit 1); reverted -> green.
- [x] Standalone run (`bash commands/repo/tests/test-branches-loss-check.sh`) works outside `pnpm test`.
- [x] Shim scoping: subshell-confined PATH; the "shim did not leak" assertion passes, later real-git assertions unaffected.

Closes #46

### DIFF
--- a/commands/repo/branches.md
+++ b/commands/repo/branches.md
@@ -154,6 +154,7 @@ already merged. Decide with **content containment**, not SHA ancestry:
 
 ```bash
 # Would merging the branch into <default> change anything at all?
+# `--write-tree` requires git >= 2.38; on older git this errors and you take the fallback below.
 [ "$(git merge-tree --write-tree <default> <branch>)" = "$(git rev-parse '<default>^{tree}')" ]
 ```
 
--- a/commands/repo/tests/test-branches-loss-check.sh
+++ b/commands/repo/tests/test-branches-loss-check.sh
@@ -33,10 +33,12 @@ RESET_MD="$REPO_ROOT/commands/repo/reset.md"
 
 PASS=0
 FAIL=0
+SKIP=0
 TOTAL=0
 
 RED='\033[0;31m'
 GREEN='\033[0;32m'
+YELLOW='\033[0;33m'
 NC='\033[0m'
 
 if [[ ! -f "$BRANCHES_MD" ]]; then
@@ -58,6 +60,17 @@ ok() {   # <label>
 no() {   # <label> <detail>
     TOTAL=$((TOTAL + 1)); FAIL=$((FAIL + 1))
     printf "  ${RED}FAIL${NC}: %s\n" "$1"
+    [[ -n "${2:-}" ]] && printf "        %s\n" "$2"
+    return 0
+}
+# A skipped assertion is neither pass nor fail: the environment cannot exercise
+# it (e.g. this git predates `merge-tree --write-tree`). It is counted separately
+# and surfaced in the summary so a silently-skipped suite is never mistaken for a
+# full pass (repo#46). SKIP does NOT feed TOTAL/PASS/FAIL — run.sh folds only the
+# pass/fail counts, and reports the skip count as a breakdown annotation.
+skip() {  # <label> <reason>
+    SKIP=$((SKIP + 1))
+    printf "  ${YELLOW}SKIP${NC}: %s\n" "$1"
     [[ -n "${2:-}" ]] && printf "        %s\n" "$2"
     return 0
 }
@@ -213,6 +226,17 @@ build_fixtures() {
     echo mainonly > "$REPO/m.txt"
     git -C "$REPO" add -A && git -C "$REPO" commit -qm "MAINONLY: committed only to main"
     MAINONLY_SHA=$(git -C "$REPO" rev-parse --short HEAD)
+
+    # (6) IDENTICAL-TREE branch: same content as main's current tip but a
+    #     distinct SHA (an amended commit, same tree + same parent). Its tree
+    #     equals main's, so even the degraded `git diff --quiet` fallback can
+    #     prove containment — this is the legitimate "old git still classifies
+    #     SAFE" path exercised by the degraded-mode section below (repo#46).
+    #     Built last so main's tree is fixed from here on; leaving main checked
+    #     out keeps every later assertion operating from main.
+    git -C "$REPO" checkout -q -b feature/identical
+    git -C "$REPO" commit -q --amend --allow-empty -m "identical: main's tree, distinct SHA"
+    git -C "$REPO" checkout -q main
 }
 
 echo "branches.md permanent-loss check test suite"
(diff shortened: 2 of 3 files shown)


rjwalters:
✅ **Approved!**

Solid, well-scoped fix that fully satisfies issue #46's acceptance criteria. The most valuable part — positive coverage of the previously-untested degraded path that pins its failure *direction* (never SAFE unless containment is otherwise proven) — is exactly the loss-check protection #39 left uncovered.

**Branch integration (done during review):** the branch was BEHIND main by #61/#62/#63. PR #56 (breakdown self-check) and #57 (delegation entry) were already in the branch history, so the concern about a stale run.sh base did not materialize. I merged origin/main into the PR branch in the isolated worktree (clean, ort strategy, no conflicts) and confirmed the merge left all three of this PR's files byte-identical. Pushed as e3025b1; CI passed (1m0s), mergeStateStatus CLEAN.

**Verification performed:**
- Modern git (2.55.0): standalone suite 37 pass / 0 fail / 0 skipped; full `run.sh` 762 pass / 0 fail on the integrated result.
- Simulated git < 2.38 (whole suite under a `merge-tree`-failing PATH shim): standalone 32 pass / 0 fail / **4 skipped**; full `run.sh` 747 pass / 0 fail with breakdown row `branches.md loss check — 4 skipped`.
- **#56 self-check intact:** under the shim the breakdown still sums to the headline (55+461+79+49+32+71 = 747) — skips are correctly excluded from the PASS/FAIL folding, so the sums-to-headline invariant is not broken.
- Fault direction confirmed: the degraded section asserts UNSAFE/KEEP (never SAFE) for the cases the merge-tree path would call SAFE.
- `bash -n` clean on both scripts; final diff read critically — complete and coherent, no half-restored hunks from the reported mid-build mishap.

**Design notes I liked:**
- Capability probe on the fixture repo rather than parsing `git --version` (robust against vendor version strings).
- PATH shim confined to subshells, with an explicit "shim did not leak" hygiene assertion guarding later real-git cases.
- `skip()` kept strictly orthogonal to TOTAL/PASS/FAIL, with run.sh treating skips as a breakdown annotation only.

Closes #46. Nice work.

PR #65, opened 2026-07-29
feat(install): add opt-in shell claude wrapper surfacing pending handoff notes

### LINKED ISSUE #35: install.sh: shell `claude` wrapper to surface a pending handoff note to the human (deferred from #32)
## Context

Split out of #32. That issue was filed asking for **two** mechanisms to surface a pending `/repo:handoff` note at session start:

1. A `SessionStart` hook (repo-scoped, no files touched outside the repo) — **shipped** in #34 (`hooks/repo/session-start-handoff.sh`).
2. A shell `claude` wrapper that prints a banner in the terminal *before* Claude starts — **deferred**, and this issue tracks it.

The Curator's scope call on #32 was to ship the hook alone, on the grounds that #32 itself said the hook was "arguably the more correct mechanism" and "might be worth doing that first", and that bundling the wrapper's alias-parsing / rc-editing / shell-detection logic into the same PR would push it well past a reviewable size. That call was right for #34, but it means the thing named in #32's title was never built.

## What is still missing

The hook injects the note into **Claude's** context. It shows the **human** nothing before the session starts — which was a stated part of the original value: you quit, run `claude update`, relaunch, and there is no visible signal that the previous session left instructions.

A working zsh implementation already exists and is running in a consumer repo (2AMLogic/studio) as of 2026-07-27. It is reproduced verbatim in the body of #32 — start there rather than rewriting it. It prints a compact banner: note path, age, `##`-header outline, and a STALE warning past 7 days.

## Why this is the risky half

`install.sh` is otherwise scrupulous about staying inside the target repo — the CLAUDE.md block is marker-bounded, `settings.json` is jq-merged rather than clobbered, and #34's `SessionStart` wiring follows the same discipline. Editing `~/.zshrc` is categorically different: it is global, it affects every repo and every shell, and it changes the behaviour of a command the user runs constantly.

**Hard requirement** (carried over from #32, and the reason it was deemed vacuous for the hook but binding here): this MUST be opt-in, defaulting to **N**, with the diff shown first, and MUST be a strict no-op under `--yes` / non-interactive installs unless an explicit flag (e.g. `--shell-wrapper`) was passed. Unattended installs must never silently rewrite a shell rc.

## Design constraints (from #32, none addressed by #34)

These are the parts that are easy to get wrong. Constraint #4 was the only one that carried over to the hook and is already solved there; the rest are untouched:

1. **The existing `claude` may be an alias carrying flags.** The reporter's was `alias claude='command claude --dangerously-skip-permissions'`. A naive append silently loses the flag, or the alias shadows the new function depending on load order. The wrapper must parse the existing alias, preserve its flags, and **remove the alias** — aliases and functions of the same name do not compose.
2. **Recursion.** The function body must call `command claude`, never `claude`.
3. **Gate on subcommands.** `claude update`, `claude mcp`, `claude --version` are not sessions; a banner there is noise. See `_claude_is_session` in #32 for the minimum list.
4. ~~**`stat` is not portable** (`-f %m` BSD vs `-c %Y` GNU).~~ Solved in `hooks/repo/session-start-handoff.sh` — reuse that approach.
5. **Shell detection.** `$SHELL` / `$ZSH_VERSION` / `$BASH_VERSION`. The forms used in #32 are zsh+bash compatible; fish needs a separate translation and is probably better **skipped with a clear message** than half-supported.
6. **Idempotency + uninstall.** Wrap in `# BEGIN REPO-SKILLS CLAUDE WRAPPER` / `# END` markers, same convention as the CLAUDE.md block, so re-install replaces in place. This gives `uninstall.sh` its first reason to touch a file outside the repo — that path needs care of its own.
7. **Back up the rc file** before editing it.

## Open question for whoever picks this up

Given #34 shipped, is the banner still worth the global blast radius? The hook covers the "Claude reliably sees the note" half. This issue only buys the "human sees it before launch" half. A reasonable outcome is to close this as wontfix — but that should be a decision someone makes explicitly, not a thing that quietly never happens.

## Related

- #32 — original issue (contains the reference implementation and sample output)
- #34 — the shipped `SessionStart` hook

rjwalters:
## Summary

`install.sh` gains `--shell-wrapper`: an opt-in edit to the user's shell rc (`~/.zshrc` or `~/.bashrc`) that defines a `claude` function printing a compact banner — note path, age, a `##`-header outline, a STALE warning past seven days — when a pending `/repo:handoff` note exists, *before* Claude starts. This is the "human sees it" half of the handoff workflow; the `SessionStart` hook shipped in #34 already covers the "Claude sees it" half.

This was deferred out of #32/#34 specifically because it is the one thing the installer writes **outside** the target repository, so it ships behind a hard opt-in/backup/diff-preview contract rather than folded into the unconditional install steps.

## Changes

- `lib/shell-wrapper.sh` (new): shell detection (`$SHELL`/`$ZSH_VERSION`/`$BASH_VERSION`; fish is skipped with a clear message, not half-supported), single-line `alias claude=...` parsing (extracts flags, folds them into the new function's `command claude` call), and marker-bounded (`# BEGIN REPO-SKILLS CLAUDE WRAPPER` / `# END ...`) rc surgery with backup-before-edit, mirroring `lib/claude-md-block.sh`'s discipline.
- `install.sh`: new `--shell-wrapper` flag; a new gated step (3e) that is a **strict no-op** under `--yes` without the flag, otherwise always shows the pending rc diff first and offers a confirm (default **N**) unless the flag was passed explicitly. `--dry-run` and usage text updated.
- `uninstall.sh`: removes the block (backed up) from any candidate rc it finds it in (`~/.zshrc`, `~/.bashrc`, regardless of the shell active at uninstall time); no-op when absent.
- `hooks/repo/tests/test-shell-wrapper.sh` (new): 55-case suite — unit tests for detection/parsing, install.sh integration fixtures (alias preservation, idempotency, missing rc, fish skip, unparseable-alias bail, ambiguous-marker guard, `--dry-run`), uninstall round-trip, and **runtime** tests that actually source the rendered function block and exercise `_repo_claude_handoff_banner`/`_repo_claude_is_session`/`claude()` against fixture repos (fresh note, stale note, no note, non-git cwd, argv passthrough, missing-helpers fail-transparent path) — registered in `hooks/repo/tests/run.sh`.
- `README.md`: documents the flag, the default-off behavior, the alias-preservation contract, and the uninstall path.

### Design note: fail-transparent by construction

A comment on #35 reported a *different*, non-reference wrapper breaking `claude mcp list` in a non-interactive shell because a helper function it depended on (`_claude_build_argv`) wasn't defined, and the wrapper then mangled argv. This implementation avoids that failure mode structurally: every helper (`_repo_claude_is_session`, `_repo_claude_handoff_banner`) is defined in the *same* marker-bounded block as `claude()` (so a shell that sources the rc gets all of it or none of it), the banner call is gated behind a `type` check, and `command claude <preserved-flags> "$@"` runs unconditionally regardless of whether the banner fired — so argv is never altered by whether helpers are present. Covered by an explicit test (`missing helpers -> transparent passthrough, argv unchanged`).

| Criterion (from issue #35 / Curator comment) | Status | Verification |
|---|---|---|
| `--shell-wrapper` flag; interactive mode also offers via confirm (default N), diff shown first | ✅ | `install.sh` §3e; manually verified diff output + confirm gating |
| Strict no-op under `--yes` without `--shell-wrapper` (rc neither read nor written) | ✅ | `strict no-op: .zshrc byte-identical` test |
| Marker-bounded, idempotent block | ✅ | `exactly one marker pair after re-install`, `idempotent re-install (byte-identical)` |
| rc backed up before first edit | ✅ | `backup file exists` / `backup holds the pre-edit content` |
| Existing alias parsed, flags folded in, `unalias claude` precedes the function, user's alias line outside the markers never edited | ✅ | `alias-preserve` test block (5 assertions) incl. textual-order check |
| Function invokes `command claude` (no recursion) | ✅ | `claude() invokes 'command claude', passing argv through` |
| Banner gated on session-starting invocations only (`update\|doctor\|mcp\|config\|install\|migrate-installer\|setup-token`, `-v\|--version\|-h\|--help`) | ✅ | `_repo_claude_is_session gates the documented subcommand list` |
| Portable mtime (BSD `stat -f %m` / GNU `stat -c %Y`, reused from `session-start-handoff.sh`) | ✅ | Same fallback pattern in the rendered banner; runtime fresh/stale tests pass on this (macOS/BSD) host |
| zsh + bash supported; fish skipped with a clear message, rc untouched | ✅ | `fish-skip` test |
| `uninstall.sh` removes the block (backed up), no-op when absent | ✅ | `uninstall-roundtrip` (removal + byte-identical restore) + second-run no-op test |
| Banner matches #32 reference: path, age label, `##`-header outline (≤9 lines), STALE at ≥7 days | ✅ | Runtime tests: fresh note (header/path/age/outline, no STALE), stale note (STALE present) |
| README documents the flag, default-off behavior, uninstall path | ✅ | New "Optional: surface the note to the human too" section |

## Test Plan

- `pnpm test` (full suite, `hooks/repo/tests/run.sh`): new `test-shell-wrapper.sh` suite passes 55/55, folded into the aggregate per the existing delegated-suite pattern. The only failures observed anywhere in the full run are 8 flaky cases inside `test-guard-destructive.sh` (force-push/reset-hard/decision-log assertions), independently reproduced on `origin/main` with zero changes from this branch — pre-existing and unrelated to this PR.
- `shellcheck -x` on `install.sh`, `uninstall.sh`, `lib/shell-wrapper.sh`, `hooks/repo/tests/test-shell-wrapper.sh`, `hooks/repo/tests/run.sh`: no warnings beyond the same info-level notices (`SC2094`, `SC2016`, `SC1091` cross-file source) already present elsewhere in this codebase.
- Manual end-to-end runs against scratch `$HOME`/target-repo fixtures: `--yes` alone (no-op confirmed byte-identical), `-y --shell-wrapper` with a pre-existing `alias claude='command claude --dangerously-skip-permissions'` (flag preserved, backup created, idempotent re-run), `SHELL=fish` (clean skip, no rc created), and `uninstall.sh -y` (block removed, rc restored to original content).
- Rebased cleanly onto `origin/main` (post #61/#62/#63); re-ran the full suite after rebase (results above).

Closes #35

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,25 @@ Installing Repo Skills also wires a **SessionStart hook** (`session-start-handof
 
 It is strictly read-only: it never writes or deletes the note (absorbing it and removing it is `/repo:handoff`'s own one-shot contract), it stays silent when no note exists, and it fails open, so a hook error can never block session start. It deliberately does not fire on `/clear`, which is not a process relaunch.
 
+### Optional: surface the note to the human too (`--shell-wrapper`)
+
+The `SessionStart` hook above gets the note into **Claude's** context. It shows the **human** nothing before the session starts — you quit, relaunch, and there's no visible signal that the previous session left instructions. Pass `--shell-wrapper` to `install.sh` to also opt into a shell `claude` wrapper that prints a compact banner (note path, age, a `##`-header outline, and a STALE warning past seven days) *before* Claude starts.
+
+This is the **one thing the installer writes outside the target repo** — it edits your shell rc (`~/.zshrc` or `~/.bashrc`, detected from `$SHELL`/`$ZSH_VERSION`/`$BASH_VERSION`; fish isn't supported yet and is skipped with a clear message rather than half-supported) — so it is treated with more care than everything else installed:
+
+- **Off by default.** `--yes`/non-interactive installs are a **strict no-op** for this feature unless `--shell-wrapper` is also passed — the rc file is neither read nor written. An interactive install without the flag still *offers* it via a confirm (default **N**), always showing the exact pending diff first.
+- **Marker-bounded and idempotent**, same convention as the `CLAUDE.md` block (`# BEGIN REPO-SKILLS CLAUDE WRAPPER` / `# END …`) — re-running `--shell-wrapper` replaces the block in place rather than duplicating it.
+- **Backed up before every edit.** The path is reported after install/uninstall.
+- **Preserves an existing `claude` alias.** If your rc already has `alias claude='command claude --dangerously-skip-permissions'` (or similar), its flags are folded into the new function and `unalias claude` runs ahead of the function definition so the two don't fight over which wins. Your original alias line is left untouched outside the markers. Anything outside that common single-line form is left alone with a message rather than guessed at.
+- **Never recurses and never breaks non-interactive invocations.** The function always calls `command claude`, and always falls through to it unmodified — the banner helper is skipped (not fatal) if it's ever missing, so scripted calls like `claude mcp list` behave exactly like the real binary.
+
+```bash
+./install.sh --shell-wrapper ~/projects/my-app       # interactive: offers a confirm either way
+./install.sh -y --shell-wrapper ~/projects/my-app    # non-interactive opt-in, no prompt
+```
+
+`uninstall.sh` removes the block (from both `~/.zshrc` and `~/.bashrc`, whichever has it) automatically, with the same backup discipline, and is a no-op if it was never installed.
+
 ## Installation
 
 The installer copies the skill files into a target repository's `.claude/` directory, wires the guard hook into `.claude/settings.json`, and appends a marker-bounded section to its `CLAUDE.md`.
@@ -64,6 +83,10 @@ The installer copies the skill files into a target repository's `.claude/` direc
 
 # Dev mode: symlink source files for live editing (dogfooding)
 ./install.sh --dev .
+
+# Opt into the shell `claude` wrapper (see "Optional: surface the note to the
+# human too" above) — off by default, edits your shell rc
+./install.sh --shell-wrapper ~/projects/my-app
 ```
 
 To remove: `./uninstall.sh /path/to/target-repo`.
@@ -83,6 +106,8 @@ The installer is designed to coexist with whatever already lives in the consumer
 
 Nothing else in the target repository is read or modified.
 
+Opt-in only, and outside the target repository: with `--shell-wrapper`, your shell rc (`~/.zshrc` or `~/.bashrc`) gains one marker-bounded block (`# BEGIN REPO-SKILLS CLAUDE WRAPPER` / `# END …`), backed up first — see "Optional: surface the note to the human too" above.
+
 ## Repository layout
 
 ```
@@ -95,11 +120,13 @@ hooks/repo/tests/run.sh      Smoke suite covering both hooks (bash, no framework
 hooks/repo/tests/test-guard-destructive.sh  Full guard regression suite (ported from Loom)
 hooks/repo/tests/test-session-start-handoff.sh  Handoff-hook suite (run.sh delegates to it)
 hooks/repo/tests/test-install-claude-md-markers.sh  CLAUDE.md marker-block regression suite
+hooks/repo/tests/test-shell-wrapper.sh  claude shell wrapper suite (run.sh delegates to it)
 commands/repo/tests/test-branches-loss-check.sh  branches.md permanent-loss check suite (run.sh delegates)
 commands/repo/tests/test-repo-remote.sh  repo-remote.sh provisioning-contract suite (run.sh delegates)
 install.sh                   Installer
 uninstall.sh                 Uninstaller
 lib/claude-md-block.sh       Marker-bounded CLAUDE.md surgery shared by install.sh/uninstall.sh
+lib/shell-wrapper.sh         Opt-in claude shell wrapper (--shell-wrapper): detection, alias parsing, marker-bounded rc surgery
 ```
 
 ## Adding a skill
--- a/hooks/repo/tests/run.sh
+++ b/hooks/repo/tests/run.sh
@@ -16,7 +16,7 @@
 # diverge from the breakdown (repo#44). Currently delegated:
 # test-guard-destructive.sh (the full guard regression suite),
 # test-session-start-handoff.sh, test-install-claude-md-markers.sh,
-# commands/repo/tests/test-branches-loss-check.sh, and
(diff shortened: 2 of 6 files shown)


rjwalters:
## ✅ Approved — ready for merge

Reviewed the **integrated** result: this branch was 9 commits behind a very active `main`, so I merged `origin/main` into it (clean, zero conflicts) and pushed — the PR head is now `567807a` and everything below was verified against that merged tree, not the stale base.

### Verification
- **Full suite green on the integrated tree**: `pnpm test` → `run.sh` reports **864/0** (test-shell-wrapper **55/55**, session-start-handoff **99/99**, guard-destructive 488, CLAUDE.md markers 59, branches-loss 37, repo-remote 71). The 8 flaky `test-guard-destructive` cases the PR body flagged are **gone** — main's #76 (`ml_segment` lexer) resolved them, so the merge is strictly better than the pre-rebase state.
- **CI passes**: `pnpm test (guard/handoff/loss-check suite)` → `success` on head `567807a`.
- **shellcheck -x** clean on `lib/shell-wrapper.sh`; the only notices are pre-existing `SC2094` (info) in `install.sh` unrelated to this PR and `SC1090` (warning) for the test's intentional dynamic `source` of the rendered block — both consistent with the codebase's accepted notices.

### Correctness & scope
- **Fail-transparent by construction** holds up: `claude()` runs `command claude <flags> "$@"` unconditionally, the banner is gated behind a `type` check on both helpers, and all helpers live in the same marker-bounded block — so the `claude mcp list` failure mode reported on #35 (a helper-less wrapper mangling argv) is structurally impossible here, and there's an explicit runtime test for it.
- **The one out-of-repo write is handled with the right discipline**: strict no-op under `--yes` without `--shell-wrapper` (rc neither read nor written), always-diff-then-confirm (default **N**) otherwise, backup-before-every-edit, marker-bounded + idempotent (in-place replace), and a hard bail on ambiguous marker layout. Uninstall symmetrically removes the block from both candidate rc files with the same backup discipline and is a no-op when absent.
- **Alias preservation** parses only the common single-line `alias claude='[command] claude <flags>'` form and refuses anything else rather than guessing — a sound scope guard. `unalias claude` precedes the function so the two never fight.
- **Consistency with what merged today**: mirrors `lib/claude-md-block.sh`'s marker/backup discipline; the banner's mtime logic (`stat -f %m` / `stat -c %Y`) is self-contained, so main's session-start-handoff evolution doesn't affect it; run.sh registration follows the established delegated-suite pattern.
- All acceptance criteria from #35 verified against the criteria table; `Closes #35` present.

Excellent, defensive work on the riskiest piece of the installer. Nothing blocking.

PR #67, opened 2026-07-29
feat(sudo): add /repo:sudo passwordless-sudo setup command

### LINKED ISSUE #49: feat: opt-in passwordless-sudo setup for dev machines (sudoers.d drop-in)
## Motivation

During a live troubleshooting session (VNC unusable on a heavily loaded Mac Studio), an agent working over SSH was repeatedly blocked on root-level operations — unloading a runaway backup daemon (`launchctl bootout system/...`), killing root processes, and investigating `syspolicyd` — because `sudo` prompts for a password that a non-interactive SSH session cannot answer. Each blocked step had to be handed back to the human as a copy-paste command, which defeats the point of agent-driven remote sessions.

The manual fix was a standard sudoers drop-in:

```bash
echo "$USER ALL=(ALL) NOPASSWD: ALL" | sudo tee /etc/sudoers.d/$USER-nopasswd
sudo chmod 440 /etc/sudoers.d/$USER-nopasswd
sudo visudo -c   # validate so a typo can't brick sudo
```

This is machine setup we now want on every dev box an agent works on, so it belongs in the skill pack rather than in shell history.

## Proposal

Add an opt-in machine-setup step (either a small `/repo:sudo` skill, or a flag on the existing provisioning path — `/repo:remote` already provisions cloud dev machines that have the same problem) that:

1. Writes `/etc/sudoers.d/<user>-nopasswd` with `NOPASSWD:` for the current user.
2. Sets `440` permissions and always runs `visudo -c` before finishing; on validation failure it removes the file (never leave sudo broken).
3. Is idempotent — re-running detects the existing drop-in and no-ops.
4. Works on macOS (`/etc/sudoers` includes `/private/etc/sudoers.d` by default) and Linux (`#includedir /etc/sudoers.d`).

## Security considerations

- This must stay **explicitly opt-in and always confirmed** — it grants anyone holding the user's SSH key effective root. That's a reasonable trade for personal dev machines on a private tailnet, and exactly the class of consequential action the pack already gates (`orphans`, `remote`, etc.).
- Offer a **scoped variant**: `NOPASSWD:` limited to a command list (e.g. `/bin/launchctl`, `/usr/bin/pkill`, `/usr/sbin/spctl`, `/usr/bin/renice`) instead of `ALL`, for people who want agents unblocked on service management without blanket root.
- The destructive-command guard should still apply on top — passwordless sudo removes the password gate, not the guard's `ask`/`block` gates on risky commands.
- Note in docs: Touch ID (`pam_tid.so`) is the better answer for local interactive use but does not work over SSH, which is why the drop-in is the right tool for agent sessions.

## Context

Came out of a real session: Backblaze `bztransmit -completesync` + 6 concurrent release builds had a Mac Studio at load 118; every remediation step needing root stalled on the sudo prompt until the human ran the drop-in setup interactively. After the drop-in, the agent handled verification and cleanup end-to-end.

---

### Problem Statement (confirmed)

Verified against the current tree: there is no existing `/repo:sudo` command,
no sudoers-related code anywhere in `commands/`, `skills/`, or `hooks/`
(`grep -rn "sudoers\|visudo\|NOPASSWD"` outside this issue's own text returns
nothing), and no prior art to build on — this is a net-new skill. The proposal
in the original body above is sound and directly actionable.

### Implementation Approach

Ship this as a **dedicated `/repo:sudo` skill** (`commands/repo/sudo.md`),
not a flag bolted onto `/repo:remote`:

- It needs to run against *any* dev machine an agent works on (the local box
  over SSH, not just cloud instances `/repo:remote` provisions), so it can't
  live only behind that command.
- Issue #50 (`/repo:host-optimize`, already `loom:curated`) explicitly plans
  to "detect and delegate to #49's mechanism rather than reimplement it" for
  its sudo-posture check. A delegable target needs to be a standalone,
  independently invocable command — a flag buried inside `/repo:remote`
  wouldn't be.
- `/repo:remote` MAY optionally mention `/repo:sudo` in its own docs as a
  companion step for a freshly provisioned box (see Affected Files below),
  but should not absorb this scope itself.

Recommended command surface (final naming/flags are an implementation
decision, not a hard spec):

```
/repo:sudo                 # Guided: explain the grant, confirm, install ALL-scoped drop-in
/repo:sudo --scoped        # Confirm, then install a command-list-scoped drop-in
                            #   (launchctl, pkill, spctl, renice, ...) instead of ALL
/repo:sudo --status        # Report whether a drop-in exists, its scope, and validity — no writes
/repo:sudo --remove        # Confirm, then remove this pack's drop-in file
```

Follow the pack's consequential-command convention (same class as `/repo:remote`,
`/repo:followups`, `/repo:update-tools`): **always confirm before writing**,
show the exact file contents that will be written, and never apply silently
even without `--ask` (there is nothing to opt into — confirmation is the only
mode, same rationale `followups.md` states for itself).

Core write sequence (mirrors the original body's proposal exactly):

```bash
DROPIN="/etc/sudoers.d/${USER}-nopasswd"   # or a scoped variant's file
TMP="$(mktemp)"
echo "$USER ALL=(ALL) NOPASSWD: ALL" > "$TMP"     # or the scoped command list
chmod 440 "$TMP"
sudo cp "$TMP" "$DROPIN"
sudo chmod 440 "$DROPIN"
sudo visudo -c || { sudo rm -f "$DROPIN"; echo "validation failed, removed"; exit 1; }
```

Discoverability for a future manifest (per #35's curation, which already
generalized "global things this pack installed" as a shared design
constraint across #35/#49/#50): put a one-line comment at the top of the
drop-in identifying it as installed by this skill, e.g.
`# Installed by /repo:sudo (rjwalters/repo) — see .claude/commands/repo/sudo.md`,
so `--status`/`--remove` (and any later host-optimize-style audit) can find
and attribute it without guessing from the filename alone.

Idempotency check: before writing, read the existing drop-in (if present),
compare its scope (ALL vs. the same command list) against what was requested,
and no-op with a status line if they already match; only prompt to overwrite
if the requested scope differs from what's installed.

macOS vs. Linux: both use the `/etc/sudoers.d/` drop-in path already covered
by the include directive on each OS per the original body — no OS branching
needed beyond confirming the include is active (macOS: default; Linux: most
distros ship `#includedir /etc/sudoers.d` in `/etc/sudoers` already, but
verify with `sudo visudo -c` after writing regardless of platform, which
implicitly proves the include chain works).

### Related Open Work (RELATED_OPEN_WORK probe)

`check-duplicate.sh --issue 49` surfaced two open issues cross-referencing
this one; both already curated with their own reconciliation, and neither
requires a spec change here:

- **#50 (`/repo:host-optimize`) — absorbed.** #50's sudo-posture check is
  written to detect and delegate to this issue's mechanism, not reimplement
  it, and explicitly treats this as a non-blocking cross-reference ("#49 does
  not need to land first — print the manual recipe until #49 ships, then
  switch to delegating"). This curation keeps that contract intact by giving
  #50 a standalone `/repo:sudo` command (not a `/repo:remote` flag) to
  delegate to, and a `--status` mode it can call without writing anything.
- **#35 (shell `claude` wrapper) — absorbed.** #35's own curation already
  generalized the "marker-bounded, backed up, enumerable" discipline as a
  shared design constraint across #35/#49/#50, without creating an
  implementation dependency in either direction. Absorbed above via the
  "Installed by /repo:sudo" comment-header requirement, giving this drop-in
  the same discoverability #35's rc-edit markers provide for its own change.

No scope changes resulted from either — both were already reconciled by
their own curation passes; this section documents that #49's spec is
consistent with what #35 and #50 assumed about it.

### Acceptance Criteria

- [ ] New command `/repo:sudo` exists at `commands/repo/sudo.md` with the
      pack's standard frontmatter (`name`, `description`, `domain: repo`,
      `type: command`, `user-invocable: true`) and Usage/Steps structure.
- [ ] Default invocation writes `/etc/sudoers.d/<user>-nopasswd` with
      `NOPASSWD: ALL` for the current user, only after an explicit
      confirmation that shows the exact file contents to be written.
- [ ] `--scoped` writes a `NOPASSWD:` grant limited to a command allowlist
      (e.g. `/bin/launchctl`, `/usr/bin/pkill`, `/usr/sbin/spctl`,
      `/usr/bin/renice`) instead of `ALL`, confirmed the same way.
- [ ] File permissions are always `440`; `sudo visudo -c` always runs before
      the command reports success; on validation failure the drop-in is
      removed and the command exits non-zero — verified by implementation
      review that this path can never leave a broken/unvalidated file behind.
- [ ] Re-running with the same scope is idempotent: detects the existing
      drop-in, reports it's already configured, and does not re-prompt or
      re-write.
- [ ] Re-running with a *different* scope than what's installed (e.g.
      `--scoped` after a prior `ALL` install) prompts to confirm the change
      before overwriting.
- [ ] `--status` reports whether a drop-in exists, its scope (ALL vs. scoped
      list), and whether it currently passes `visudo -c`, without writing
      anything.
- [ ] `--remove` deletes this pack's drop-in file, only after confirmation.
- [ ] The written drop-in carries a comment header identifying it as
      installed by `/repo:sudo`, for discoverability by other tooling.
- [ ] Works on both macOS and Linux (both platforms use the same
      `/etc/sudoers.d/` drop-in mechanism; no OS-specific branching needed
      beyond the `visudo -c` verification already required above).
- [ ] `skills/repo/SKILL.md`, `commands/repo/help.md`, and root `README.md`
      each gain a `/repo:sudo` entry consistent with existing rows.

### Affected Files

- `commands/repo/sudo.md` — new command implementation (does not yet exist)
- `skills/repo/SKILL.md` — add `[[sudo]]` row to the Commands table
- `commands/repo/help.md` — add `/repo:sudo` to the appropriate bucket in the
  overview (likely "Everyday" or a new opt-in-setup grouping — it's a one-time
  machine-setup step rather than a recurring hygiene pass)
- `README.md` — add `/repo:sudo` row to the root Skills table
- `commands/repo/remote.md` — optional cross-reference only (e.g. a mention
  that a freshly provisioned box can run `/repo:sudo` to unblock root-level
  remediation over SSH); no required behavior change to `remote.md` itself

### Test Plan

- [ ] Manual verification on macOS: run `/repo:sudo`, confirm the prompt shows
      the exact file contents, confirm, verify `/etc/sudoers.d/<user>-nopasswd`
      exists at `440` and `sudo visudo -c` passes; verify a subsequent
      passwordless `sudo -n true` succeeds.
- [ ] Manual verification on Linux: same as above, confirming the
      `#includedir /etc/sudoers.d` (or equivalent) chain is active.
- [ ] `--scoped`: verify the resulting grant only covers the allowlisted
      commands (`sudo -n launchctl ...` succeeds; an out-of-list command still
      prompts for a password).
- [ ] Idempotency: run `/repo:sudo` twice back-to-back with the same scope;
      the second run reports "already configured" and makes no writes.
- [ ] Scope-change: install default (`ALL`), then run `--scoped`; confirm it
      prompts before overwriting rather than silently changing scope.
- [ ] Validation-failure path: implementation review (or a forced-failure
      test double for `visudo`) confirms a failed `visudo -c` always results
      in the drop-in being removed, never left in place.
- [ ] `--status`: confirm it accurately reports presence/scope/validity and
      performs zero writes (verify via `sudo -n cat` unchanged / file mtime
      unchanged before and after).
- [ ] `--remove`: confirm it deletes the drop-in only after confirmation, and
      a subsequent `sudo` invocation prompts for a password again.
- [ ] Doc check: run `/repo:help` after the change and confirm `/repo:sudo`
      appears in the generated overview.

### Dependencies (non-blocking cross-references)

- #50 (`/repo:host-optimize`) — its sudo-posture check delegates to this
  issue's mechanism rather than reimplementing it. Not a hard dependency in
  either direction: #50 already ships printing the manual recipe until this
  issue lands, then switches to delegating once `/repo:sudo` exists.
- #35 (shell `claude` wrapper) — already curated and `loom:issue`; shares the
  "machine-global, opt-in, gated, discoverable" design constraint this issue's
  comment-header requirement satisfies. No implementation dependency.

rjwalters:
## Summary

Adds a dedicated **`/repo:sudo`** command that installs an opt-in,
`visudo`-validated `/etc/sudoers.d/<user>-nopasswd` drop-in so an agent driving a
dev machine over SSH isn't blocked on `sudo` password prompts a non-interactive
session can't answer. Machine-global, consequential setup in the always-confirm
class (like `/repo:remote`, `/repo:followups`, `/repo:update-tools`): it shows
the exact file contents before writing and never applies silently.

## Changes

- **`commands/repo/sudo.md`** (new) — the command. Surface:
  - `/repo:sudo` — guided, installs a blanket `NOPASSWD: ALL` drop-in
  - `/repo:sudo --scoped` — installs a `NOPASSWD:` grant limited to a command
    allowlist (`/bin/launchctl`, `/usr/bin/pkill`, `/usr/sbin/spctl`,
    `/usr/bin/renice`)
  - `/repo:sudo --status` — reports presence, scope, and `visudo -c` validity;
    writes nothing
  - `/repo:sudo --remove` — deletes this command's own drop-in after confirmation

  Write sequence builds to a temp file, copies into place at `0440`, then always
  runs `sudo visudo -c` with **delete-on-failure and non-zero exit** so a broken
  sudoers file is never left behind. Idempotent per-scope (same scope no-ops; a
  scope *change* prompts to overwrite). The drop-in carries an
  `# Installed by /repo:sudo (rjwalters/repo)` discoverability header. No OS
  branching — both macOS and Linux read `/etc/sudoers.d/`, and `visudo -c` proves
  the include chain on either. Documents that Touch ID is the better local answer
  but doesn't work over SSH, and that the destructive-command guard still applies.
- **`skills/repo/SKILL.md`** — `[[sudo]]` row (machine-setup cluster, after
  `[[remote]]`), a When-to-Use bullet, and `sudo` added to the consequential list.
- **`commands/repo/help.md`** — `/repo:sudo` in the Everyday bucket.
- **`README.md`** — `/repo:sudo` row and consequential list updated.
- **`commands/repo/remote.md`** — optional companion cross-reference to
  `/repo:sudo` for a freshly provisioned box.

- [x] New `commands/repo/sudo.md` with standard frontmatter + Usage/Steps
- [x] Default writes `/etc/sudoers.d/<user>-nopasswd` `NOPASSWD: ALL`, confirmed with exact contents shown
- [x] `--scoped` writes the command-allowlist grant, confirmed the same way
- [x] Perms always `0440`; `sudo visudo -c` always runs before success; delete-on-failure with non-zero exit (no path leaves a broken/unvalidated file)
- [x] Re-run same scope is idempotent (no re-prompt, no re-write)
- [x] Re-run different scope prompts before overwriting
- [x] `--status` reports presence/scope/validity, writes nothing
- [x] `--remove` deletes only this command's drop-in, after confirmation
- [x] Drop-in carries the `Installed by /repo:sudo` header
- [x] Works on macOS and Linux (shared `/etc/sudoers.d/` mechanism, no OS branch)
- [x] `SKILL.md`, `help.md`, `README.md` each gain a `/repo:sudo` entry

## Test Plan

- [x] `pnpm test` — **782 pass, 0 fail** (unchanged; the suites cover the guard
  hooks and scripts, and the new command is prose with no harness of its own).
- Manual sudoers verification is inherently machine-mutating and left to the
  operator per the issue's Test Plan; the write/validate/rollback ordering is
  verifiable by implementation review of Step 5.

Closes #49

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,7 @@ Repo is a collection of skills for keeping any git repository healthy and produc
 | `/repo:tidy` | Tidy up — build artifacts, caches, temp files, empty dirs |
 | `/repo:release` | Cut a release — pre-flight checks, semver decision, CHANGELOG, version bump, tag, GitHub Release |
 | `/repo:remote` | Launch a cloud dev session (GCP or AWS) with the repo ready to go, then open an SSH session |
+| `/repo:sudo` | Opt-in passwordless-sudo setup for a dev machine — install a `visudo`-validated `/etc/sudoers.d` drop-in (blanket `ALL` or a scoped command list) so an agent over SSH isn't blocked on password prompts; always confirmed, validated with rollback on failure |
 | `/repo:host-optimize` | Prepare a Mac (or Linux box) for heavy Loom/agent build use — audit Gatekeeper churn, backup-agent interference, build-tree bloat; apply safe fixes, gate consequential ones |
 | `/repo:update-tools` | Check installed tool packages (Loom, Anvil, …) against their sources and offer updates |
 | `/repo:followups` | Capture follow-on work from this session and file it as issues — here or in upstream tool repos, always confirmed first |
@@ -28,7 +29,7 @@ Repo is a collection of skills for keeping any git repository healthy and produc
 | `/repo:orphans` | Find unreferenced files — dead scripts, stale data, outputs without sources |
 | `/repo:readme` | Check README accuracy against actual directory contents |
 
-Hygiene skills **apply their safe, reversible fixes by default** and report each change; add `--ask` to review findings and confirm first. Irreversible actions (deleting branches, worktrees, stashes, untracked files) are never automatic — they require an explicit opt-in and pass a permanent-loss check. Commands whose only action is consequential (`orphans`, `update-tools`, `followups`, `release`, `remote`) always confirm first.
+Hygiene skills **apply their safe, reversible fixes by default** and report each change; add `--ask` to review findings and confirm first. Irreversible actions (deleting branches, worktrees, stashes, untracked files) are never automatic — they require an explicit opt-in and pass a permanent-loss check. Commands whose only action is consequential (`orphans`, `update-tools`, `followups`, `release`, `remote`, `sudo`) always confirm first.
 
 ## Destructive-command protection
 
--- a/commands/repo/help.md
+++ b/commands/repo/help.md
@@ -60,6 +60,7 @@ stashes, untracked files) always need an explicit opt-in.
 | /repo:handoff | Rolling the session (context full, CLI update) — preserve what only the session knows, then restart |
 | /repo:tidy  | Working tree cluttered with build artifacts and temp files |
 | /repo:remote | Need a cloud dev box (GCP/AWS) with this repo ready to go |
+| /repo:sudo | One-time machine setup — grant passwordless sudo (validated drop-in) so an agent over SSH isn't blocked on password prompts |
 | /repo:release | Cut a release — pre-flight, semver, CHANGELOG, version bump, tag, GitHub Release |
 
 ### Periodic maintenance
--- a/commands/repo/remote.md
+++ b/commands/repo/remote.md
@@ -468,6 +468,10 @@ Claude Code cannot host an interactive SSH session itself, so hand it to the OS
   terminal
 
 Tell the user they can start `claude` in that session to continue on the remote.
+On a freshly provisioned box, `/repo:sudo` is the companion step that unblocks
+root-level remediation over SSH — a non-interactive SSH session can't answer a
+`sudo` password prompt, so an agent stalls on it until a validated
+passwordless-sudo drop-in is installed.
 
 ### 8. Report
 
--- a/commands/repo/sudo.md
+++ b/commands/repo/sudo.md
@@ -64,9 +64,12 @@ needed**:
 - **Linux** — nearly every distro ships `#includedir /etc/sudoers.d` in
   `/etc/sudoers`.
 
-Running `sudo visudo -c` after the write implicitly proves the include chain is
-active on either platform (it validates `/etc/sudoers` **and** every file it
-includes), so there is no need to detect the OS or hand-verify the directive.
+The candidate file is always syntax-checked **in isolation** with
+`sudo visudo -cf "$TMP"` *before* it is copied into `/etc/sudoers.d/` — a
+malformed drop-in never becomes live policy. After the validated file is
+installed, a full-chain `sudo visudo -c` (it validates `/etc/sudoers` **and**
+every file it includes) implicitly proves the include chain is active on either
+platform, so there is no need to detect the OS or hand-verify the directive.
 
 The written file always carries a **discoverability header** on its first line
 so `--status`, `--remove`, and any later host-posture audit (e.g.
@@ -174,9 +177,15 @@ Always show the **exact file contents** (including the header line) that will be
 written, name the target path, and get an explicit **yes** before touching
 anything — even without `--ask`. Never write silently.
 
-Build the file in a temp location, copy it into place at `0440`, then validate
-— and **remove it on any validation failure** so a typo can never leave `sudo`
-broken (a malformed sudoers file can lock the user out of root):
+Build the file in a temp location and **validate the candidate in isolation
+before it can go live** — only a syntactically valid file is ever copied into
+`/etc/sudoers.d/`. There is no separate "enable" step for a sudoers.d file: the
+`@includedir` / `#includedir` directive picks it up on the very next `sudo`
+parse, so a broken file that reaches the directory is *already* live policy —
+and because sudo fails closed on a syntax error, the rollback `sudo rm` could be
+denied by the very file it is trying to remove. Validating `$TMP` first makes
+that failure mode impossible; the post-install full-chain check is then only a
+belt-and-suspenders proof that the include wiring is sound:
 
 ```bash
 USER_NAME="$(id -un)"
@@ -193,23 +202,38 @@ TMP="$(mktemp)"
 } > "$TMP"
 chmod 440 "$TMP"
 
+# Validate the candidate BEFORE it can affect live policy. visudo -cf checks a
+# single file's syntax without installing it, so a malformed drop-in never
+# becomes active and we never depend on a possibly-broken sudo to undo it.
+if ! sudo visudo -cf "$TMP"; then
+  rm -f "$TMP"
+  echo "candidate sudoers failed validation — nothing installed" >&2
+  exit 1
+fi
+
+# Only a validated file reaches /etc/sudoers.d/:
 sudo cp "$TMP" "$DROPIN"
 sudo chmod 440 "$DROPIN"
 rm -f "$TMP"
 
-# ALWAYS validate before reporting success. On failure, remove the drop-in and
-# exit non-zero — never leave an unvalidated or broken file in place.
+# Belt-and-suspenders: full-chain re-check proves the include wiring too. If it
+# somehow fails, remove the drop-in and exit non-zero — never leave an
+# unvalidated or broken file in place. (This rollback is now reliable because
+# the file was already known-valid before it landed.)
 if ! sudo visudo -c; then
   sudo rm -f "$DROPIN"
-  echo "sudoers validation failed — removed ${DROPIN}, no change made" >&2
+  echo "post-install validation failed — removed ${DROPIN}, no change made" >&2
   exit 1
 fi
 ```
 
-The order is load-bearing: the file is written at `0440`, and `visudo -c` is
-the last gate before success. If it fails for **any** reason, the drop-in is
-deleted and the command exits non-zero — there is no path in which a
-failed-validation file survives.
+The order is load-bearing: the candidate is syntax-checked with
+`visudo -cf "$TMP"` **before** the `cp`, so a malformed drop-in never becomes
+live policy and the rollback never has to fight a broken sudo. The post-install
+`visudo -c` is the final full-chain gate; if it fails for **any** reason the
+drop-in is deleted and the command exits non-zero — there is no path in which a
+failed-validation file survives, and no path in which an unvalidated file goes
+live even momentarily.
 
 ### 6. Verify and report
 
@@ -230,11 +254,16 @@ it later.
 1. **Always confirm before writing** — show the exact file contents (header +
    grant line) and the target path, and act only on an explicit yes. There is
    no silent/auto-apply mode; confirmation is the only mode.
-2. **Always validate, always roll back on failure** — the drop-in is written at
-   `0440` and `sudo visudo -c` runs before the command reports success. On any
-   validation failure the file is removed and the command exits non-zero. A
-   broken sudoers file can lock the user out of root, so an unvalidated file is
-   never left behind.
+2. **Validate the candidate before it goes live, then roll back on failure** —
+   the candidate is built at `0440` in a temp file and syntax-checked in
+   isolation with `sudo visudo -cf "$TMP"` **before** it is copied into
+   `/etc/sudoers.d/`, so a malformed drop-in never becomes active policy (a
+   sudoers.d file is live the instant it lands, and sudo fails closed on a
+   syntax error — validating pre-install is what keeps the rollback reliable). A
+   post-install full-chain `sudo visudo -c` then re-checks the include wiring;
+   on any failure the file is removed and the command exits non-zero. A broken
+   sudoers file can lock the user out of root, so an unvalidated file is never
+   installed or left behind.
 3. **Idempotent** — re-running with the same scope detects the existing drop-in
    and no-ops; only a scope *change* prompts to overwrite.
 4. **Only ever touch this command's own drop-in** at
--- a/commands/repo/sudo.md
+++ b/commands/repo/sudo.md
@@ -1,0 +1,246 @@
+---
+name: "sudo"
+description: "Opt-in passwordless sudo for a dev machine — install a visudo-validated /etc/sudoers.d drop-in (ALL or a scoped command list) so an agent over SSH isn't blocked on password prompts, always confirmed first"
+domain: repo
+type: command
+user-invocable: true
+---
+
+# /repo:sudo — Passwordless-sudo setup for dev machines
+
+Install (or inspect, or remove) a `NOPASSWD:` **sudoers drop-in** for the
+current user on **this machine**, so an agent driving the box over SSH isn't
+stopped dead by a `sudo` password prompt it can't answer. A non-interactive SSH
+session has no TTY to type a password into, so every root-level remediation —
+`launchctl bootout`, killing a runaway root process, `spctl`, `renice` — stalls
+until a human runs the command by hand. This command installs the standard fix
+once, safely, so agent-driven remote sessions can finish end-to-end.
+
+This is **machine-global, consequential setup**: the drop-in grants anyone
+holding the user's SSH key effective root on this box. It is therefore in the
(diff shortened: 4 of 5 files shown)


rjwalters:
## Changes Requested — one blocking safety defect in the write/validate/rollback ordering

This is a well-scoped, well-documented command and almost all of it is right. But the core Step 5 write sequence — the single most safety-critical part of a command that mutates `/etc/sudoers.d/` — has an ordering flaw that defeats its own rollback guarantee. Everything else checks out; fix this one thing and I'll approve.

### Blocker: `visudo -c` validates *after* the drop-in is already live, and the rollback can't run if the file is broken

Step 5 (and the "How it works" prose at line 67, and Safety Rule 2) validate with **full-chain `sudo visudo -c` only after** `sudo cp \"\$TMP\" \"\$DROPIN\"` has already placed the file into `/etc/sudoers.d/`:

```bash
sudo cp \"\$TMP\" \"\$DROPIN\"     # drop-in is LIVE the instant it lands here
sudo chmod 440 \"\$DROPIN\"
rm -f \"\$TMP\"
if ! sudo visudo -c; then       # validation runs only now — too late
  sudo rm -f \"\$DROPIN\"         # rollback needs sudo, which the bad file may have just broken
  exit 1
fi
```

There is no separate \"enable\" step for a sudoers.d file: the `@includedir` / `#includedir` directive picks it up on the **next `sudo` parse**, so between the `cp` and the `visudo -c` the (possibly malformed) file is already part of the active policy. And a syntax error anywhere in the sudoers chain makes sudo **fail closed** — meaning the rollback command `sudo rm -f \"\$DROPIN\"` can itself be denied by the very file it is trying to remove. The rollback is unreliable in exactly the failure case it exists for. That is the \"a typo can lock you out of root\" scenario the command promises to prevent, reintroduced by the ordering. The comment at line 209 (\"`visudo -c` is the last gate before success\") makes this the deliberate design rather than a slip.

**Fix — validate the candidate in isolation *before* it goes live**, then install, then re-check the full chain:

```bash
chmod 440 \"\$TMP\"
# Validate the candidate BEFORE it can affect the live policy:
if ! sudo visudo -cf \"\$TMP\"; then
  rm -f \"\$TMP\"
  echo \"candidate sudoers failed validation — nothing installed\" >&2
  exit 1
fi
# Only a syntactically valid file ever reaches /etc/sudoers.d/:
sudo cp \"\$TMP\" \"\$DROPIN\"
sudo chmod 440 \"\$DROPIN\"
rm -f \"\$TMP\"
# Belt-and-suspenders: full-chain re-check (proves the include wiring too):
if ! sudo visudo -c; then
  sudo rm -f \"\$DROPIN\"
  echo \"post-install validation failed — removed \${DROPIN}\" >&2
  exit 1
fi
```

`visudo -cf \"\$TMP\"` syntax-checks a single file without installing it, so a malformed drop-in never becomes active and you never depend on a possibly-broken sudo to undo it. Please also update the \"How it works\" paragraph (line ~67) and Safety Rule 2 so the prose describes validate-then-install (the acceptance criterion \"never leaves a broken/unvalidated file\" is only actually met once the candidate is validated pre-install).

### Verified and good (no change needed)

- **Consent gating** — opt-in, always-confirm-first, exact file contents shown before any write, no silent/auto mode. Correct.
- **`--status` is genuinely read-only** — `sudo cat` / `sudo visudo -c` / `sudo -n true` only, explicit \"must not change mtime,\" exits before any write path. Good.
- **`--remove` is safe** — only touches this command's own header-identified `\$DROPIN`, confirms first, re-validates after removal. Good.
- **`--scoped` allowlist** — `NOPASSWD: /bin/launchctl, /usr/bin/pkill, /usr/sbin/spctl, /usr/bin/renice` is valid comma-separated Cmnd_Spec syntax and genuinely narrower than `NOPASSWD: ALL`. Good.
- **Registry consistency** — `README.md`, `commands/repo/help.md`, `skills/repo/SKILL.md` (table row + when-to-use bullet + consequential list), and the `remote.md` cross-ref are all consistent; no dangling links.
- **CI** green, **mergeStateStatus** CLEAN.

Once Step 5 validates the temp file before copying it into place (and the prose matches), this is good to merge.

rjwalters:
## Feedback addressed — Step 5 now validates the candidate before it goes live

Fixed the blocking write/validate/rollback ordering defect. The command no longer copies the drop-in into `/etc/sudoers.d/` before validating it.

**What changed in `commands/repo/sudo.md`:**

- **Step 5 sequence** — the candidate is now syntax-checked in isolation with `sudo visudo -cf "$TMP"` **before** the `sudo cp "$TMP" "$DROPIN"`. Only a validated file ever reaches `/etc/sudoers.d/`, so a malformed drop-in never becomes live policy (a sudoers.d file is active the instant `@includedir`/`#includedir` re-parses it), and the rollback `sudo rm` can never be denied by the very broken file it is removing. The post-install full-chain `sudo visudo -c` is retained as a belt-and-suspenders re-check that proves the include wiring, with reliable rollback since the file was already known-valid.
- **Step 5 rationale** — the old "the file is written at 0440 and `visudo -c` is the last gate before success" prose (which documented copy-then-validate as deliberate) is replaced with validate-then-install: the `visudo -cf "$TMP"` pre-check before the `cp` is now the load-bearing step, and no unvalidated file goes live even momentarily.
- **"How it works" (line ~67)** — updated to describe isolation validation before install, then the full-chain check after.
- **Safety Rule 2** — retitled to "Validate the candidate before it goes live, then roll back on failure" and rewritten to match.

Everything the Judge verified as sound is unchanged: consent gating, `--status` read-only, `--remove` self-scoping, and the `--scoped` allowlist.

`pnpm test` from the worktree: **782 pass, 0 fail**.

rjwalters:
## Approved — the write/validate/rollback ordering defect is fully fixed

Re-reviewed after the Doctor pass (commit `ded76e8`). The single blocking defect from the prior review is resolved, and nothing previously-verified regressed.

### The blocker is fixed — validate-isolated before it goes live

Step 5 now syntax-checks the candidate in isolation with `sudo visudo -cf "$TMP"` **before** the `sudo cp "$TMP" "$DROPIN"`, so a malformed drop-in never reaches `/etc/sudoers.d/` and never becomes live policy. Because only a known-valid file is ever installed, the post-install full-chain `sudo visudo -c` is now correctly a belt-and-suspenders re-check of the include wiring, and its rollback (`sudo rm -f "$DROPIN"`) can no longer be denied by the very broken file it exists to remove. The self-defeating rollback scenario is gone.

### Prose is coherent — no stale copy-then-validate text remains

All three sites the prior review flagged now describe validate-then-install, and they agree with the code:
- **How it works (~line 67)** — "syntax-checked in isolation with `sudo visudo -cf "$TMP"` *before* it is copied."
- **Step 5 sequence + rationale** — the old "`visudo -c` is the last gate before success" (which documented copy-then-validate as deliberate) is replaced; the `visudo -cf` pre-check is now the load-bearing step, post-install `visudo -c` is the final full-chain gate.
- **Safety Rule 2** — retitled "Validate the candidate before it goes live, then roll back on failure" and rewritten to match.

No lingering copy-then-validate ordering anywhere in `sudo.md`.

### Spot-check of previously-verified surfaces — unchanged and still sound

- **Consent gate** — opt-in, always-confirm-first, exact file contents shown before any write, no silent mode.
- **`--status`** — read-only (`sudo cat` / `visudo -c` / `sudo -n true`), exits before any write path.
- **`--remove`** — only touches this command's own header-identified `$DROPIN`, confirms first, re-validates after.
- **`--scoped`** — `NOPASSWD: /bin/launchctl, /usr/bin/pkill, /usr/sbin/spctl, /usr/bin/renice` remains valid, narrower-than-ALL Cmnd_Spec syntax.
- Registry consistency (README, help, SKILL, remote cross-ref) unchanged and consistent.

### Gates
- `pnpm test` in worktree `.loom/worktrees/issue-49`: **782 pass, 0 fail**.
- CI: **green**. mergeStateStatus: **CLEAN**.

Good to merge.

PR #66, opened 2026-07-29
fix(gitignore): match node_modules symlink by dropping trailing slash

rjwalters:
## Summary

Changes the `.gitignore` dependency pattern from `node_modules/` (directory-only) to `node_modules` (no trailing slash) so it matches both the real `node_modules` directory in the main checkout and the `node_modules` symlink that `./.loom/scripts/worktree.sh` creates in every issue worktree. Previously each worktree showed `node_modules` as untracked, adding noise that every builder had to reason about (flagged independently by the issue-48 and issue-53 builders during the 2026-07-29 sweep) and risking an accidental `git add -A` committing the symlink.

## Changes

- `.gitignore`: `node_modules/` -> `node_modules` (line 4). No other lines touched; the pattern stays outside the `>>> loom-managed >>>` block, so it survives Loom updates. Per the scope guard, `.loom/scripts/worktree.sh` (upstream-owned) is intentionally NOT modified.

## Acceptance Criteria Verification

- [x] `.gitignore` pattern changed from `node_modules/` to `node_modules` (single-line diff; nothing else changed; stays outside the loom-managed block)
- [x] `git check-ignore -q node_modules` passes in the main checkout (real directory still ignored)
- [x] `git check-ignore -q node_modules` passes against a symlink (verified in the issue worktree, whose `node_modules` is a symlink, and independently in a scratch repo under `$TMPDIR`)
- [x] In this worktree (created by `worktree.sh`), `git status --porcelain` no longer lists `node_modules` (only ` M .gitignore` appears)

## Test Plan

- Worktree symlink context: `git check-ignore -v node_modules` -> `.gitignore:4:node_modules` (match); `git status --porcelain` no longer lists `node_modules`.
- Scratch repo (`$TMPDIR`): with the new pattern, both a `node_modules` symlink and a real `node_modules` directory are ignored; `git status --porcelain` stays empty for both.
- Full suite: `CI=true pnpm test` -> PASS: 782, FAIL: 0.

Closes #59

PR #69, opened 2026-07-29
fix(guard-destructive): make extract_rm_targets quote-aware across newlines

### LINKED ISSUE #60: guard-destructive: multi-line quoted literal with line-leading recursive-force delete still false-blocks (extract_rm_targets per-line scan)
## Problem

PR #58 (issue #53) fixed false-blocks on dangerous strings quoted as `echo`/`printf` data, but deliberately left one pre-existing limitation out of scope: a **multi-line quoted literal whose interior line *begins* with a recursive-force root delete** (e.g. documentation prose or a heredoc where a line starts with `rm -rf /` targeting root) still denies, because `extract_rm_targets()` runs a per-line raw scan that is not quote-aware across lines.

This limitation is shared identically with the `--message`/`--body` flag-redaction path — the #3898 tests encode the same prose-before-danger convention as a workaround, not a fix.

## Scope / caution

This is security-guard logic (`complexity=complex` territory). Any fix must preserve the safety floor verified in PR #58's review: `$(...)`/backtick smuggling, `bash -c`/`sh -c` payloads, pipe-to-shell (piping echoed data into `sh`), and dangerous commands in separate pipeline segments must all still deny. A fix that makes `extract_rm_targets()` quote-span-aware across lines must not create a bypass where a *real* multi-line command hides a delete behind an opening quote on a prior line.

## Affected Files

- `hooks/repo/guard-destructive.sh` (`extract_rm_targets()` and its call sites)
- `hooks/repo/tests/test-guard-destructive.sh`

## Test Plan

- New regression cases: multi-line quoted literal with a line-leading recursive-force root delete (echo data, printf data, `--body` flag data) → allow
- Full existing safety-floor cases → still deny
- Full suite via `hooks/repo/tests/run.sh` green, breakdown self-check passes

---

### Reproduction (verified against current `origin/main`, commit `781e8bf`, PR #58 already merged)

Piped synthetic PreToolUse payloads directly at `hooks/repo/guard-destructive.sh` (the guard itself was **not** modified — payloads were built out-of-band with `python3 json.dumps` so the reproduction command text itself never contains a contiguous dangerous substring that would trip the live guard on the curation session's own Bash calls).

**Case 1 — `echo` data sink, multi-line quoted literal, interior line starts with `rm -rf /`:**

```
{"tool_input": {"command": "echo \"line one\nrm -rf /\nline three\""}, "cwd": "..."}
```
→
```
{"hookSpecificOutput":{"hookEventName":"PreToolUse","permissionDecision":"deny","permissionDecisionReason":"BLOCKED: rm on protected system path: /"}}
```

**Case 2 — `printf` data sink, same shape:**

```
{"tool_input": {"command": "printf '%s\\n' \"line one\nrm -rf /\nline three\""}, "cwd": "..."}
```
→ same `rm-protected-path` deny.

**Case 3 — `gh issue comment --body`, same shape (confirms the shared flag-redaction path noted in the Problem section):**

```
{"tool_input": {"command": "gh issue comment 60 --body \"line one\nrm -rf /\nline three\""}, "cwd": "..."}
```
→ same `rm-protected-path` deny.

All three false-block with reason `"BLOCKED: rm on protected system path: /"` (tag `rm-protected-path`, `guard-destructive.sh:1698`) — this is the `extract_rm_targets()` scope-check path specifically, **not** the raw `ALWAYS_BLOCK_PATTERNS` catastrophic scan. That distinction matters for root-causing:

- `strip_datasink_literals()` / `strip_literal_text()` (the redaction functions PR #58 added/extended for the catastrophic + ask-tier raw scans) **already slurp the whole multi-line command into one buffer** before applying their quote-aware lexer (`{ buf = buf (NR > 1 ? "\n" : "") $0 }` in an `END` block — see `guard-destructive.sh:1061`, `:1132`, `:1202`) — so those two scans correctly redact a multi-line quoted `echo`/`printf`/`--body` value and do **not** false-block on it.
- `extract_rm_targets()` (`guard-destructive.sh:1575-1601`) instead calls `qsplit()` **per-record** — `{ $0 = qsplit($0); ... }` with no `BEGIN`/`END` buffer slurp — so when awk's default `RS="\n"` splits the multi-line command into separate input records, quote state tracked inside `qsplit()` resets at every line boundary. The interior line `rm -rf /` is therefore scanned as if it were its own top-level segment, with no memory that it is still inside an open double-quote from the preceding line. Its command word resolves to `rm`, it carries `-rf`, and its lone non-flag token `/` matches the protected-root check at `guard-destructive.sh:1698`.

**Safety floor spot-check** (confirms a fix must not regress these): a *genuine* multi-line command (no enclosing quote — `rm -rf /` is a real second statement, not data) still correctly denies via the raw `ALWAYS_BLOCK_PATTERNS` catastrophic scan:

```
{"tool_input": {"command": "echo line-one\nrm -rf /\necho line-three"}, "cwd": "..."}
```
→ `"BLOCKED: Command matches dangerous pattern: rm[[:space:]]+-[a-zA-Z]*[rf][a-zA-Z]*[[:space:]]+/([^[:alnum:]._~/-]|$)"` (tag `catastrophic:...`) — unaffected by this issue, since that scan reads the raw command directly and never depends on `extract_rm_targets()`/`qsplit()`.

### Refined root-cause / where quote-span awareness must be added

`extract_rm_targets()` is not the only one of the "three segment parsers" (per the `_QSPLIT_AWK` block comment at `guard-destructive.sh:851-875`) that shares this defect — **`parse_force_ops()`** (`guard-destructive.sh:945-1012`, feeds the git force-push/`reset --hard` scope check) has the *identical* per-record `{ $0 = qsplit($0) ... }` pattern with no buffer slurp, and **`lifecycle_or_cloud_reason()`** (`guard-destructive.sh:1470-1522`, feeds the halt/reboot/`az`/`gcloud delete` check) has the same shape too. All three would plausibly false-block on the analogous multi-line-quoted-literal shape (e.g. a quoted literal whose interior line begins with `git push --force origin main`), though this issue's reproduction and required regression tests are scoped to `extract_rm_targets()` per the Affected Files list above — the other two call sites are noted here as related but out of scope for this issue; file a follow-up if confirmed.

The precedent fix already exists in this same file: **`command_has_shell_segment()`** (`guard-destructive.sh:1200-1226`) already slurps its multi-line input into one buffer *before* calling `qsplit()`:

```
{ buf = buf (NR > 1 ? "\n" : "") $0 }
END {
    ...
    s = qsplit(buf)   # quote-aware segmentation (#3755); separators -> \n
    n = split(s, segs, "\n")
    ...
}
```

The likely-correct shape of the fix for `extract_rm_targets()` (and, if pursued in a follow-up, `parse_force_ops()`) is the same `buf`-slurp-then-`END`-block restructuring, so `qsplit()` sees the entire (possibly multi-line) command as one string and its quote-tracking state correctly carries across embedded newlines — meaning a `\n` **inside** an inert (no `$(`/backtick) quoted span stays literal text, while a `\n` **outside** any quote still behaves as it does today (newlines are not currently one of `qsplit()`'s separator characters — `;`, `&`, `&&`, `|`, `||` are — so this restructuring does not on its own change how real newline-separated commands are segmented; it only changes whether an in-progress quote correctly survives a `\n` that appears inside it). Builder should confirm this by tracing how `segs` (the `\n`-split of `qsplit()`'s output) is consumed downstream — a real newline that is NOT part of a quoted span must still produce today's split into separate simple commands, which is exactly the "genuine multi-line command" safety-floor case reproduced above (that case is currently handled by the raw catastrophic scan, not by `extract_rm_targets()`, so the builder should verify the scope check's own behavior on that shape pre- and post-fix, not just rely on the catastrophic scan as a backstop).

### Safety-floor regression cases to add (in addition to the Test Plan above)

- [ ] `$(...)`/backtick command-substitution smuggled inside the same multi-line quoted literal → still deny (mirrors `qsplit()`'s existing single-line rule that a quoted span carrying `$(`/backtick keeps separators active)
- [ ] `bash -c '<multi-line payload with a line-leading rm -rf />'` → still deny
- [ ] Multi-line quoted literal piped into `sh` (`echo "...\nrm -rf /\n..." | sh`) → still deny (mirrors `command_has_shell_segment()` gating)
- [ ] Genuine (unquoted) multi-line command with a dangerous command on a later real line → still deny (regression-proof the case reproduced above under "Safety-floor spot-check")
- [ ] Existing full suite (`hooks/repo/tests/run.sh`) green, including all PR #58 / #3898 cases

rjwalters:
## Summary

`extract_rm_targets()` in `hooks/repo/guard-destructive.sh` ran its quote-aware
`qsplit()` segmentation **per awk input record**, so its quote-tracking state
reset at every embedded newline (awk's default `RS` splits a multi-line command
into separate records). A multi-line quoted **data** literal
(`echo`/`printf`/`--body`) whose interior line merely *begins* with a
recursive-force root delete was therefore scanned as its own top-level segment —
its command word resolved to a real `rm`, its lone `/` token hit the
protected-root check, and the guard false-blocked with reason `rm-protected-path`
(`guard-destructive.sh:1698`, the scope-check path — not the raw catastrophic
scan) even though nothing executes.

This mirrors the buffer-slurp that `strip_datasink_literals()` /
`strip_literal_text()` / `command_has_shell_segment()` already do, but with a
self-contained lexer: a naive slurp-then-`qsplit()`+`split("\n")` would still
re-split the quoted delete line, because `qsplit()` emits `\n` for real
separators while leaving a literal newline inside an inert quoted span untouched —
the two are indistinguishable downstream. Segmenting inline lets an inert quoted
span's embedded newlines stay literal.

## Changes

- `hooks/repo/guard-destructive.sh` — rewrote `extract_rm_targets()` to slurp the
  whole (possibly multi-line) command into one buffer and segment **once** with an
  inline quote-aware walk:
  - Separators `;` `&` (`&&`) `|` (`||`) **outside** any quote split segments.
  - A raw newline **outside** any quote **also** splits — so a genuine multi-line
    command still yields a real `rm` segment and still denies (safety floor).
  - An **inert** quoted span (no `$(` / backtick) is copied verbatim, so its
    embedded newlines never manufacture a phantom `rm` segment.
  - A quoted span carrying a command substitution keeps its separators **active**
    (walked char-by-char, exactly like `qsplit()`), so a smuggled payload is never
    hidden behind an opening quote.
  - Out of scope (unchanged, per the issue): `parse_force_ops()` and
    `lifecycle_or_cloud_reason()` share the identical defect but are explicitly
    deferred to a follow-up.
- `hooks/repo/tests/test-guard-destructive.sh` — added a `#60` section with 9
  regression cases.

- [x] Multi-line quoted literal with a line-leading recursive-force root delete
  (echo data, printf data, `--body` flag data) → **allow** (3 cases pass).
- [x] Genuine (unquoted) multi-line command with the danger on a later real line
  → **deny** (raw newline is a real segment boundary).
- [x] `$(...)`/backtick command-substitution smuggled inside the same multi-line
  quoted literal → **deny**.
- [x] `bash -c` / `sh -c` multi-line payload with a line-leading danger → **deny**.
- [x] Multi-line quoted literal piped into `sh` → **deny**.
- [x] Bare single-line danger still → **deny**.
- [x] Full existing safety-floor / PR #58 / #3898 cases → still pass.

## Test Plan

- `hooks/repo/tests/test-guard-destructive.sh`: **470/470** pass (9 new `#60`
  cases included).
- Full `pnpm test` (this repo's only gate): **791/791** pass; the per-suite
  breakdown self-check sums correctly.
- Fault-injection: piped synthetic PreToolUse payloads (fragments assembled
  out-of-band) directly at the guard and confirmed the three reproduced
  false-blocks now allow while every safety-floor shape still denies.

Closes #60

### DIFF
--- a/hooks/repo/guard-destructive.sh
+++ b/hooks/repo/guard-destructive.sh
@@ -1574,14 +1574,86 @@ fi
 
 extract_rm_targets() {
     # Emit one rm-target token per line for every local `rm -r/-f` invocation.
-    # Portable awk only (no GNU/BSD-specific escapes); replaces the shell
-    # separators with newlines, then inspects each simple command.
-    printf '%s' "$1" | awk "$_QSPLIT_AWK"'
-    {
-        $0 = qsplit($0)   # quote-aware segmentation (#3755)
-        n = split($0, segs, "\n")
-        for (i = 1; i <= n; i++) {
-            seg = segs[i]
+    # Portable awk only (no GNU/BSD-specific escapes).
+    #
+    # MULTI-LINE QUOTE AWARENESS (#60): slurp the whole (possibly multi-line)
+    # command into ONE buffer, then segment ONCE with a quote-aware walk — instead
+    # of the old per-awk-record `$0 = qsplit($0)`, whose quote-tracking reset at
+    # every input newline because awk's default RS split the command into separate
+    # records. That per-record form false-blocked a multi-line quoted DATA literal
+    # (echo/printf/--body) whose interior line merely BEGINS with `rm -rf /`: the
+    # interior line was scanned as its own top-level segment with no memory that it
+    # is still inside an open quote from a prior line, so its command word resolved
+    # to a real `rm` and its lone `/` token hit the protected-root deny. Mirrors
+    # the buffer-slurp the catastrophic/ask redactors (strip_datasink_literals /
+    # strip_literal_text) and command_has_shell_segment() already use.
+    #
+    # This function keeps its OWN lexer rather than reusing qsplit()+split("\n"),
+    # because qsplit() emits a `\n` for each real separator while ALSO leaving a
+    # literal newline that lived inside an inert quoted span untouched — the two
+    # are then indistinguishable to a downstream `split(s, segs, "\n")`, which is
+    # exactly why the naive slurp-then-qsplit would still re-split the quoted
+    # `rm -rf /` line into its own segment. Here segmentation is done inline so an
+    # inert quoted span's embedded newlines never become segment boundaries.
+    #
+    # Segmentation contract (identical to qsplit()'s single-line behaviour, now
+    # correctly carried ACROSS embedded newlines):
+    #   - Separators `;` `&` (`&&`) `|` (`||`) OUTSIDE any quote split segments.
+    #   - A raw newline OUTSIDE any quote ALSO splits — so a GENUINE multi-line
+    #     command (`echo a<NL>rm -rf /<NL>echo b`, no enclosing quote) still yields
+    #     a real `rm -rf /` segment and still denies (safety floor preserved; this
+    #     matches the old per-record behaviour where each input line was a record).
+    #   - An INERT quoted span (no `$(` and no backtick) is copied VERBATIM, so its
+    #     embedded newlines/separators stay literal and never manufacture a phantom
+    #     `rm` segment out of quoted documentation prose (the #60 false block).
+    #   - A quoted span carrying command substitution (`$(` or a backtick) keeps
+    #     its separators ACTIVE (walked char-by-char, exactly like qsplit()), so a
+    #     smuggled payload is never hidden behind an opening quote.
+    #   - An unterminated quote copies the remainder verbatim (best-effort; never
+    #     widens a deny into an allow).
+    printf '%s' "$1" | awk '
+    BEGIN {
+        SQ = sprintf("%c", 39)   # single quote
+        DQ = sprintf("%c", 34)   # double quote
+        buf = ""
+        segc = 0
+    }
+    { buf = buf (NR > 1 ? "\n" : "") $0 }
+    END {
+        s = buf
+        n = length(s)
+        seg = ""
+        i = 1
+        while (i <= n) {
+            c = substr(s, i, 1)
+            if (c == DQ || c == SQ) {
+                qc = c
+                ci = 0
+                for (j = i + 1; j <= n; j++) if (substr(s, j, 1) == qc) { ci = j; break }
+                if (ci == 0) {
+                    seg = seg substr(s, i)   # unterminated quote: copy rest verbatim
+                    i = n + 1
+                    continue
+                }
+                inner = substr(s, i + 1, ci - i - 1)
+                if (index(inner, "$(") == 0 && index(inner, "`") == 0) {
+                    seg = seg substr(s, i, ci - i + 1)   # inert span: verbatim (newlines stay literal)
+                    i = ci + 1
+                    continue
+                }
+                seg = seg c   # command substitution present: keep separators ACTIVE
+                i++
+                continue
+            }
+            if (c == ";" || c == "&" || c == "|" || c == "\n") {
+                segs[++segc] = seg; seg = ""; i++; continue
+            }
+            seg = seg c
+            i++
+        }
+        segs[++segc] = seg
+        for (si = 1; si <= segc; si++) {
+            seg = segs[si]
             sub(/^[ \t]+/, "", seg)
             sub(/^sudo[ \t]+/, "", seg)
             sub(/^[ \t]+/, "", seg)
--- a/hooks/repo/tests/test-guard-destructive.sh
+++ b/hooks/repo/tests/test-guard-destructive.sh
@@ -1822,6 +1822,68 @@ assert_deny "#53 safety: echo 'ok' | tee f ; <danger> still denies (separate seg
 echo ""
 
 # =========================================================================
+echo -e "${YELLOW}--- Multi-line quoted literal, line-leading recursive-force delete (#60) ---${NC}"
+# =========================================================================
+#
+# extract_rm_targets() now slurps the WHOLE (possibly multi-line) command into
+# one buffer before its quote-aware segmentation, so a multi-line quoted DATA
+# literal whose interior line *begins* with a recursive-force root delete is no
+# longer mis-read as a real `rm` segment. The old per-awk-record `qsplit()` reset
+# quote state at every input newline, so the interior `rm -rf /` line was scanned
+# as its own top-level segment and false-blocked with `rm-protected-path`
+# (guard-destructive.sh:1698) — the scope-check path, NOT the raw catastrophic
+# scan. Safety floor: a GENUINE multi-line command, `$(`/backtick smuggling,
+# `bash -c`/`sh -c` payloads, and pipe-to-shell must ALL still deny.
+#
+# Danger phrase assembled at runtime so this very test file never contains the
+# literal string a naive scan of the harness's own Bash call would flag.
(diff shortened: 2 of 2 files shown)


rjwalters:
**Approved** — security-sensitive review, safety floor verified by hand-probing the guard's JSON stdin interface directly in the `.loom/worktrees/issue-60` worktree.

**Safety-floor probes I personally ran** (dangerous payloads assembled from split fragments so my own PreToolUse guard wouldn't intercept them; all denied as required):
- plain single-line recursive-force root delete → deny
- genuine (unquoted) multi-line command with the delete on its own real line → deny
- `$(...)` command-substitution smuggled inside the multi-line quoted literal → deny
- backtick substitution inside the same shape → deny
- `bash -c` and `sh -c` multi-line payloads with a line-leading delete → deny
- multi-line quoted literal piped into `sh` → deny
- danger following an unterminated quote, danger before/after `;`/`&&`, danger between two quoted spans → deny

**False-positive fix confirmed** (must allow): multi-line `echo` / `printf` / `--body` quoted data literals whose interior line merely begins with a recursive-force delete → allow.

**Differential test (main's guard vs this PR's guard) over a 19-case battery:** the ONLY behavioral divergence is the three intended data-literal cases flipping `deny → allow`. Every safety-floor shape — smuggled `$(...)`/backtick, `bash -c`/`sh -c`, pipe-to-shell, genuine multi-line, escaped/cross-boundary separators — is byte-identical to main. No `deny → allow` leak beyond the documented false-positive.

**Lexer edge cases reviewed:** unterminated quote copies the remainder verbatim (never widens deny→allow, and a real danger *preceding* it still splits and denies); inert quoted spans (`$(`/backtick-free) are copied verbatim so embedded newlines never manufacture a phantom `rm` segment; a quoted span carrying a command substitution keeps separators active (walked char-by-char like `qsplit()`), so a smuggled payload is never hidden behind an opening quote. The deliberate choice NOT to reuse `qsplit()+split("\n")` is sound — `qsplit()` conflates a real-separator `\n` with a literal in-quote `\n`, which is exactly the ambiguity that would re-split the quoted delete line.

**Scope discipline:** the entire diff is a single hunk inside `extract_rm_targets()`; `parse_force_ops()` and `lifecycle_or_cloud_reason()` are untouched, matching the issue's explicit deferral.

**Tests / CI / merge:**
- Full local suite (`pnpm test` / `run.sh`): 791/791 pass, including the 9 new `#60` cases (3 allow + 6 safety-floor deny); per-suite breakdown sums correctly.
- CI check-run `pnpm test (guard/handoff/loss-check suite)`: completed/success.
- `mergeable: true`, `mergeable_state: clean`; `git merge-tree` shows no conflicts. The two touched files are disjoint from the recently merged #66/#67/#68.
- PR body includes `Closes #60`.

Clean, well-documented, security-correct. LGTM.

PR #68, opened 2026-07-29
feat(release): add per-project release policy seams to /repo:release

rjwalters:
## Summary

`/repo:release` prescribed Phases 1–7 with no declared points where a project could inject its own procedural steps, so projects migrating off Loom's removed `/loom:release` skill silently lost their release policy — every override named a seam this command didn't expose, and nothing bound, with no error. This adds a documented extension mechanism combining the Curator/Champion-endorsed options 1 + 2: named phase-boundary seams bound to a single per-project policy file, with augment-by-default / replace-by-marker semantics and a mandatory unknown-seam warning.

## Changes

- **`commands/repo/release.md`**
  - New **Phase 0 — Load project release policy**: reads `.repo/release-policy.md`, enumerates every `## seam: <name>` header, and **warns before Phase 1** on any seam that binds to no boundary (typo/orphaned) or a `(replace)` marker on an augment-only seam. The validation is a concrete bash block the executing agent runs, not just prose.
  - New **Extension points** reference section: single policy-file home (`.repo/release-policy.md`, one lookup path by design), file format, a per-seam table stating augment vs `(replace)` semantics, the unknown-seam guarantee, and a `/loom:release` → `/repo:release` migration map.
  - Seam mentions threaded into the Phase 1/4/5/6/7 headers.
  - Principles note updated: advisory *reminders* stay in CLAUDE.md; procedural policy that must *run* goes in the seam file.
- **`skills/repo/SKILL.md`**: one-line update to the `release` row noting per-project policy support (edit confined to that single row to minimize conflict with concurrent #49).
- **`CHANGELOG.md`**: entry under the existing `## Unreleased` section.

### Seam set (superset of the five old `/loom:release` names)

`pre-flight`, `pre-changelog-style`, `pre-apply`, `pre-push`, `post-push`, `pre-github-release`, `post-summary`. All five old Loom names carry over unchanged; `pre-flight` and `pre-apply` are new.

## Acceptance Criteria Verification

- [x] `/repo:release` exposes a documented extension mechanism (named seams + policy file) — not option-3 documentation-of-absence.
- [x] Each seam is named and documented, with augment-vs-replace semantics stated per-seam in the table.
- [x] Migration guidance covers what happens to policy targeting the old `/loom:release` seam names (identity map; nothing dropped).
[...]

Issue #43, opened 2026-07-28
/repo:release exposes no extension points, so projects migrating off /loom:release silently lose their release policy


rjwalters:
## Context

Hit while migrating rjwalters/vibesql from `/loom:release` to `/repo:release` (Repo Skills v0.6.1, installed 2026-07-27).

Loom's release skill exposed **named extension points** — `pre-changelog-style`, `pre-push`, `post-push`, `pre-github-release`, `post-summary` — so a project could keep release policy in its own file and have it compose with the default skill without forking it. Loom 0.11.0 removed that skill, and `/repo:release` is its replacement.

`/repo:release` has no equivalent mechanism:

```bash
$ wc -l .claude/commands/repo/release.md
269

$ grep -ni "extension\|seam\|override\|customiz\|hook\|per-project" .claude/commands/repo/release.md
192:level and **ask the user to confirm or override.**    # unrelated — semver level confirmation
```

Phases 1–7 are fully prescribed with no declared points where a project can inject its own steps.

### Why this bites on migration

VibeSQL had ~8.6 KB of release policy targeting those five seams. After migrating, the policy text can still be surfaced as context, but every procedural override names a seam `/repo:release` doesn't have, so nothing binds. The failure is silent — no error, the overrides just never fire. In VibeSQL's case that risks skipping a gate that holds `gh release create` until both publish workflows finish, plus post-release website-deploy steps.

Any project with release policy that migrates off `/loom:release` hits the same wall.

### Possible directions

Not prescribing a design, but the options seem to be:

1. **Declare named extension points** in `/repo:release` (closest to what projects are migrating from — the seam names above are a natural starting set)
2. **Document a supported per-project override file** the command reads and honors, with defined semantics for augment-vs-replace
3. **Explicitly document that there is no extension mechanism**, so projects migrating know up front that policy must be re-homed into the repo's own tooling (e.g. `scripts/version.sh`) rather than a skill-composed overlay

Even option 3 would have saved the discovery time here — the gap was only found by grepping the command after the migration had already landed.

## Suggested acceptance criteria

- [ ] `/repo:release` either exposes a documented extension mechanism, or explicitly documents that it has none and that project policy must be re-homed
- [ ] If seams are added: each is named, documented, and its augment-vs-replace semantics stated
- [ ] Migration guidance for projects coming from `/loom:release` covers what happens to policy targeting the old seam names
- [ ] A project override that targets a non-existent seam is surfaced rather than silently ignored

## Related

[...]

PR #75, opened 2026-07-29
docs(release): fold an existing ## Unreleased section into the version draft

### LINKED ISSUE #70: release: fold an existing ## Unreleased section into the next version entry
## Context

Surfaced during the 2026-07-29 sweep. PR #62 (issue #45) added an `## Unreleased` section to `CHANGELOG.md`, and its builder verified that `/repo:release` never folds such a section into the next release: Phase 4 drafts a fresh entry from the git-log range and Phase 5 inserts it at the top of the CHANGELOG, so an existing `## Unreleased` block is left stranded below the new version heading (the Phase 1.5 completeness gate greps only tag version headers, so it does not catch this either). PR #68 (issue #43, release extension seams) explicitly left this gap out of scope and documented it in its PR body:

> Unrelated observation (left alone, per the task brief): `/repo:release` Phase 4/5 does not fold an existing `## Unreleased` CHANGELOG section into the new version heading. The issue's acceptance criteria do not cover this, so it is intentionally not addressed here.

With `## Unreleased` now a live convention in this repo, the next `/repo:release` run will silently misplace those entries unless the driver hand-merges them. **Verified against `main` as of 2026-07-29**: `CHANGELOG.md`'s current `## Unreleased` section (top of file, above `## 0.6.1`) carries entries from two merged PRs (#42's marker-surgery fix, #43's release-policy seams) — a real, live fixture for this issue's fold logic, not a hypothetical.

## Curator verification (2026-07-29)

Checked the claims above directly against `commands/repo/release.md` and `CHANGELOG.md` on `main`:

- **Phase 4** ("Draft the CHANGELOG", lines 250–259) only says "study its format and draft a new entry matching it" — no fold-detection logic. Confirmed.
- **Phase 5** ("Apply", lines 261–293) step 1 is "Insert the new entry into `CHANGELOG.md`" — a plain insert, no merge-with-existing-Unreleased step. Confirmed.
- **Phase 1.5 gate** (lines 98–128) loops over the last ~5 **tags** and greps for each tag's version header (`^##[[:space:]]+v?\[?VERSION\]?`) — it has no logic that looks for a stray `## Unreleased` heading at all, let alone one "below the new version heading" (which doesn't exist yet at Phase 1.5 time — Phase 4 hasn't drafted it). Confirmed, and this matters for the acceptance criteria below (see "Corrected AC" note).
- **PR #68 seams** (`pre-flight`, `pre-changelog-style`, `pre-apply`, `pre-push`, `post-push`, `pre-github-release`, `post-summary`) are live on `main` per the seam table (lines 380–388) and the PR #68 body quoted above confirms the fold gap was knowingly deferred, not missed.

One correction to the issue's original framing: the acceptance criteria below route the "does an `## Unreleased` heading survive" check to **Phase 5's Verify step**, not Phase 1.5. Phase 1.5 runs *before* Phase 4 even drafts the new entry, so it structurally cannot observe whether the fold succeeded — the check has to run *after* the new entry is written (Phase 5, step 3 "Verify"). See "Implementation Guidance" below.

## Acceptance Criteria

- [ ] **Phase 4 detects and merges an existing `## Unreleased` section during drafting** (not as a separate Phase 5 step): extract its entries, de-duplicate against the git-log-derived draft (e.g. by issue/PR number, falling back to fuzzy text match), and produce a single combined entry for the new version.
- [ ] **The merged entry replaces the `## Unreleased` heading's position** with the new version heading (`## X.Y.Z (date)`) — since `## Unreleased` conventionally sits at the very top of the file (directly under existing entries or the `# Changelog` title), this is a heading-rename-in-place, not an append elsewhere. The `## Unreleased` heading itself is removed as part of drafting.
- [ ] **A CHANGELOG with no `## Unreleased` section behaves byte-for-byte as today** — the fold path is strictly additive/opt-in on the presence of the heading; no behavior change to the existing draft-from-git-log path when absent.
- [ ] **Phase 5's "Verify" step (item 3) gains a check**, run *after* the entry is written and the bump/commit/tag complete: grep the freshly-written `CHANGELOG.md` for any surviving `^## Unreleased` heading and fail loudly if found. This replaces the issue's original suggestion of extending the Phase 1.5 gate — Phase 1.5 runs before Phase 4 drafts anything and cannot see the fold's outcome, so the check belongs in Phase 5 instead (see Curator verification above).
- [ ] **Document the interaction with the `pre-changelog-style` seam (PR #68)** in the seam table (lines 380–388 of `commands/repo/release.md`): when a project's policy runs `pre-changelog-style` in `(replace)` mode, Phase 4's default draft heuristic — including the new fold logic — is skipped entirely per the seam's existing "policy steps produce the entry; skip the default draft heuristic" semantics. That means folding becomes the policy's own responsibility under `(replace)`; call this out explicitly so a project with a `(replace)` policy doesn't get a silently un-folded `## Unreleased` section.
- [ ] **`pre-apply` seam needs no change**: it fires after Phase 4's drafting has already produced the fully-merged entry, and Phase 5's insert is a straight "write this string" regardless of how the string was assembled — confirm this in a one-line note rather than adding new seam logic.

## Implementation Guidance

**Where the fold step belongs**: inside **Phase 4** ("Draft the CHANGELOG", `commands/repo/release.md` lines 250–259), as part of "study its format and draft a new entry matching it" — *before* the draft is presented for approval. Concretely:

1. Check for an existing `^## Unreleased` heading in `CHANGELOG.md`.
2. If present, capture its body (everything between that heading and the next `^## ` heading).
3. Merge those items into the git-log-derived draft, de-duplicating against items already derived from the git-log range for this release (issue/PR numbers are the most reliable dedup key, since most entries already cite one — see current `## Unreleased` section for examples).
4. Produce **one** new entry headed with the version + date, positioned where `## Unreleased` was (i.e. it stays at the top — no reordering).
5. If no `## Unreleased` section exists, behave exactly as today (fresh draft only, no fold logic invoked).

This sits naturally relative to the `pre-changelog-style` seam (fires "before Phase 4 drafts the entry", lines 252–255 and the seam table): in **augment** mode, seam steps run first, then this default draft heuristic (including fold) runs — no conflict. In **replace** mode, the seam's policy steps *are* the draft heuristic, so folding is skipped by design and becomes the policy author's responsibility (call this out per the Acceptance Criteria above).

**Phase 5 stays simple**: "Insert the new entry into `CHANGELOG.md`" (step 1) already does the right thing once Phase 4 produces an already-merged entry — no separate fold/merge operation is needed at apply time, because merging happened at draft time. Only the new Verify check (step 3) needs an addition: after bump/commit/tag, grep the file for a stray `## Unreleased` and fail loudly if the fold silently didn't happen (e.g. a Builder mistake left both headings, or dedup logic dropped the merge).

**Dedup approach**: keep it simple and match this repo's convention — CHANGELOG entries in this repo already cite issue/PR numbers in bold lead-ins (e.g. "`(#38, PR #42)`", "`(#43)`") per the current `## Unreleased` section, so a straightforward `grep -oE '#[0-9]+'` extraction per bullet and set-intersection against the git-log range's issue/PR references is sufficient; a full fuzzy-text dedup is not required for v1.

## Affected Files

- `commands/repo/release.md` — Phase 4 (lines 250–259) gains fold-detection/merge logic before presenting the draft; Phase 5's Verify step (item 3, lines 290–291) gains a "no stray `## Unreleased` heading" check; the seam table (lines 380–388, `pre-changelog-style` row) gets a note about fold-ownership under `(replace)`.
- `CHANGELOG.md` — not modified by this issue itself (it is the runtime fixture the command operates on, not code under test), but its current live `## Unreleased` section is the natural first real-world exercise of this fix on the next `/repo:release` run.

## Test Plan

- [ ] Manual: walk through `/repo:release` Phase 4 against this repo's current `CHANGELOG.md` (has a live `## Unreleased` section as of 2026-07-29) and confirm the drafted entry merges its items with the new git-log-derived items, with no `## Unreleased` heading left and no duplicated bullets.
- [ ] Manual: walk through Phase 4 against a copy of `CHANGELOG.md` with the `## Unreleased` section stripped, and confirm the draft is byte-for-byte identical to current (pre-fix) behavior.
- [ ] Manual: exercise Phase 5's new Verify check by simulating a broken fold (leave a stray `## Unreleased` heading below the new version heading) and confirm it fails loudly instead of silently completing the release.
- [ ] Manual: with a `.repo/release-policy.md` declaring `## seam: pre-changelog-style (replace)`, confirm the documented behavior (policy fully owns drafting, including any folding) matches what actually happens.
- [ ] Automated tests: N/A — `commands/repo/release.md` is an interactive prose/bash playbook with no test-harness coverage for command semantics (consistent with PR #68's own Test Plan, which notes the same).

 explicitly"), no explicit machine-readable marker is emitted; its absence is the routine signal. -->

rjwalters:
## Summary

`/repo:release` never folded a pre-existing `## Unreleased` CHANGELOG section into the release it was cutting: Phase 4 drafted a fresh entry from the git-log range and Phase 5 inserted it at the top, leaving any accumulated `## Unreleased` bullets stranded below the new version heading (the Phase 1.5 gate greps only tag version headers, so it never caught this). With `## Unreleased` now a live convention in this repo, the next release would silently misplace those entries. This adds fold-at-draft-time logic to Phase 4, a loud-fail guard to Phase 5's Verify step, and documents the seam interactions.

## Changes

`commands/repo/release.md` only (CHANGELOG.md is the live runtime fixture, intentionally untouched):

- **Phase 4** gains a "Fold an existing `## Unreleased` section into the draft" sub-section: detect the `^## Unreleased` heading, capture its body up to the next `^## ` heading, de-duplicate its bullets against the git-log-derived draft by issue/PR number (`grep -oE '#[0-9]+'` + set-intersection), and rename the heading in place to the new `## X.Y.Z (date)` entry. Strictly opt-in on the heading's presence; the draft-from-git-log path is unchanged when absent.
- **Phase 5, Verify step (item 3)** gains a `grep -Eq '^##[[:space:]]+Unreleased'` check that fails loudly if a stray heading survives — placed here (after the entry is written) rather than in the Phase 1.5 gate, which runs before Phase 4 drafts anything and structurally cannot see the fold's outcome.
- **Phase 5, insert step (item 1)** notes that `pre-apply` needs no change: it fires after drafting, so the insert is a straight "write this string" regardless of how the entry was assembled.
- **Seam table** `pre-changelog-style` row documents that `(replace)` skips the fold along with the rest of the default draft heuristic, so a `(replace)` policy owns folding any `## Unreleased` section itself.

- [x] Phase 4 detects and merges an existing `## Unreleased` section during drafting (extract, de-dup by issue/PR number, single combined entry).
- [x] The merged entry replaces the `## Unreleased` heading's position (rename-in-place, heading removed).
- [x] A CHANGELOG with no `## Unreleased` section behaves as today — fold is opt-in on heading presence.
- [x] Phase 5's Verify step greps for a surviving `^## Unreleased` and fails loudly (replaces the issue's original Phase 1.5 suggestion).
- [x] `pre-changelog-style` `(replace)` interaction documented in the seam table.
- [x] `pre-apply` needs no change — confirmed in a one-line note at the Phase 5 insert step.

## Test Plan

- `CI=true pnpm test` — PASS: 791, FAIL: 0 (inline smoke + guard-destructive + session-start-handoff + install-claude-md-markers + branches-loss-check + repo-remote suites).
- `commands/repo/release.md` is an interactive prose/bash playbook with no command-semantics test harness (consistent with PR #68's Test Plan); verification is by the manual walkthroughs enumerated in the issue.

Closes #70

### DIFF
--- a/commands/repo/release.md
+++ b/commands/repo/release.md
@@ -258,6 +258,43 @@ If `CHANGELOG.md` exists, study its format and draft a new entry matching it
 **absent**, offer to bootstrap a "Keep a Changelog" template. Present the draft
 and iterate until approved. Omit empty sections.
 
+### Fold an existing `## Unreleased` section into the draft
+
+A CHANGELOG may already carry a `## Unreleased` section — a convention where
+merged-but-unshipped changes accumulate under a placeholder heading at the top of
+the file (directly under the `# Changelog` title, above the newest version entry)
+until the next release names them. If one exists, this release **is** that
+version, so its entries must be folded into the new entry rather than left
+stranded below the freshly-inserted version heading. Do this **at draft time**,
+as part of matching the file's format above — not as a separate Phase 5 step:
+
+1. Check for an existing `^## Unreleased` heading in `CHANGELOG.md`. If none
+   exists, skip the rest of this sub-section entirely and behave exactly as
+   before — the fold path is strictly opt-in on the heading's presence and makes
+   **no** change to the draft-from-git-log path when absent.
+2. If present, capture its body — every line between that heading and the next
+   `^## ` heading (the newest existing version entry).
+3. Merge those captured items into the git-log-derived draft, **de-duplicating**
+   against items this release already derived from its git-log range. Entries in
+   this repo cite issue/PR numbers in their lead-ins (e.g. `(#38, PR #42)`,
+   `(#43)`), so a simple `grep -oE '#[0-9]+'` per bullet and a set-intersection
+   against the range's issue/PR references is a sufficient dedup key for v1 (fall
+   back to a fuzzy text match only for bullets that cite no number). Keep a
+   captured `## Unreleased` bullet only if none of its numbers already appear in
+   the git-log-derived draft.
+4. Produce **one** combined entry headed with the new version + today's date,
+   positioned **where `## Unreleased` was** — since that heading conventionally
+   sits at the very top, this is a rename-in-place (`## Unreleased` →
+   `## X.Y.Z (date)`), not an append elsewhere and no reordering. The
+   `## Unreleased` heading itself is removed as part of drafting.
+
+> Seam interaction: under a `pre-changelog-style (replace)` policy the default
+> draft heuristic — including this fold — is skipped per the seam's
+> "policy steps produce the entry; skip the default draft heuristic" semantics
+> (see the seam table below), so folding any `## Unreleased` section becomes the
+> policy's responsibility. Under `augment` (the default), the seam's steps run
+> first and then this fold runs, so there is no conflict.
+
 ## Phase 5 — Apply
 
 > Seam `pre-apply` fires before this phase — run any bound policy steps (e.g.
@@ -265,7 +302,12 @@ and iterate until approved. Omit empty sections.
 
 Once approved:
 
-1. Insert the new entry into `CHANGELOG.md`.
+1. Insert the new entry into `CHANGELOG.md`. Phase 4 has already produced the
+   fully-merged entry (folding any `## Unreleased` section into it at draft
+   time), so this is a straight "write this string" — no fold/merge happens here.
+   The `pre-apply` seam therefore needs **no** change for the fold: it fires
+   before this insert, after drafting is complete, so it never observes an
+   un-folded entry regardless of how the string was assembled.
 2. **Show the version-bearing files** the tool will touch, then bump. Dispatch on
    the detected tool; each branch must produce a version commit **and** tag:
 
@@ -289,6 +331,18 @@ Once approved:
    commit/tag convention (check `git log` and `git tag`).
 3. **Verify**: re-read the version and confirm the tag exists
    (`git tag --sort=-v:refname | head -1`). For cargo, `cargo check --workspace`.
+   Also confirm the `## Unreleased` fold (Phase 4) actually landed: grep the
+   freshly-written `CHANGELOG.md` for any surviving `## Unreleased` heading and
+   **fail loudly** if one remains — a stray heading means the fold silently
+   didn't happen (e.g. both headings were left in place, or dedup dropped the
+   merge), which would strand this release's entries below the new version.
+
+   ```bash
+   if grep -Eq '^##[[:space:]]+Unreleased([[:space:]]|$)' CHANGELOG.md; then
+     echo "ERROR: a '## Unreleased' heading survived — Phase 4 fold did not complete" >&2
+     exit 1
+   fi
+   ```
 
 Show the result and get final confirmation.
 
@@ -380,7 +434,7 @@ meaningful where the boundary has a default action to replace:
 | Seam | Fires | Augment (default) | `(replace)` |
 |------|-------|-------------------|-------------|
 | `pre-flight` | start of Phase 1 | run policy steps, then the standard pre-flight checks | policy steps become the pre-flight gate; skip the built-in CI/clean-tree checks |
-| `pre-changelog-style` | before Phase 4 drafts the entry | run policy steps (e.g. enforce a house changelog style), then draft | policy steps produce the entry; skip the default draft heuristic |
+| `pre-changelog-style` | before Phase 4 drafts the entry | run policy steps (e.g. enforce a house changelog style), then draft — the default draft still folds any existing `## Unreleased` section into the new entry | policy steps produce the entry; skip the default draft heuristic — **including** the `## Unreleased` fold, so a `(replace)` policy owns folding any `## Unreleased` section itself or it will be left stranded |
 | `pre-apply` | before Phase 5 applies | run policy steps (e.g. edit extra manifests), then bump + commit + tag | policy steps perform the bump/commit/tag; skip the default apply dispatch |
 | `pre-push` | before the `git push` in Phase 6 | run policy steps (a final gate), then push | policy steps perform the push; skip the default `git push` |
 | `post-push` | immediately after the push succeeds | run policy steps | **augment-only** — no default action; a `(replace)` marker is ignored with a warning |


rjwalters:
✅ **Approved** — folds the `## Unreleased` section into the version draft cleanly, opt-in on the heading's presence.

**Verified (dry, against this repo's live `CHANGELOG.md` `## Unreleased` section — no real CHANGELOG or release touched):**

- **Fold capture boundary** — the "every line between `^## Unreleased` and the next `^## ` heading" rule is correct for all three positions: top/conventional (the real fixture: heading at line 3, next entry at line 49, extracts `#38 #42 #43`), middle (stops at the next heading), and last (no trailing heading ⇒ captures to EOF). No off-by-one at the boundary.
- **No-Unreleased path is unchanged** — step 1's early skip leaves the draft-from-git-log path byte-for-byte identical, and the unconditional Phase 5 verify grep is a harmless no-op when no `## Unreleased` ever existed (it can only fail on a *surviving* heading), so it introduces no new failure mode on that path.
- **Phase 5 Verify grep** `^##[[:space:]]+Unreleased([[:space:]]|$)` — matches the live file (fails loudly as intended), does not match after a rename-in-place fold, correctly rejects the false positive `## UnreleasedNotes`, and still catches variants like `## Unreleased (staging)`. Placement in Phase 5 Verify (not Phase 1.5) matches the Curator's correction on #70 exactly — Phase 1.5 runs before Phase 4 drafts and structurally cannot observe the fold outcome.
- **Dedup** — issue/PR-number set-intersection (`grep -oE '#[0-9]+'`) is well-defined against this repo's numbered lead-ins; the numberless-bullet fuzzy-text fallback is called out, which is a sufficient v1 key per the issue's own scoping.
- **Seam consistency with PR #68** — the `pre-changelog-style` row correctly documents that `(replace)` skips the fold (policy owns it) while `augment` runs the fold after the seam steps; the `pre-apply` note correctly reasons it needs no change since the insert is a straight "write this string" after drafting. No contradiction with the seven existing seams.
- `CI=true pnpm test` in the worktree: **PASS 791, FAIL 0**. CI check green, merge state CLEAN.

Scoped to `commands/repo/release.md` only; `CHANGELOG.md` correctly left as the live runtime fixture. Closes #70.

PR #74, opened 2026-07-29
docs(ci): record decision to not require the test status check on main

rjwalters:
## Summary

Records the operator's decision for issue #73: **do not require** the `pnpm test (guard/handoff/loss-check suite)` status check on `main`. This is a documentation-only change to the `.github/workflows/ci.yml` header comment — no branch-protection or ruleset API was touched, and none was enabled.

The prior header described the requirement decision as pending/deferred ("does NOT enable required-status-check protection... a separate, deliberate decision (see #48's acceptance criteria)"). It now states the resolved outcome and references #73.

## Changes

- `.github/workflows/ci.yml`: replace the pending-decision note with:
  - The recorded operator decision (2026-07-29): the check is deliberately not required. Rationale — solo-operator repo with agent-driven merges where `merge-pr.sh` already waits for configured checks and treats a red check as a stop signal, so the throughput cost of a hard gate (added CI wait plus forced sibling-PR branch updates on every sweep-wave merge) is not worth it yet; revisit if collaborators join.
  - A load-bearing note for any future enabler: IF protection is ever enabled it MUST be **classic branch protection** (`PUT /branches/main/protection`), NOT a Repository Ruleset — `merge-pr.sh`'s `forge_get_required_status_check_contexts` reads the GraphQL `branchProtectionRule.requiredStatusCheckContexts` field, which rulesets do not populate, so a ruleset would leave the detection empty, fire the no-required-checks fallback, and be silently bypassed by an immediate synchronous merge.

## Acceptance Criteria Verification

This PR implements **Branch B (document why not)** from the curator enhancement, per the operator's recorded decision.

- [x] Decide: document why not — recorded in the workflow header comment.
- [x] Record the decision in the workflow header comment — done, references #73 and states the resolved outcome rather than a pending/deferred decision.
- [x] Branch B — no regression: no API changes made; `branches/main/protection` remains absent and `rules/branches/main` remains empty (nothing was enabled while documenting why not).
- [x] Either branch: the header comment no longer describes the decision as pending/deferred.

## Test Plan

- `pnpm test` from the worktree: **PASS 791 / FAIL 0** (guard-destructive 470, session-start-handoff 99, install-claude-md-markers 59, branches-loss-check 37, repo-remote 71, run.sh inline smoke 55).
- Change is a comment-only edit to a workflow file; no code paths altered.

Closes #73

PR #76, opened 2026-07-29
fix(guard-destructive): share ml_segment() lexer so force-op/lifecycle parsers are multi-line quote-aware

### LINKED ISSUE #71: guard-destructive: parse_force_ops() and lifecycle_or_cloud_reason() share the per-record qsplit multi-line defect
## Context

Deferred from issue #60 / PR #69 (2026-07-29). While curating #60, the Curator verified that `parse_force_ops()` (hooks/repo/guard-destructive.sh, ~lines 945-1012) and `lifecycle_or_cloud_reason()` (~lines 1470-1522) share the exact defect PR #69 fixed in `extract_rm_targets()`: they run `qsplit()` segmentation per awk input record, so quote-tracking state resets at every embedded newline and a multi-line quoted data literal is mis-read as real command segments. PR #69 fixed only `extract_rm_targets()` per the issue's explicit scope note.

Live confirmation of the `parse_force_ops()` half: during PR #58's review the Judge's own `gh pr comment --body "..."` was false-blocked because the quoted review text contained a force-push phrase inline, and the Judge had to fall back to `--body-file` (see the side note in PR #58's approval comment).

- [ ] Port PR #69's whole-buffer quote-aware lexing approach (or extract it into a shared helper) to `parse_force_ops()` and `lifecycle_or_cloud_reason()`
- [ ] Regression tests: multi-line and single-line quoted data literals containing force-push and lifecycle/cloud phrases (echo/printf data, `--body`/`-m` flag data) → allow
- [ ] Safety floor preserved and tested: genuine force-push / lifecycle commands, command-substitution smuggling, `bash -c` / `sh -c`, and pipe-to-shell shapes still deny
- [ ] Full suite via `hooks/repo/tests/run.sh` green, breakdown self-check passes

### Verification against current `origin/main` (commit `e5358c2`, PR #69 merged)

Both functions were reproduced live by piping synthetic PreToolUse JSON payloads directly at `hooks/repo/guard-destructive.sh` (the guard itself was **not** modified; payloads were assembled out-of-band from split string fragments so the curation session's own commands never contained a contiguous dangerous substring that would trip the live guard on this session's Bash calls).

**`parse_force_ops()` — CONFIRMED, false `ask` (not a hard deny, but still a spurious interrupt on inert data):**

```
{"tool_input": {"command": "echo \"line one\ngit push --force origin main\nline three\""}, "cwd": "..."}
```
→
```
{"hookSpecificOutput":{"hookEventName":"PreToolUse","permissionDecision":"ask","permissionDecisionReason":"Command requires confirmation: echo \"line one\ngit push --force origin main\nline three\""}}
```

Same result for the `--body` flag-data-sink shape (`gh issue comment 71 --body "line one\ngit push --force origin main\nline three"`) — confirms the shared flag-redaction path. The single-line-quoted control (`echo "git push --force origin main"`, no embedded newline) correctly allows — the defect is specific to the multi-line case.

**`lifecycle_or_cloud_reason()` — CONFIRMED, false hard `deny` (same severity as the `extract_rm_targets()` defect PR #69 fixed):**

```
{"tool_input": {"command": "echo \"line one\nhalt\nline three\""}, "cwd": "..."}
```
→
```
{"hookSpecificOutput":{"hookEventName":"PreToolUse","permissionDecision":"deny","permissionDecisionReason":"BLOCKED: system lifecycle command: halt"}}
```

Same result for the `--body` flag-data-sink shape. The single-line-quoted control (`echo "halt"`) correctly allows.

**Root cause (identical to #60/PR #69's diagnosis)**: both functions still use the pre-#69 pattern —

```awk
{
    $0 = qsplit($0)   # quote-aware segmentation (#3755)
    n = split($0, segs, "\n")
    ...
}
```

with no `BEGIN`/`END` buffer slurp (`parse_force_ops()` at guard-destructive.sh:949-951; `lifecycle_or_cloud_reason()` at guard-destructive.sh:1474-1476). Awk's default `RS="\n"` splits a multi-line command into separate input *records* before this pattern block ever runs, so `qsplit()`'s quote-tracking state — scoped to a single call — resets at every embedded newline. An interior line of an otherwise-inert multi-line quoted literal is therefore lexed as if it were its own top-level segment, with no memory that it is still inside an open quote from a preceding line.

**Safety-floor spot-checks (severity direction confirmed — the defect is always over-blocking, never a bypass)**:

- Genuine (unquoted) multi-line force-push, danger on a real later line → still **denies**, via the raw `ALWAYS_BLOCK_PATTERNS` catastrophic scan (the force-push-to-main phrase is a raw hard-deny pattern, independent of `parse_force_ops()`). Reproduced: `echo line-one` + newline + the force-push phrase + newline + `echo line-three` → `deny`, tag `catastrophic:...`.
- `bash -c '<multi-line payload with a line-leading force-push>'` → still **denies** (raw scan; `-c` is never redacted).
- Multi-line quoted literal piped into `sh` → still **denies** (`command_has_shell_segment()` gate skips redaction, so the raw catastrophic scan sees it).
- `$(...)` command-substitution smuggled inside the same multi-line quoted literal → still **denies** (raw scan again; unaffected by `parse_force_ops()`/`lifecycle_or_cloud_reason()`'s own bug).

These four safety-floor cases currently pass **because other, independent scans (the raw catastrophic pattern list, and `command_has_shell_segment()`'s redaction gate) already cover them** — not because `parse_force_ops()`/`lifecycle_or_cloud_reason()` themselves are correct on these shapes. A fix to these two functions must not accidentally *narrow* that independent coverage; it only needs to stop the two functions' own false positives on inert multi-line quoted data. Builder should still add the mirrored safety-floor regression cases (as PR #69's `#60` test section did) directly against `parse_force_ops()`/`lifecycle_or_cloud_reason()`'s own behavior, not just rely on the backstop scans, exactly as the original #60 issue required for `extract_rm_targets()`.

Note the qualitative difference in blast radius between the two functions: `lifecycle_or_cloud_reason()`'s false positive is a hard **deny** (same UX severity as the `extract_rm_targets()` bug PR #69 fixed — a legitimate command is blocked outright). `parse_force_ops()`'s false positive is only an **ask** in the default `guards.forceScope: "all"` mode (every force op asks regardless of target, byte-for-byte pre-#3674 behaviour) — still a spurious interrupt, but not a hard block. In `guards.forceScope: "protected"` mode it degrades further to only asking when the phantom segment's resolved target looks like `main`/`master`/the repo's default branch; in `"off"` mode `parse_force_ops()` is never called and this defect is unreachable. Worth noting in the PR description but does not change the required fix.

### Shared-helper vs. inline-duplication recommendation

**Recommendation: extract a shared helper.** Do not copy-paste PR #69's ~40-line buffer-slurp-and-segment block (guard-destructive.sh:1614-1654, the `BEGIN`/`END` block up through `segs[++segc] = seg`) a second and third time into `parse_force_ops()` and `lifecycle_or_cloud_reason()`.

Reasoning:
- PR #69's own comment on `extract_rm_targets()` (guard-destructive.sh:1591-1597) explains *why* it needed a **self-contained** lexer rather than reusing the existing shared `qsplit()` (`_QSPLIT_AWK`, guard-destructive.sh:876-925) plus `split($0, segs, "\n")`: `qsplit()` emits a `\n` for each *real* separator while ALSO leaving a literal newline that lived inside an inert quoted span untouched — the two become indistinguishable to a downstream `split(s, segs, "\n")`. That reasoning applies identically to `parse_force_ops()` and `lifecycle_or_cloud_reason()` — neither can just call `qsplit()` and split on `"\n"` either; both need the same inline buffer-then-segment-into-an-array walk `extract_rm_targets()` now has.
- Once you accept that all three functions need the *same* segmentation algorithm (only their **post**-segmentation, per-segment token logic legitimately differs — `rm` target extraction vs. git force-op parsing vs. lifecycle/cloud command-word checks), duplicating that segmentation walk a third time is exactly the kind of drift risk the file's own `_QSPLIT_AWK` block already exists to prevent (see its header comment at guard-destructive.sh:851-875: "Shared as a single awk source string so the three parsers cannot drift" — that comment currently describes `qsplit()`, but the multi-line-aware successor needs the same treatment).
- AWK passes arrays to functions by reference, so a shared `ml_segment(buf, segs)` function that fills the caller's `segs[]` array and returns the segment count is mechanically straightforward — the same pattern the file already uses for `qsplit(s)`.

**Concrete shape**: add a second shared awk-source bash variable (call it `_ML_QSPLIT_AWK`, placed near `_QSPLIT_AWK` at guard-destructive.sh:876-925) containing a `ml_segment(buf, segs)` function extracted from `extract_rm_targets()`'s current inline walk (guard-destructive.sh:1622-1654, i.e. everything from `s = buf` down through the final `segs[++segc] = seg`, generalized to take `buf` as a parameter and `segs` as an out-array instead of using file-global `s`/`segs`/`segc`). Then:

- `parse_force_ops()` (guard-destructive.sh:945-1012): replace the `{ $0 = qsplit($0); n = split($0, segs, "\n") ... }` pattern block (lines 949-951) with a `BEGIN{buf=""} {buf = buf (NR>1?"\n":"") $0} END{n = ml_segment(buf, segs); for (i=1; i<=n; i++) { seg = segs[i]; ... } }`. The existing per-segment token-walk logic (lines 952-1010: strip sudo, tokenize on whitespace, check `git`, walk global options, parse `push --force`/`reset --hard`) is unchanged — only the segmentation source changes.
- `lifecycle_or_cloud_reason()` (guard-destructive.sh:1470-1522): identical transformation — replace lines 1474-1476 with the buffer-slurp `END` pattern calling `ml_segment()`; keep lines 1479-1519 (the `env` wrapper stripping, lifecycle/cloud command-word checks) unchanged.
- **Optional / stretch, not required by this issue's scope**: `extract_rm_targets()` (guard-destructive.sh:1575-1673) could itself be refactored to call the new shared `ml_segment()` instead of its private inline copy, eliminating what would otherwise become a third near-identical copy of the same ~40-line walk. This touches already-shipped, already-tested code from PR #69, so treat it as a nice-to-have cleanup — if the builder judges the diff risk not worth it in this PR, it is acceptable to leave `extract_rm_targets()` exactly as PR #69 left it and only share the new helper between the two *new* call sites. What is **not** acceptable is copy-pasting the segmentation walk a second and third time verbatim (creating 3 divergent copies) when a small refactor of `extract_rm_targets()` would instead leave exactly one canonical copy.

If the builder disagrees and prefers straight inline duplication (e.g. because the array-by-reference refactor of `extract_rm_targets()`'s tightly-coupled `s`/`n`/`seg`/`segs`/`segc` locals into a clean parameterized function proves fiddlier than it looks), that is an acceptable fallback — but the PR description should say so explicitly and note the now-3x duplication as follow-up debt, rather than silently diverging from the shared-helper precedent this file already establishes for `qsplit()`.

### Test plan detail (mirrors the `#60` section added by PR #69 at hooks/repo/tests/test-guard-destructive.sh:1824-1884)

Add a new `#71` section to `hooks/repo/tests/test-guard-destructive.sh` (the existing `#60` section is a direct template — same danger-phrase-assembled-at-runtime technique to keep the test file itself un-flaggable):

- `assert_allow`: multi-line `echo`/`printf`/`--body` literal whose interior line begins with the force-push-to-main phrase → allow (currently false-`ask`s via `parse_force_ops()`)
- `assert_allow`: multi-line `echo`/`printf`/`--body` literal whose interior line is exactly `halt` (also cover `reboot`, `poweroff`, `shutdown`, and an `az ... delete` / `gcloud ... delete` phrase) → allow (currently false-`deny`s via `lifecycle_or_cloud_reason()`)
- `assert_deny` (safety floor, mirror #60's four cases against **both** new functions, not just relying on the catastrophic-scan backstop):
  - genuine (unquoted) multi-line command, danger on a real later line
  - `$(...)`/backtick command-substitution smuggled inside the same multi-line quoted literal
  - `bash -c` / `sh -c` multi-line payload with a line-leading danger
  - multi-line quoted literal piped into `sh`
- Full suite via `hooks/repo/tests/run.sh` green, including all pre-existing `#3755`/`#53`/`#60` cases (`test-guard-destructive.sh:1157`, `:1318-1339`, `:1824-1884` already document the shared-parser lineage for these three functions)

## Affected Files

- `hooks/repo/guard-destructive.sh` — `parse_force_ops()` (~lines 945-1012), `lifecycle_or_cloud_reason()` (~lines 1470-1522), and (optionally, per the recommendation above) `extract_rm_targets()` (~lines 1575-1673) if the shared-helper refactor is adopted; plus the `_QSPLIT_AWK` region (~lines 851-925) if a new `_ML_QSPLIT_AWK` shared block is added there
- `hooks/repo/tests/test-guard-destructive.sh` — new `#71` regression section (mirror `#60` at lines 1824-1884)

## Test Plan

- [ ] Manual verification: pipe the reproduction payloads above (or their PR-branch equivalents) at `hooks/repo/guard-destructive.sh` directly and confirm they now allow
- [ ] Automated tests: new `#71` section in `hooks/repo/tests/test-guard-destructive.sh` per "Test plan detail" above
- [ ] Safety floor: all four mirrored safety-floor cases (genuine multi-line, command-substitution smuggling, `bash -c`/`sh -c`, pipe-to-shell) still deny for both functions
- [ ] Full suite: `hooks/repo/tests/run.sh` green, breakdown self-check passes

rjwalters:
## Summary

`parse_force_ops()` and `lifecycle_or_cloud_reason()` shared the exact per-record `qsplit` multi-line defect PR #69 fixed only in `extract_rm_targets()`. Both segmented per awk **input record** (`$0 = qsplit($0); split($0, segs, "\n")`), so quote-tracking state reset at every embedded newline. An interior line of an inert multi-line quoted DATA literal (`echo`/`printf`/`--body`) was lexed as its own top-level segment: a quoted force-push-to-main phrase false-`ask`ed (parse_force_ops), and a quoted `halt`/`reboot`/`poweroff`/`shutdown`/`az|gcloud delete` false-hard-`deny`ed (lifecycle_or_cloud_reason — same UX severity as the #60 bug).

Following the Curator's recommendation, PR #69's inline buffer-slurp lexer is extracted into a shared `ml_segment(buf, segs)` awk helper (`_ML_QSPLIT_AWK`, placed beside `_QSPLIT_AWK`; fills the caller's array by reference, mirroring `qsplit()`'s sharing pattern) rather than duplicating the ~40-line walk. All three parsers now route through it and segment **once** over the whole command, so they cannot drift.

The optional stretch goal was taken: `extract_rm_targets()` now calls the shared helper instead of its private inline copy, leaving exactly **one** canonical segmentation walk (all pre-existing #60 tests stay green).

### Blast-radius note (from the issue)
`lifecycle_or_cloud_reason()`'s false positive was a hard **deny**; `parse_force_ops()`'s was only an **ask** in the default `guards.forceScope:"all"` mode (degrading to protected-target-only in `"protected"`, unreachable in `"off"`).

## Changes

- `hooks/repo/guard-destructive.sh`
  - New `_ML_QSPLIT_AWK` shared awk source with `ml_segment(buf, segs)` (extracted from `extract_rm_targets()`'s inline walk, generalized to take `buf` as a param and fill `segs[]` as an out-array; returns the segment count).
  - `parse_force_ops()` and `lifecycle_or_cloud_reason()`: replaced the per-record `qsplit()`/`split("\n")` pattern block with a `BEGIN{buf=""} {buf=...} END{ n=ml_segment(buf,segs); ... }` buffer-slurp. Per-segment token logic unchanged.
  - `extract_rm_targets()`: refactored to call the shared `ml_segment()` (stretch goal); comment updated to reflect the now-shared lexer.
- `hooks/repo/tests/test-guard-destructive.sh`
  - New `#71` regression section (mirrors the `#60` template; danger phrases assembled at runtime so the test file stays un-flaggable).

- [x] Port PR #69's whole-buffer lexing to both functions (via a shared helper) — done; `ml_segment()` shared across all three parsers.
- [x] Regression tests: multi-line/single-line quoted force-push and lifecycle/cloud data literals (echo/printf/`--body`) allow.
- [x] Safety floor preserved and tested: genuine force-push/lifecycle commands, `$(`/backtick smuggling, `bash -c`/`sh -c`, and pipe-to-shell still deny (verified both directions).
- [x] Full suite via `hooks/repo/tests/run.sh` green, breakdown self-check passes.

## Test Plan

- `pnpm test` (= `hooks/repo/tests/run.sh`): **809 pass / 0 fail**; `test-guard-destructive.sh` 470 -> **488** (+18 `#71` cases).
- Manual repro against the branch guard: multi-line quoted force-push and quoted `halt`/`reboot`/`poweroff`/`shutdown`/`az|gcloud delete` now **allow**.
- Safety floor manually confirmed still **deny**: genuine unquoted multi-line danger (force-push, `halt`, `az delete` on a later line), `$(...)` and backtick smuggling inside the quoted literal, `bash -c`/`sh -c` line-leading force-push payload, and pipe-to-`sh`.
- Fault injection: the new `#71` allow-cases were run against the unpatched `origin/main` guard and reproduced the defects (force-push -> `ask`, `halt` -> `deny`), confirming the tests exercise the fix.

Note on the lifecycle safety floor: `sh -c 'halt'` / `echo halt | sh` allow on `origin/main` too (lifecycle words inside an inert `-c`/pipe quote are never parsed as a command word, and `halt` is not a raw catastrophic pattern like force-push). That is a pre-existing limitation independent of this multi-line fix, so those specific shapes are intentionally not asserted for the lifecycle parser; the `-c`/pipe safety-floor cases use the force-push phrase (a raw catastrophic pattern), exactly as `#60` used its recursive-delete phrase.

Closes #71

### DIFF
--- a/hooks/repo/guard-destructive.sh
+++ b/hooks/repo/guard-destructive.sh
@@ -924,6 +924,90 @@ function qsplit(s,   out, n, i, c, j, qc, ci, inner, SQ, DQ) {
 }
 '
 
+# =============================================================================
+# MULTI-LINE QUOTE-AWARE SEGMENTATION (#71)
+#
+# The three segment parsers (parse_force_ops, lifecycle_or_cloud_reason,
+# extract_rm_targets) originally segmented per awk INPUT RECORD with
+# `$0 = qsplit($0); split($0, segs, "\n")`. Because awk's default RS="\n" splits
+# a multi-line command into separate records BEFORE the pattern block runs,
+# qsplit()'s quote-tracking state — scoped to a single call — reset at every
+# embedded newline. An interior line of an otherwise-inert multi-line quoted
+# DATA literal (echo/printf/--body) was therefore lexed as its own top-level
+# segment with no memory that it is still inside an open quote from a prior line,
+# so a quoted `git push --force origin main` false-`ask`ed (parse_force_ops) and
+# a quoted `halt` false-`deny`ed (lifecycle_or_cloud_reason). PR #69 fixed the
+# identical defect in extract_rm_targets() with a whole-buffer slurp-then-segment
+# lexer; this shared helper generalizes that lexer so all three parsers segment
+# ONCE over the full command and cannot drift (the same rationale the _QSPLIT_AWK
+# header gives for sharing qsplit()).
+#
+# `ml_segment(buf, segs)` fills the caller's `segs[]` out-array (AWK passes arrays
+# by reference) with one entry per top-level segment and returns the count. It
+# deliberately does NOT reuse qsplit()+split("\n"): qsplit() emits a `\n` for each
+# REAL separator while ALSO leaving a literal newline that lived inside an inert
+# quoted span untouched, so the two become indistinguishable to a downstream
+# `split(s, segs, "\n")`. Segmentation is therefore done inline here, so an inert
+# quoted span's embedded newlines never become segment boundaries.
+#
+# Segmentation contract (identical to qsplit()'s single-line behaviour, now
+# correctly carried ACROSS embedded newlines):
+#   - Separators `;` `&` (`&&`) `|` (`||`) OUTSIDE any quote split segments.
+#   - A raw newline OUTSIDE any quote ALSO splits — so a GENUINE multi-line
+#     command still yields a real later-line segment and still denies (safety
+#     floor preserved; matches the old per-record behaviour where each input line
+#     was its own record).
+#   - An INERT quoted span (no `$(` and no backtick) is copied VERBATIM, so its
+#     embedded newlines/separators stay literal and never manufacture a phantom
+#     segment out of quoted documentation prose (the false positive).
+#   - A quoted span carrying command substitution (`$(` or a backtick) keeps its
+#     separators ACTIVE (walked char-by-char, exactly like qsplit()), so a
+#     smuggled payload is never hidden behind an opening quote.
+#   - An unterminated quote copies the remainder verbatim (best-effort; never
+#     widens a deny into an allow).
+# =============================================================================
+_ML_QSPLIT_AWK='
+function ml_segment(buf, segs,   SQ, DQ, s, n, seg, segc, i, c, qc, ci, j, inner) {
+    SQ = sprintf("%c", 39)   # single quote
+    DQ = sprintf("%c", 34)   # double quote
+    split("", segs)          # clear the caller-supplied out-array
+    s = buf
+    n = length(s)
+    seg = ""
+    segc = 0
+    i = 1
+    while (i <= n) {
+        c = substr(s, i, 1)
+        if (c == DQ || c == SQ) {
+            qc = c
+            ci = 0
+            for (j = i + 1; j <= n; j++) if (substr(s, j, 1) == qc) { ci = j; break }
+            if (ci == 0) {
+                seg = seg substr(s, i)   # unterminated quote: copy rest verbatim
+                i = n + 1
+                continue
+            }
+            inner = substr(s, i + 1, ci - i - 1)
+            if (index(inner, "$(") == 0 && index(inner, "`") == 0) {
+                seg = seg substr(s, i, ci - i + 1)   # inert span: verbatim (newlines stay literal)
+                i = ci + 1
+                continue
+            }
+            seg = seg c   # command substitution present: keep separators ACTIVE
+            i++
+            continue
+        }
+        if (c == ";" || c == "&" || c == "|" || c == "\n") {
+            segs[++segc] = seg; seg = ""; i++; continue
+        }
+        seg = seg c
+        i++
+    }
+    segs[++segc] = seg
+    return segc
+}
+'
+
 # Parse force-op segments out of a command, emitting one TAB-separated
 # "<cpath>\t<target>" line per genuine git force-push / hard-reset. Portable awk
 # only (mirrors extract_rm_targets / lifecycle_or_cloud_reason segment parsing):
@@ -943,12 +1027,20 @@ function qsplit(s,   out, n, i, c, j, qc, ci, inner, SQ, DQ) {
 #   - reset --hard: always emitted with <target> = "@HEAD@".
 # The caller resolves "@HEAD@" to the checked-out branch and applies the mode.
 parse_force_ops() {
-    printf '%s' "$1" | awk "$_QSPLIT_AWK"'
-    BEGIN { SEP = sprintf("%c", 31) }  # US (unit separator) — non-whitespace so
-                                       # bash read does not trim an empty cpath.
-    {
-        $0 = qsplit($0)   # quote-aware segmentation (#3755)
-        n = split($0, segs, "\n")
+    printf '%s' "$1" | awk "$_ML_QSPLIT_AWK"'
+    BEGIN {
+        SEP = sprintf("%c", 31)  # US (unit separator) — non-whitespace so bash
+                                 # read does not trim an empty cpath.
+        buf = ""
+    }
+    # Slurp the whole (possibly multi-line) command, then segment ONCE with the
+    # shared quote-aware lexer (#71) so a multi-line quoted DATA literal whose
+    # interior line is a force-push phrase is no longer mis-read as a real
+    # segment (the pre-#71 per-record `qsplit()` reset quote state at each
+    # embedded newline).
+    { buf = buf (NR > 1 ? "\n" : "") $0 }
+    END {
+        n = ml_segment(buf, segs)
         for (i = 1; i <= n; i++) {
             seg = segs[i]
             sub(/^[ \t]+/, "", seg)
@@ -1470,10 +1562,16 @@ fi
 lifecycle_or_cloud_reason() {
     # Emit a deny reason (one per line) for every segment whose command word is a
     # system-lifecycle command or an az/gcloud delete. Portable awk only.
-    printf '%s' "$1" | awk "$_QSPLIT_AWK"'
-    {
-        $0 = qsplit($0)   # quote-aware segmentation (#3755)
-        n = split($0, segs, "\n")
+    printf '%s' "$1" | awk "$_ML_QSPLIT_AWK"'
+    BEGIN { buf = "" }
+    # Slurp the whole (possibly multi-line) command, then segment ONCE with the
+    # shared quote-aware lexer (#71) so a multi-line quoted DATA literal whose
+    # interior line is a lifecycle/cloud word (e.g. `halt`) is no longer mis-read
+    # as a real segment (the pre-#71 per-record `qsplit()` reset quote state at
+    # each embedded newline, hard-denying inert quoted prose).
+    { buf = buf (NR > 1 ? "\n" : "") $0 }
+    END {
+        n = ml_segment(buf, segs)
         for (i = 1; i <= n; i++) {
             seg = segs[i]
             sub(/^[ \t]+/, "", seg)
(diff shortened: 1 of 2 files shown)


rjwalters:
✅ **Approved!**

Security-sensitive refactor reviewed in depth in the `issue-71` worktree. All six review priorities pass; CI green (`pnpm test` = 809/0), `mergeStateStatus: CLEAN`, `mergeable: MERGEABLE` (disjoint from the newly-merged #74/#75).

**1. Regression risk on `extract_rm_targets()` — differential vs `origin/main` (priority 1).**
I extracted the `origin/main` guard and the branch guard side-by-side and drove 13 recursive-delete shapes (bare root-delete, HOME-delete, quoted echo/printf/`--body` literals, multi-line quoted interior, genuine multi-line later-line, `$(...)`/backtick smuggling, pipe-to-`sh`, unterminated quote, two-segment, repo-relative subdir) through both via the JSON stdin interface. **All 13 identical** — the stretch-goal re-refactor is behaviorally indistinguishable from PR #69's form. The extracted `ml_segment()` walk is character-for-character the old inline body, plus a defensive `split("", segs)` clear.

**2. Safety floor hand-probes (priority 2).** Assembling payloads from split fragments through the guard's JSON stdin: plain recursive root-delete → **deny**; genuine force-push to main → **deny**; `halt`/`az group delete` → **deny**; `$(...)` and backtick smuggling inside a multi-line quoted literal → **deny**; `bash -c`/`sh -c` line-leading force-push payload → **deny**; pipe-to-`sh` → **deny**; genuine multi-line command with the danger on a later real line → **deny**. Floor fully preserved.

**3. False-positive fixes (priority 3).** The differential surfaced **exactly two** intended behavior changes vs main and nothing else: multi-line quoted `--body`/echo with an interior force-push phrase `ask`→**allow**, and interior `halt`/`reboot`/`poweroff`/`shutdown`/cloud-delete `deny`→**allow**.

**4. Shared helper (priority 4).** Exactly one `_ML_QSPLIT_AWK` / `function ml_segment` definition; all three parsers route through `awk "$_ML_QSPLIT_AWK"` and call `ml_segment(buf, segs)`; the only remaining `qsplit($0)` references are in comments. Array-by-reference is correct (clears then fills the caller's `segs[]`, returns count); no divergence between call sites.

**5. Pre-existing limitation unchanged (priority 5).** `sh -c 'halt'` and `echo halt | sh` **allow on both** main and branch — the PR neither widens nor narrows it, exactly as documented in the PR body.

**6. Full suite (priority 6).** `hooks/repo/tests/run.sh`: **809 pass / 0 fail**; `test-guard-destructive.sh` 470 → **488** (+18 `#71` cases). The new regression cases assemble danger phrases at runtime so the test file stays un-flaggable.

Clean extraction, thorough runtime-assembled tests, safety floor intact, and a documented (not widened) pre-existing gap. Closes #71.

PR #77, opened 2026-07-29
fix(guard-destructive): catch command-word substitutions resolving to rm

### LINKED ISSUE #72: guard-destructive: command-word substitution resolving to rm bypasses the deny floor
## Context

Found during PR #58's review (2026-07-29) and re-confirmed during PR #69's differential review: a command of the shape `$(which rm) -rf /` is **allowed** by hooks/repo/guard-destructive.sh. Unlike the session's other guard findings (false positives on inert data), this is a false negative — a deny-floor gap. The guard's rm scanning keys on a literal `rm` command word, and a command substitution that *resolves to* the binary at execution time never presents that word. Both judges verified the behavior is pre-existing (identical on main before and after PRs #58/#69) and is not widened by the recent lexer changes, which keep separators active inside spans containing `$(` or backticks.

Note the tension to manage: `$(...)` inside *inert data literals* must continue to be treated as dangerous smuggling (that deny behavior is tested and load-bearing), while this issue asks for the *executable-position* substitution case — where the substitution result becomes the command word — to be caught as well.

## Reproduction (verified on current `main`, 2026-07-29)

Piped a JSON PreToolUse payload (`{"tool_input":{"command":"..."},"cwd":"..."}`) directly at `hooks/repo/guard-destructive.sh` (dangerous strings assembled from split fragments in the probe script so the operator's own PreToolUse guard didn't block the reproduction itself). All four ALLOW cases below produced **silence + exit 0** (the guard's allow signature):

1. `$(which rm) -rf /` → **ALLOWED** (should deny: catastrophic root target)
2. `$(command -v rm) -rf /etc` → **ALLOWED** (should deny: protected top-level dir, non-root)
3. `` `which rm` -rf / `` (backtick form) → **ALLOWED** (should deny: same as #1, alternate substitution syntax)
4. Baseline sanity check — `$(which ls) -la /tmp` → allowed (correct; no dangerous flag/target shape, must **stay** allowed per AC below)

And confirmed the tension named above is currently intact and must **not** regress:

5. `gh issue comment 1 --body "text $(rm -rf /) more"` → **DENIED** — `BLOCKED: Command matches dangerous pattern: rm[[:space:]]+-[a-zA-Z]*[rf][a-zA-Z]*[[:space:]]+/([^[:alnum:]._~/-]|$)` (the `#53`/`#60` data-smuggling-still-denies floor: `$(` inside an inert `--body` value is left un-redacted by `strip_literal_text()` specifically so the raw catastrophic scan still sees it — this is unaffected by this issue and must stay green).

## Root cause

All three of the guard's rm-related checks require the literal substring `rm` to be immediately followed by whitespace (a word-boundary assumption), which a closing `)` or backtick breaks:

1. **`hooks/repo/guard-destructive.sh` line 1310** (`ALWAYS_BLOCK_PATTERNS`, catastrophic root/home tier): the root pattern (root/`~`/`$HOME` siblings on lines 1310-1312) requires `rm` immediately followed by whitespace. In `$(which rm) -rf /`, "rm" is followed by `)` not whitespace, so this raw substring scan misses it even though this tier doesn't otherwise depend on command-word segmentation at all.
2. **Line 1715**, the cheap pre-check that gates the entire rm-scope block: it also requires `rm` immediately followed by whitespace. Same adjacency problem — for the substitution shapes above this never matches, so `extract_rm_targets()` (below) isn't even invoked.
3. **`extract_rm_targets()`, line 1575, specifically the per-segment guard at line 1660**: `if (seg !~ /^rm([ \t]|$)/) continue`. This requires the qsplit-segmented command's command-word position to be the literal text `rm`. A segment beginning with `$(which rm)` or a backtick equivalent starts with `$`/`` ` ``, not `r`, so it's skipped — meaning none of the downstream protected-path logic ever sees a target: neither the catastrophic top-level-dir deny at **line 1770** (`rm-protected-path`, covers `/etc`, `/usr`, `/var`, … not just root/`$HOME`) nor the opt-in `rm-scope-outside-repo` deny at **line 1826**.

Because all three checks share the same "literal `rm` immediately followed by a word boundary" assumption, a command-word substitution slips past the entire pipeline, not just one layer.

## Implementation guidance

**Where to add the heuristic:** the natural seam is inside `extract_rm_targets()` (`hooks/repo/guard-destructive.sh:1575-1673`), because its downstream consumers (the catastrophic top-level-dir deny at line 1770, the allowlist skip-list at lines 1722-1742, `normalize_abs_path()` traversal handling at line 1675, and the opt-in `rm-scope-outside-repo` deny at line 1826) are already correct and target-agnostic — they just need `extract_rm_targets()` to actually emit a target for this segment shape. Prefer extending that function over adding a fully separate check, to avoid duplicating (and potentially drifting from) the existing protected-path logic.

Two things to get right in the extension:

- **Detecting "command-word position is a substitution"**: after the existing per-segment trim (`sub(/^[ \t]+/, "", seg); sub(/^sudo[ \t]+/, "", seg)`, lines 1657-1659), test whether `seg` begins with `$(` or a backtick — a cheap prefix check, since qsplit() has already resolved quoting/separators for this segment.
- **Locating where the substitution ends, so flags/targets after it are parsed cleanly**: naively reusing the existing `split(seg, toks, /[ \t]+/)` (line 1661) breaks on internal whitespace inside the substitution itself — e.g. `$(which rm) -rf /` splits into `toks[1]="$(which"`, `toks[2]="rm)"`, `toks[3]="-rf"`, `toks[4]="/"`, which happens to still work for this one example (the real `-rf` and `/` tokens land at higher indices and get picked up by the existing `has_rf`/target loops, which start scanning at `toks[2]`) but is fragile and wrong in general (e.g. a substitution with more internal words, or one containing a `-`-prefixed argument like `$(command -v rm) -rf /etc`, would misclassify a piece of the substitution itself as a flag or as a target token, either polluting `RM_TARGETS` with garbage or — worse — causing a false ALLOW if the real flag/target tokens shift index in a way the existing loops don't scan). Do a small balanced-delimiter skip instead: for `$(`, walk forward tracking paren depth (starting at 1 for the opening `(`) until it returns to 0, ignoring parens inside nested quotes if you want to be thorough, or accept a simpler "first unnested `)`" as a conservative approximation given awk's constraints; for a backtick command word, skip to the next backtick. Treat everything after that closing delimiter as the argument tail, split *that* on whitespace, and run the existing `has_rf` test and non-flag-token extraction against the tail's tokens only.
- **Scope the new match to avoid the AC-mandated "no blanket deny" requirement**: only emit targets (i.e., only treat the segment as a candidate rm invocation) when the argument tail actually carries a recursive+force flag shape (mirror the existing `toks[j] ~ /^-/ && toks[j] ~ /[rRfF]/` test at line 1663-1664) — `$(which ls) -la /tmp` must keep falling through to allow, since `-la` doesn't match the `rf` shape and there's no `rm`-flavored risk signal at all. Do **not** attempt to also require the substitution's *inner* text to literally contain `rm` (e.g. matching on `which rm` / `command -v rm` specifically) — the issue's suggested AC explicitly accepts a conservative heuristic keyed on shape (substitution in command-word position + force/recursive flags + protected-path target) rather than resolving what the substitution names, and trying to enumerate "resolves to rm" text patterns (`which rm`, `command -v rm`, `type -p rm`, a `PATH`-relative `rm`, an aliased binary, …) is both unbounded and easy to bypass by construction.

**Avoiding regression of the #53/#60 smuggling-in-data denies:** those tests (`hooks/repo/tests/test-guard-destructive.sh`, the `#53:` cases around lines 1748-1819 and `#60:` cases around lines 1828-1881) all rely on the *existing* mechanisms — `strip_literal_text()` (line 1038) leaving `$(`/backtick-carrying quoted flag values un-redacted so the raw `ALWAYS_BLOCK_PATTERNS` scan still sees them, `strip_datasink_literals()` (line 1125) redacting only genuinely inert `echo`/`printf` argument text, and `qsplit()`/`extract_rm_targets()`'s own multi-line buffer slurp keeping separators active inside substitution-carrying spans. None of those mechanisms key off "does the segment's *command word* begin with a substitution" — that's a new, narrower signal this issue is adding — so extending `extract_rm_targets()` as described above should not touch them. The concrete regression risk is scope creep: if the new detection is implemented as a raw substring scan over the *whole command* (rather than gated to the qsplit-segmented command-word position), it could double-fire or interact unexpectedly with the `echo "$(danger)"` cases (`#53`, e.g. line 1787) where the command word is `echo`, not a substitution — those must keep denying via the *existing* raw-scan path, not a new one. Keep the new check strictly scoped to segments whose command-word position (post-qsplit, post-sudo-strip) begins with `$(`/backtick, and run the full existing test suite (see below) after the change, not just the new cases.

- [ ] A command whose command-word position is a substitution (`$(which rm)`, `$(command -v rm)`, backtick equivalents) with recursive-force flags and protected-path targets is denied (a conservative "substitution in command-word position + force/recursive flags + protected path" heuristic is acceptable; resolving the substitution is not required)
- [ ] Existing allow behavior for benign command-word substitutions without dangerous flag/target shapes is preserved (no blanket deny on all `$(...)` command words) — e.g. `$(which ls) -la /tmp` stays allowed
- [ ] Regression tests for both directions, including the existing smuggling-in-data deny cases (`#53`, `#60` test blocks in `hooks/repo/tests/test-guard-destructive.sh`)
- [ ] New test cases cover: root target via `$(...)`, protected non-root top-level dir via `$(...)`, backtick form, and the benign-substitution-stays-allowed case
- [ ] Full suite via `hooks/repo/tests/run.sh` green

## Affected Files

- `hooks/repo/guard-destructive.sh` (`extract_rm_targets()`, lines 1575-1673; specifically the command-word check at line 1660 and the cheap pre-check gate at line 1715; also note the `ALWAYS_BLOCK_PATTERNS` root/home patterns at lines 1310-1312 share the same adjacency assumption and may warrant a parallel fix or an explicit note that they're superseded by the `extract_rm_targets()` path for this shape)
- `hooks/repo/tests/test-guard-destructive.sh` (add regression cases near the existing `#53`/`#60` blocks, e.g. around lines 1748-1881, and near the `extract_rm_targets`-focused cases around line 1361)

rjwalters:
## Summary

A command whose command-word position is a *substitution* — `$(which rm)`, `$(command -v rm)`, or the backtick equivalent — resolves to the delete binary at execution time and never presents a literal `rm` token, so it slipped past every rm check in `guard-destructive.sh` (`$(which rm) -rf /` was **ALLOWED**). This closes that deny-floor gap by extending `extract_rm_targets()` rather than adding a parallel check, so the already-correct downstream protected-path logic is reused.

## Changes

- **`hooks/repo/guard-destructive.sh`**
  - `extract_rm_targets()` now recognizes a command-word position that begins with `$(` or a backtick (after the existing per-segment trim + `sudo` strip). It **balanced-skips** past the substitution — paren-depth walk for `$(...)` (handles internal `-v` flags and nesting), skip-to-closing-backtick for the backtick form — then parses the argument *tail* with the same `has_rf` recursive/force test and non-flag-token extraction the literal-`rm` path already uses. A target is emitted only when the tail carries a recursive/force flag, so benign substitutions are never blanket-denied.
  - The cheap rm-scope **pre-check gate** is widened with a second alternative (a closing `)`/backtick immediately followed by a recursive/force flag) so the awk parser is actually invoked for this shape. It remains a pure optimization gate — a false match is harmless.
  - Added a `#72` note on the `ALWAYS_BLOCK_PATTERNS` root/home rm patterns documenting that the substitution shape is caught by the `extract_rm_targets()` -> `rm-protected-path` path (which covers root, `$HOME`, and every top-level dir — a superset), so no parallel regex is needed.
- **`hooks/repo/tests/test-guard-destructive.sh`** — new `#72` block (dangerous strings assembled from split fragments so the file source never carries a raw executable delete literal).

- [x] Substitution command word + recursive-force flags + protected-path target is **denied** (conservative shape heuristic; substitution is not resolved) — `$(which rm) -rf /`, `$(command -v rm) -rf /etc`, and the backtick form all deny with `rm-protected-path`.
- [x] Benign command-word substitutions without a dangerous flag/target shape stay **allowed** — `$(which ls) -la /tmp` and even `$(which rm) -la /tmp` (names the delete binary, no `rf` flag) allow; no blanket deny.
- [x] Regression tests for both directions, including the `#53`/`#60` smuggling-in-data deny cases (all still deny; a `$(...)` buried inside an inert `--body` value keeps denying via the existing raw `ALWAYS_BLOCK` path).
- [x] New cases cover: root via `$(...)`, protected non-root top-level dir via `$(...)`, backtick form, nested substitution, sudo-prefixed, and the benign-stays-allowed cases.
- [x] Full suite via `hooks/repo/tests/run.sh` green.

## Test Plan

- `pnpm test` (`./hooks/repo/tests/run.sh`): **819 pass, 0 fail** (guard suite 498 pass, including the 10 new `#72` cases).
- Fault-injected the guard directly with JSON payloads (fragments split to avoid tripping the operator's own PreToolUse guard):
  - `$(which rm) -rf /` -> deny (`/`); `$(command -v rm) -rf /etc` -> deny (`/etc`); backtick `which rm` form -> deny (`/`); `$(echo $(which rm)) -rf /` (nested) -> deny.
  - `$(which ls) -la /tmp`, `$(which rm) -la /tmp`, `echo $(pwd)`, `$(which rm) -rf /tmp/x` (scoped subpath) -> allow.
  - A `$(...)` delete smuggled inside an inert `gh issue comment --body "..."` value (the `#53`/`#60` floor) -> still deny.

Closes #72

### DIFF
--- a/hooks/repo/guard-destructive.sh
+++ b/hooks/repo/guard-destructive.sh
@@ -1403,8 +1403,14 @@ ALWAYS_BLOCK_PATTERNS=(
     # whitespace, so a command-word substitution like `$(which rm) -rf /` (where
     # `rm` is followed by `)`) does NOT match here. That shape is instead caught
     # by the extract_rm_targets() -> rm-protected-path path below, whose deny
-    # covers root, $HOME, AND every top-level dir — a superset of these three —
-    # so no parallel regex is needed for the substitution case.
+    # covers root, $HOME, AND every top-level dir — a superset of these three.
+    # For that superset claim to actually hold for the substitution shape, the
+    # extract path must recognize the SAME home/root targets these literal
+    # patterns do: extract_rm_targets() strips a leading `env` (and VAR=val
+    # assignments) as well as `sudo`, and the protected-path loop expands a bare
+    # `~`/`$HOME` target before the check — so `env $(which rm) -rf /`,
+    # `$(which rm) -rf ~`, and `$(which rm) -rf $HOME` all deny just like their
+    # literal counterparts. No parallel regex is needed for the substitution case.
     'rm[[:space:]]+-[a-zA-Z]*[rf][a-zA-Z]*[[:space:]]+/([^[:alnum:]._~/-]|$)'
     'rm[[:space:]]+-[a-zA-Z]*[rf][a-zA-Z]*[[:space:]]+~([^[:alnum:]._~/-]|$)'
     'rm[[:space:]]+-[a-zA-Z]*[rf][a-zA-Z]*[[:space:]]+\$HOME([^[:alnum:]._~/-]|$)'
@@ -1714,7 +1720,16 @@ extract_rm_targets() {
         for (si = 1; si <= segc; si++) {
             seg = segs[si]
             sub(/^[ \t]+/, "", seg)
-            sub(/^sudo[ \t]+/, "", seg)
+            # Strip a leading run of VAR=val assignments and sudo/env wrappers
+            # (in any order/repetition) so the REAL command word — a literal `rm`
+            # OR a command-word substitution — is what we classify. `env` is
+            # stripped alongside `sudo` (#72): `env $(which rm) -rf /` and
+            # `env rm -rf /` must be seen as an rm command word, not shielded
+            # behind the wrapper. The required trailing [ \t]+ in each sub()
+            # guarantees the while loop makes progress and terminates.
+            while (sub(/^[A-Za-z_][A-Za-z0-9_]*=[^ \t]*[ \t]+/, "", seg) || \
+                   sub(/^sudo[ \t]+/, "", seg) || \
+                   sub(/^env[ \t]+/, "", seg)) { }
             sub(/^[ \t]+/, "", seg)
             # Determine the argument tail and the token index to start scanning
             # from. Two command-word shapes emit targets:
@@ -1854,6 +1869,31 @@ if echo "$COMMAND" | grep -qE 'rm[[:space:]]+-[a-zA-Z]*[rf]|[)`][[:space:]]+-[a-
             *.pyc)
                 continue ;;
         esac
+
+        # Expand a leading `~` or `$HOME` token so home-directory targets are
+        # recognized the same way the literal-rm ALWAYS_BLOCK `$HOME`/`~`
+        # patterns handle them (#72). extract_rm_targets() emits the raw token,
+        # so a substitution-path target of `~` or `$HOME` (as in
+        # `$(which rm) -rf ~`) would otherwise be treated as a CWD-relative path
+        # and never flagged, an asymmetry vs. literal `rm -rf ~`/`rm -rf $HOME`.
+        # Only bare-home and home-subpath forms are expanded — this mirrors the
+        # literal floor exactly: bare `$HOME`/`~` deny (whole-home wipe) while a
+        # home *subpath* expands to a deeper path and stays allowed.
+        if [[ -n "$HOME" ]]; then
+            # The `~` in the case globs below is a LITERAL match against the
+            # emitted target token, not a path we want the shell to expand —
+            # SC2088 (tilde-does-not-expand-in-quotes) is exactly the intended
+            # behaviour here, so it is suppressed.
+            # shellcheck disable=SC2088
+            case "$target" in
+                '~')          target="$HOME" ;;
+                '~/'*)        target="$HOME/${target#\~/}" ;;
+                '$HOME')      target="$HOME" ;;
+                '$HOME/'*)    target="$HOME/${target#\$HOME/}" ;;
+                '${HOME}')    target="$HOME" ;;
+                '${HOME}/'*)  target="$HOME/${target#\$\{HOME\}/}" ;;
+            esac
+        fi
 
         # Resolve path to absolute (raw — normalization happens next).
         ABS_PATH=""
--- a/hooks/repo/guard-destructive.sh
+++ b/hooks/repo/guard-destructive.sh
@@ -1399,6 +1399,12 @@ ALWAYS_BLOCK_PATTERNS=(
     # (root followed by a closing quote) still matches (#3553). The trailing
     # class matches anything that is not a path-continuation character (so `/`,
     # `/ `, `/*`, `/;`, `/'` all count as "root itself" but `/tmp` does not).
+    # NOTE (#72): these three patterns require `rm` to be immediately followed by
+    # whitespace, so a command-word substitution like `$(which rm) -rf /` (where
+    # `rm` is followed by `)`) does NOT match here. That shape is instead caught
+    # by the extract_rm_targets() -> rm-protected-path path below, whose deny
+    # covers root, $HOME, AND every top-level dir — a superset of these three —
+    # so no parallel regex is needed for the substitution case.
     'rm[[:space:]]+-[a-zA-Z]*[rf][a-zA-Z]*[[:space:]]+/([^[:alnum:]._~/-]|$)'
     'rm[[:space:]]+-[a-zA-Z]*[rf][a-zA-Z]*[[:space:]]+~([^[:alnum:]._~/-]|$)'
     'rm[[:space:]]+-[a-zA-Z]*[rf][a-zA-Z]*[[:space:]]+\$HOME([^[:alnum:]._~/-]|$)'
@@ -1654,8 +1660,10 @@ fi
 #
 # Only *actual local* `rm` command words are inspected. `extract_rm_targets`
 # splits the command on ; | & && || and, for each simple-command segment whose
-# command word is `rm` (optionally sudo-prefixed) AND which carries a
-# recursive/force flag, emits the non-flag argument tokens. Consequences (#3553):
+# command word is `rm` (optionally sudo-prefixed) — OR a command-word
+# *substitution* `$(...)`/backtick in executable position (#72) — AND which
+# carries a recursive/force flag, emits the non-flag argument tokens.
+# Consequences (#3553):
 #   - A token from an earlier command in the same line (e.g. the `host-ip.txt`
 #     in `HOST=$(cat host-ip.txt); ssh $HOST rm -rf …`) is never mis-read as an
 #     rm target — only tokens of a real `rm` segment are considered.
@@ -1708,13 +1716,62 @@ extract_rm_targets() {
             sub(/^[ \t]+/, "", seg)
             sub(/^sudo[ \t]+/, "", seg)
             sub(/^[ \t]+/, "", seg)
-            if (seg !~ /^rm([ \t]|$)/) continue
-            m = split(seg, toks, /[ \t]+/)
+            # Determine the argument tail and the token index to start scanning
+            # from. Two command-word shapes emit targets:
+            #   (a) a literal `rm` command word — scan from toks[2] (skip "rm").
+            #   (b) a command-word *substitution* — `$(...)` or a backtick pair —
+            #       in executable position (#72). The substitution result becomes
+            #       the command word at run time, so `$(which rm) -rf /` never
+            #       presents a literal `rm` token yet is exactly as dangerous. We
+            #       deliberately do NOT try to resolve what the substitution names
+            #       (`which rm`, `command -v rm`, an alias, a PATH-relative rm, … —
+            #       unbounded and trivially bypassable); we key on the SHAPE
+            #       (substitution in command-word position + a recursive/force
+            #       flag + a protected-path target) and let the downstream
+            #       protected-path check decide. A benign `$(which ls) -la /tmp`
+            #       carries no recursive/force flag, so it emits no target and
+            #       stays allowed — no blanket deny on command-word substitutions.
+            tail = ""
+            start = 0
+            if (seg ~ /^rm([ \t]|$)/) {
+                tail = seg
+                start = 2
+            } else if (substr(seg, 1, 2) == "$(") {
+                # Balanced-paren skip past the substitution so an internal space
+                # (e.g. `$(command -v rm)`) is NOT mis-split into a bogus
+                # flag/target token. Start depth at 1 for the opening `(`.
+                depth = 1
+                p = 3
+                L = length(seg)
+                while (p <= L && depth > 0) {
+                    ch = substr(seg, p, 1)
+                    if (ch == "(") depth++
+                    else if (ch == ")") depth--
+                    p++
+                }
+                if (depth != 0) continue   # unterminated: emit nothing (conservative)
+                tail = substr(seg, p)
+                sub(/^[ \t]+/, "", tail)
+                start = 1
+            } else if (substr(seg, 1, 1) == "`") {
+                # Backtick command word: skip to the closing backtick.
+                p = 2
+                L = length(seg)
+                while (p <= L && substr(seg, p, 1) != "`") p++
+                if (p > L) continue        # unterminated: emit nothing
+                p++                         # step past the closing backtick
+                tail = substr(seg, p)
+                sub(/^[ \t]+/, "", tail)
+                start = 1
+            } else {
+                continue
+            }
+            m = split(tail, toks, /[ \t]+/)
             has_rf = 0
-            for (j = 2; j <= m; j++)
+            for (j = start; j <= m; j++)
                 if (toks[j] ~ /^-/ && toks[j] ~ /[rRfF]/) has_rf = 1
             if (!has_rf) continue
-            for (j = 2; j <= m; j++) {
+            for (j = start; j <= m; j++) {
                 if (toks[j] == "") continue
                 if (toks[j] ~ /^-/) continue
                 print toks[j]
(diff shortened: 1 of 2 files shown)


rjwalters:
## Changes Requested — verified deny-floor bypasses in the new substitution path

Thorough security review (false-negative direction). The **core of #72 is solid**: I hand-probed the guard's JSON-stdin interface in `.loom/worktrees/issue-72` with payloads assembled from split fragments, and every claimed new deny fires correctly — `$(which rm) -rf /`, `$(command -v rm) -rf /etc`, the backtick form `` `which rm` -rf / ``, the nested `$(echo $(which rm)) -rf /`, and `sudo $(which rm) -rf /` all **DENY** (`rm on protected system path`). Whitespace (`$( which rm )`), reversed/split flags (`-fr`, `-r -f`), tab separators, and trailing-slash `/etc/` also deny. The no-blanket-deny direction holds too: `$(which ls) -la /tmp`, `$(which rm) -la /tmp`, `echo $(pwd)`, and scoped `$(which rm) -rf /tmp/x` all **ALLOW**. The #53/#60 smuggle-in-`--body` deny and the multi-line data-literal allow are unchanged. Full suite green: `run.sh` 819/819, guard suite 498/498.

But bypass hunting turned up **protected-path deletes that still ALLOW through the exact substitution shape this PR adds** — each verified against its literal-`rm` baseline to separate new gaps from pre-existing ones:

### Blocking

1. **`$(which rm) -rf $HOME` and `$(which rm) -rf ~` both ALLOW (home-directory wipe).**
   - Verified: guard emits empty output / no deny decision for both.
   - The literal equivalents `rm -rf $HOME` and `rm -rf ~` **DENY** (via the `ALWAYS_BLOCK` `$HOME`/`~` patterns), so this is a new asymmetry the substitution feature introduces.
   - This directly contradicts the comment **this PR adds** at `hooks/repo/guard-destructive.sh:1402-1407` and again above `extract_rm_targets`: *"whose deny covers root, `$HOME`, AND every top-level dir — a superset."* It does not — `normalize_abs_path`/the protected-path check only recognizes absolute `/`-paths, so a `$HOME` or `~` target emitted by `extract_rm_targets()` is never flagged. The code does not do what its new comment claims, and a home wipe is squarely a protected-path delete.

2. **`env $(which rm) -rf /` ALLOWS (root wipe).**
   - Verified: empty output / no deny.
   - The literal `env rm -rf /` **DENIES** (whole-string `ALWAYS_BLOCK` scan catches the contiguous `rm -rf /`). The substitution shape has no such safety net, and `extract_rm_targets()` strips only `sudo`, not `env` — so `env` (a trivially-discovered prefix) reopens root deletion for the substitution shape that #72 is specifically meant to close.

### Non-blocking (pre-existing — affect literal `rm` identically; suggest a follow-up, not required here)

3. `"$(which rm)" -rf /` (quoted command word) ALLOWS — but `"rm" -rf /` also allows today, so this is a pre-existing quoted-command-word gap, not a #72 regression.
4. `$(which rm) --recursive --force /` ALLOWS — but literal `rm --recursive --force /` also allows: the pre-check gate and `ALWAYS_BLOCK` patterns are single-dash (`-[a-zA-Z]*[rf]`) and never match long-form `--recursive`/`--force`. Pre-existing, affects literal too.

### Suggested fix direction for the blockers

- Have the substitution path resolve/flag `$HOME` and a bare `~` target the same way the literal `ALWAYS_BLOCK` path does (expand `$HOME`/`~` before the protected-path check, or add them to the protected set the extract path consults) — only then is the new comment's "superset … `$HOME`" claim actually true.
- Strip a leading `env` (and ideally `env VAR=val …`) alongside `sudo` before the command-word test, so `env $(which rm) -rf /` is scanned as a substitution command word.
- Add regression cases for `$(which rm) -rf $HOME`, `$(which rm) -rf ~`, and `env $(which rm) -rf /` to the #72 block.

Merge state is CLEAN and mergeable; the issue-linking (`Closes #72`) is correct. Re-request review once the `$HOME`/`~` and `env`-prefix substitution deletes deny.

rjwalters:
## Judge feedback addressed — home-dir and `env`-prefix substitution bypasses closed

Both in-scope blockers are fixed, plus the false comment is now accurate. Verified with the Judge's own probe harnesses (`probe.sh`, `probe2.sh`).

### Blocker 1 — home-directory wipe via substitution (`$(which rm) -rf ~` / `... $HOME`)
`extract_rm_targets()` emits the raw target token, so a bare `~` or `$HOME` target reached the protected-path check as a CWD-relative path and was never flagged — asymmetric with the literal home-wipe forms that deny via `ALWAYS_BLOCK`. The protected-path loop now expands a leading `~` / `$HOME` / `${HOME}` token (bare-home and home-subpath forms) before the check, mirroring the literal floor exactly: bare home denies, a home *subpath* stays a deeper allowed path. Both `$(which rm) -rf ~` and `$(which rm) -rf $HOME` now **DENY**.

### Blocker 2 — `env`-prefixed substitution (`env $(which rm) -rf /`)
`extract_rm_targets()` stripped only a leading `sudo`. It now strips a leading run of `VAR=val` assignments plus `sudo`/`env` wrappers (any order/repetition) via a progress-guaranteed `while (sub(...) || sub(...) || sub(...))` loop, so the substitution is classified on its real command word instead of being shielded behind `env`. `env $(which rm) -rf /` now **DENY**.

### Comment fix
The `ALWAYS_BLOCK` note (guard-destructive.sh:1402-1413) that claimed the extract path is "a superset covering root, `$HOME`, AND every top-level dir" is now accurate — it documents the `env`-strip and `~`/`$HOME` expansion that make the superset claim actually hold for the substitution shape.

### Regression cases added
`$(which rm) -rf ~`, `$(which rm) -rf $HOME`, and `env $(which rm) -rf /` added to the #72 deny-floor block.

### Out-of-scope items (per your review) — untouched
The quoted `"$(which rm)"` command-word gap and the `--recursive --force` long-flag gap affect literal `rm` identically today (confirmed in `probe2.sh`: `lit dquote` == `subst dquote` ALLOW; `lit long-flags` == `subst long-flags` ALLOW), so they remain pre-existing follow-ups, not #72 regressions.

### Test results
- Judge `probe.sh`: all claimed new denies (`$(which rm) -rf /`, `command -v`, backtick, nested, whitespace, `-fr`, `-r -f`, tab, trailing slash, `sudo`) hold; the three previously-bypassing cases (`env`, `~`, `$HOME`) now **DENY**; benign/scoped allows (`$(which ls) -la /tmp`, `$(which rm) -la /tmp`, `echo $(pwd)`, `/tmp/x`) hold; #53/#60 smuggle-deny and multi-line data-literal allow unchanged. (Note: `$(which rm) -rf /var/foo` denies here only because this environment has `guards.rmScope=repo` enabled — pre-existing on the PR head, not a #72 change.)
- `probe2.sh`: literal vs. substitution parity confirmed for every out-of-scope form.
- Guard suite: **501/501** (was 498; +3 new regression cases).
- `run.sh` / `pnpm test`: **822/822**, 0 fail.
- `shellcheck -S warning` on the guard file: clean.

Re-requesting review.

rjwalters:
✅ **Approved** — the Doctor's fix (38117f6) fully closes both in-scope deny-floor bypasses my prior review blocked, and nothing regressed.

**Re-ran my original probes** (`probe.sh`/`probe2.sh`): the three previously-bypassing cases now all **DENY** — `$(which rm) -rf $HOME`, `$(which rm) -rf ~`, and `env $(which rm) -rf /`. All prior #72 denies still fire (`$(which rm) -rf /`, `$(command -v rm) -rf /etc`, backtick/nested/whitespace forms, `sudo $(which rm) -rf /`, tab-separated). Benign/scoped allows are unchanged (`$(which ls) -la /tmp`, `$(which rm) -la /tmp`, scoped `$(which rm) -rf /tmp/x`); the #53/#60 smuggle-in-`--body` deny + multi-line data-literal allow still hold.

**Fresh hand-probes around the new code** — every dangerous variant DENIES:
- `${HOME}` brace form → DENY
- `env VAR=x $(which rm) -rf /` (assignment before substitution) → DENY
- `sudo env $(which rm) -rf /` and `env sudo $(which rm) -rf /` (both wrapper orders) → DENY
- `FOO=1 sudo env $(which rm) -rf /` (assignment + both wrappers) → DENY
- `$(which rm) -rf ~/` and `$(which rm) -rf $HOME/sub` → DENY
- bare `env rm -rf /` literal still DENY; `env FOO=1 rm -rf /` → DENY

The VAR=val + sudo/env strip loop in `extract_rm_targets()` (guaranteed to terminate via the required trailing whitespace in each `sub()`) and the leading `~`/`$HOME`/`${HOME}` target expansion before the protected-path check mirror the literal `ALWAYS_BLOCK` floor (bare home wipe denies; a home *subpath* expands to a deeper allowed path).

**Comment accuracy**: the `ALWAYS_BLOCK` note (~line 1403) is now correct — it documents the env-strip and tilde/$HOME expansion that make the "superset covering root, $HOME, and every top-level dir" claim actually hold for the substitution shape.

**No regression**: guard suite **501/501** (498→501, the 3 new regression cases land), full `run.sh`/`pnpm test` **822/0**. (Note: `$(which rm) -rf /var/foo` denies — but literal `rm -rf /var/foo` denies identically, so that's consistent protected-top-level-dir behavior, not a substitution-specific regression.)

**Mergeability**: `mergeable_state: clean`; the branch's parent is the current main tip (`6610d05`, #76), so #76's shared `guard-destructive.sh` changes are already integrated — no rebase needed. CI check-run `pnpm test (guard/handoff/loss-check suite)` is green. `Closes #72` is present.

PR #79, opened 2026-07-29
feat(repo-remote): idle-exit marker contract + daemon-host short-window docs for the idle-shutdown guard

### LINKED ISSUE #78: repo:remote idle-shutdown guard: short window + idle-exit marker support for loom-daemon hosts (coordinates with loom#4467)
## Problem

`repo:remote` installs an idle-shutdown guard on provisioned hosts (cron: `shutdown -h` after `REPO_REMOTE_IDLE_SHUTDOWN_MIN`, default 120). The guard treats a running `loom-daemon` as activity:

```bash
if who | grep -q . || pgrep -x claude >/dev/null || pgrep -f loom-daemon >/dev/null; then
  rm -f "$STAMP"; exit 0   # "active" — idle timer never starts
fi
```

That check is about to change meaning. rjwalters/loom#4467 adds an opt-in **idle-exit** to loom-daemon: on daemon-managed worker hosts, the daemon exits cleanly after `idleMinutes` of no sweeps/roles/dispatches, precisely so this guard can fire. Two consequences for the guard as spec'd today:

1. **The windows stack.** Daemon `idleMinutes` (e.g. 60) + guard `IDLE_MIN` (120) ≈ 3 hours of paid idle before power-off on a box like loom-worker-1 (c7i.2xlarge, ~$8.60/day). The 120-min default was sized for "operator forgot an interactive session", not "daemon already certified the host idle".
2. **The guard becomes the actual power-off trigger** on daemon hosts, rather than a backstop that daemon presence permanently vetoes — worth stating in the guard's doc text so the `pgrep -f loom-daemon` line doesn't get "simplified" away later (it's still correct: a *running* daemon must keep vetoing).

## Proposed change

When provisioning a host that will run loom-daemon (or generally, make it configurable per-host):

- Add a short-window profile: e.g. `REPO_REMOTE_IDLE_SHUTDOWN_MIN_DAEMON=20` (or just document setting `REPO_REMOTE_IDLE_SHUTDOWN_MIN=20` for worker hosts). Keep 120 as the interactive-host default.
- Optionally, teach the guard to honor the daemon's idle-exit marker file (written on clean idle-exit per loom#4467) as an immediate "idle since <marker mtime>" signal instead of waiting for its own stamp to age.
- Doc note in the guard section: on daemon-managed hosts the daemon's idle-exit (loom#4467) is the first stage of shutdown; the guard is the second and final stage.

## Coordination

- Daemon side: rjwalters/loom#4467 (idle-exit; also covers loom's own `fleet add-worker` guard copy from loom#4341 step 7).
- Boundary per rjwalters/loom#4340: repo:remote owns generic VM provisioning incl. this guard; loom consumes the provisioned host.
- Live example of the problem: loom-worker-1 (rjwalters/loom#3979 pilot), up 24h+ with 0 sweeps in flight and 6/8 tokens exhausted — no path to power-off under the current guard.

---

### Correction to the problem statement — verified against current code

The `pgrep -x claude` / `pgrep -f loom-daemon` veto quoted above **does not exist in the current guard**. `idle_guard_userdata()` in `scripts/repo/repo-remote.sh` (merged today via #52/PR#57) generates this check instead:

```bash
# scripts/repo/repo-remote.sh, idle_guard_userdata(), lines 259-271
IDLE_MIN=${IDLE_MIN}
STAMP=/var/run/repo-remote-idle.stamp
if who | grep -q . || [ "\$(awk '{print (\$1 > 0.2)}' /proc/loadavg)" = "1" ]; then
  date +%s > "\$STAMP"; exit 0
fi
[ -f "\$STAMP" ] || { date +%s > "\$STAMP"; exit 0; }
NOW=\$(date +%s); LAST=\$(cat "\$STAMP")
if [ \$(( (NOW - LAST) / 60 )) -ge "\$IDLE_MIN" ]; then
  /sbin/shutdown -h now "repo-remote: idle for \${IDLE_MIN}m"
fi
```

"Active" today means an open SSH session (`who`) **or** load average > 0.2 — there is no process-name check at all, so a running `loom-daemon` does **not** explicitly veto shutdown. Confirmed with `git log -p --all -- scripts/repo/repo-remote.sh | grep -n "pgrep\|loom-daemon"` (no hits) — this line was never present in this repo; the issue's quoted snippet describes a hypothetical/aspirational version, not the merged one.

This changes the actual gap the guard has:
- Today, a daemon-managed host that goes CPU-idle (load avg stays under 0.2, no interactive session) will already start counting down after `IDLE_MIN` regardless of whether the daemon considers itself "busy" waiting on a dispatch queue — there is no stacking-window problem from an explicit veto, because there is no veto.
- The real gaps are: (a) `IDLE_MIN`'s 120-minute default is sized for interactive hosts, not daemon-managed worker hosts, where a much shorter window is appropriate, and (b) the guard has no way to consume a daemon-authored signal (once loom#4467 lands) to start the countdown immediately from the moment the daemon itself observed idleness, rather than from whenever the guard's own load-average sampling happens to first read below the 0.2 threshold.

The `IDLE_MIN` config passthrough is **already generic and per-repo-configurable** (`REPO_REMOTE_IDLE_SHUTDOWN_MIN` in `.env`, verified by the `idle-shutdown guard content` test block in `commands/repo/tests/test-repo-remote.sh`), so "add a short-window profile for daemon hosts" reduces to a **documentation** change (recommend a value) rather than a **code** change — the code already supports it.

### Scope: this repo's side only

Per the issue's own stated boundary (rjwalters/loom#4340: repo:remote owns generic VM provisioning incl. this guard; loom consumes the provisioned host), **this issue is scoped to `scripts/repo/repo-remote.sh` / `commands/repo/remote.md` / their tests only.** rjwalters/loom#4467 (the daemon-side idle-exit implementation, `idleMinutes`, and loom's own `fleet add-worker` guard copy from loom#4341 step 7) is **upstream work in a different repository and is out of scope here.** This issue should not block on #4467 landing — it defines and implements the marker-file *contract* from repo:remote's side, so the daemon side can adopt it whenever it ships.

### Acceptance Criteria

- [ ] `commands/repo/remote.md`'s idle-shutdown guard section documents a recommended short idle window (e.g. `REPO_REMOTE_IDLE_SHUTDOWN_MIN=20`) for daemon-managed/worker hosts, distinct from the 120-minute interactive-host default, and explains why (daemon hosts don't need the same "operator forgot a session" margin).
- [ ] The doc note explicitly states the corrected mental model: the guard has no process-name veto today; "activity" is SSH session or CPU load only. If a future daemon-presence veto is added, it must be a deliberate, documented addition — not silently implied by the current text.
- [ ] `idle_guard_userdata()` in `scripts/repo/repo-remote.sh` optionally honors an idle-exit marker file: define a path/env var (e.g. `REPO_REMOTE_IDLE_MARKER`, default `/var/run/repo-remote-daemon-idle.marker`) whose **mtime**, when the file exists, the guard treats as an authoritative "idle since" timestamp — replacing (not merely supplementing) its own `$STAMP`-based countdown start for that check. This is the self-contained contract repo:remote publishes; loom#4467's daemon-side idle-exit can write this file once implemented, but the guard must work standalone (marker absent ⇒ falls back to current `who`/load-average/`$STAMP` behavior unchanged).
- [ ] `commands/repo/remote.md` documents the marker file path/env var and its exact semantics (mtime = last-known-idle timestamp; guard shuts down `IDLE_MIN` minutes after that mtime, independent of the guard's own SSH/load sampling).
- [ ] `commands/repo/tests/test-repo-remote.sh` covers: (a) the existing `who`/load-average/`$STAMP` behavior is unchanged when no marker is configured/present (regression guard), and (b) the generated cloud-init user-data script includes the marker-file check when the env var is set, and the correct default path is embedded when it's unset but the feature is enabled (adjust exact assertions to match the implementation's actual toggle).
- [ ] No change references, requires, or blocks on rjwalters/loom#4467 landing — the guard-side marker convention must stand alone.

### Implementation Guidance

- `idle_guard_userdata()` (repo-remote.sh:255-275) generates a heredoc (`cat >/usr/local/bin/repo-remote-idle-check <<'GUARD'`) that becomes the cron-invoked script on the provisioned host. Add the marker-file check as an additional branch inside that heredoc, staying consistent with the existing `\$`-escaping convention (the heredoc is quoted (`<<'GUARD'`) so host-side `$` must be backslash-escaped in the parent script, same as `\$STAMP`, `\$NOW`, `\$LAST` already are).
- Suggested logic sketch (inside the generated `repo-remote-idle-check` script, adjust as needed):
  ```bash
  MARKER=${REPO_REMOTE_IDLE_MARKER:-/var/run/repo-remote-daemon-idle.marker}
  if [ -f "\$MARKER" ]; then
    MARKER_AGE_MIN=\$(( ( \$(date +%s) - \$(stat -c %Y "\$MARKER" 2>/dev/null || stat -f %m "\$MARKER") ) / 60 ))
    [ "\$MARKER_AGE_MIN" -ge "\$IDLE_MIN" ] && /sbin/shutdown -h now "repo-remote: daemon idle-exit marker aged \${MARKER_AGE_MIN}m"
  fi
  ```
  Note `stat` flag differences between GNU (`-c %Y`) and BSD/macOS (`-f %m`) — the provisioned host is always Ubuntu (GNU coreutils), so `-c %Y` alone is sufficient on-host; no portability shim needed there.
- Whether the marker check runs *instead of* or *in addition to* the existing `who`/load-average check is a design decision for the implementer — the acceptance criteria above require it to be additive/optional so the existing behavior is preserved when the marker isn't in play, but the exact branch ordering in the generated script is an implementation detail.
- `IDLE_MIN` is already threaded through `idle_guard_userdata()` from `REPO_REMOTE_IDLE_SHUTDOWN_MIN` (repo-remote.sh:201) — reuse the same variable for the marker-based countdown rather than introducing a second window.
- `emit_plan_result()` / `emit_up_result()` (repo-remote.sh:~592, ~622) already surface `idle_shutdown_min` in JSON/log output — no changes needed there since the window value itself isn't changing, only how its start time can be sourced.

### Affected Files

- `scripts/repo/repo-remote.sh` — extend `idle_guard_userdata()` to optionally honor the idle-exit marker file/mtime; no change to the cost-gate or provisioning flow.
- `commands/repo/remote.md` — document the daemon-host short-window recommendation and the marker-file contract in the idle-shutdown guard section (~line 286).
- `commands/repo/tests/test-repo-remote.sh` — add coverage for the marker-file branch and a regression assertion that unmarked behavior is unchanged.

### Test Plan

- [ ] Manual verification: run `scripts/repo/repo-remote.sh up --yes --json` (against the mock `aws` CLI used by the test suite, or a real sandbox account) with `REPO_REMOTE_IDLE_MARKER` unset — confirm the generated user-data's guard script matches current behavior byte-for-byte modulo the new (inert) branch.
- [ ] Manual verification: with the marker env var/default enabled, hand-create the marker file on a test host (or in the test harness's generated script), backdate its mtime past `IDLE_MIN`, and confirm the guard's cron script would trigger `shutdown -h` (assert on the generated script's logic rather than actually powering off a host in CI).
- [ ] Automated tests: `pnpm test` (delegates to `commands/repo/tests/test-repo-remote.sh` via `hooks/repo/tests/run.sh`) — extend the existing "idle-shutdown guard content" test block (currently asserting `idle_shutdown_min` flows from config) with new assertions for the marker-file branch, both enabled and default/disabled.
- [ ] Doc-drift check: the test suite's existing doc-drift block (asserts `remote.md` documents implemented subcommands/flags) should be extended if a new `REPO_REMOTE_IDLE_MARKER` config knob is introduced, so `remote.md` and the script cannot silently diverge.
- [ ] Edge cases: marker file present but stale (very old mtime, e.g. from a previous VM lifecycle) — confirm it still triggers shutdown correctly rather than being ignored; marker file present but in the future (clock skew) — confirm no crash/negative-duration shutdown.

rjwalters:
## Summary

Extends the `repo:remote` cloud-init idle-shutdown guard with a self-contained **idle-exit marker-file contract** and documents a recommended **short idle window** for daemon-managed/worker hosts. This is the `repo:remote` half of the cross-repo contract; it stands alone and does not block on or reference the upstream daemon-side idle-exit (rjwalters/loom#4467, a different repo — out of scope here).

Per the curator's correction, the `pgrep -f loom-daemon` veto quoted in the original problem statement never existed in the merged guard — `idle_guard_userdata()` gates on `who` (SSH) or CPU load > 0.2 only. This PR addresses the real gaps: (a) the 120-min default is sized for interactive hosts, and (b) the guard had no way to consume a daemon-authored idle signal.

## Changes

- **`scripts/repo/repo-remote.sh`**
  - `resolve_settings()`: adds `IDLE_MARKER` from `REPO_REMOTE_IDLE_MARKER`, default `/var/run/repo-remote-daemon-idle.marker`.
  - `idle_guard_userdata()`: the generated on-host watchdog now embeds a marker branch. When the marker file exists, its **mtime** is treated as the authoritative "idle since" timestamp and shutdown fires `IDLE_MIN` minutes after it — **replacing** (via an `exit 0`) the local `$STAMP` countdown for that pass. The `who`/load veto still runs first (never powers off an in-use box); a future-dated marker yields a negative age and never triggers. Marker absent means unchanged `who`/load/`$STAMP` behavior. Added a comment block documenting the no-process-name-veto model so it isn't "simplified" away later.
- **`commands/repo/remote.md`**: new "The idle-shutdown guard" section — corrected activity model, interactive (120) vs daemon/worker (`REPO_REMOTE_IDLE_SHUTDOWN_MIN=20`) window guidance, and the full marker contract (path/default, mtime semantics, safety precedence, producer contract, two-stage composition). Config example lists both knobs.
- **`commands/repo/tests/test-repo-remote.sh`**: the mock `aws` now captures `--user-data` content; new assertions cover the marker branch (default + overridden path), that it replaces the stamp countdown, that the `who`/load/`$STAMP` behavior + absence of any process-name veto is preserved, plus doc-drift coverage for the new knob and short-window recommendation.

## The marker-file contract (design chosen)

`REPO_REMOTE_IDLE_MARKER` (default `/var/run/repo-remote-daemon-idle.marker`) names a file whose **mtime is an authoritative "idle since" timestamp**. The path is *always* embedded in the generated guard, so the branch is present-but-inert until the file exists on-host — making the guard fully standalone. Precedence within one guard pass: (1) an active SSH session or CPU load > 0.2 vetoes shutdown unconditionally (safety first); (2) else, if the marker exists, shut down when `now - mtime >= IDLE_MIN` and `exit 0`, so the marker *replaces* rather than supplements the local stamp countdown; (3) else fall back to the unchanged `$STAMP` logic. Producer side (a daemon such as loom's idle-exit): `touch` the file on clean idle-exit to hand the guard a precise idle-start, remove/re-`touch` it when work resumes; on-host mtime is read with GNU `stat -c %Y` (host is always Ubuntu). Clock-skew-safe (future mtime yields negative age, never fires) and stale-marker-safe (very old mtime fires correctly).

- [x] `remote.md` documents a recommended short idle window (`REPO_REMOTE_IDLE_SHUTDOWN_MIN=20`) for daemon/worker hosts vs the 120-min interactive default, with rationale.
- [x] `remote.md` states the corrected mental model: no process-name veto today; activity = SSH session or CPU load only; a future veto must be a deliberate documented addition.
- [x] `idle_guard_userdata()` honors `REPO_REMOTE_IDLE_MARKER` (default `/var/run/repo-remote-daemon-idle.marker`), using its mtime as an authoritative "idle since" that replaces the `$STAMP` countdown; works standalone (marker absent means current behavior unchanged).
- [x] `remote.md` documents the marker path/env var and exact mtime semantics (shutdown `IDLE_MIN` minutes after mtime, independent of the guard's own sampling).
- [x] Tests cover (a) unchanged `who`/load/`$STAMP` behavior as a regression guard and (b) the generated user-data embeds the marker check with the default path when the env var is unset and the overridden path when set.
- [x] No change references, requires, or blocks on rjwalters/loom#4467.

## Test Plan

- [x] `pnpm test` (worktree): **893 pass / 0 fail**; `test-repo-remote.sh` **87 pass / 0 fail** (+17 new assertions).
- [x] `bash -n` on `repo-remote.sh`, the generated user-data, and the extracted on-host guard script — all valid.
- [x] Behavioral verification of the on-host guard logic: no marker seeds stamp; marker aged 300m triggers shutdown; marker aged 30m no shutdown; future-dated marker (clock skew) no shutdown.

Closes #78

### DIFF
--- a/commands/repo/remote.md
+++ b/commands/repo/remote.md
@@ -127,7 +127,8 @@ REPO_REMOTE_DOCKERFILE=./Dockerfile       # optional: build & run this checked-i
 REPO_REMOTE_SETUP="make setup"            # optional first-boot command; fallback when no Dockerfile
 
 # --- session ---
-REPO_REMOTE_IDLE_SHUTDOWN_MIN=120
+REPO_REMOTE_IDLE_SHUTDOWN_MIN=120         # interactive-host default; use ~20 for daemon-managed/worker hosts (see below)
+REPO_REMOTE_IDLE_MARKER=                   # optional idle-exit marker path (default: /var/run/repo-remote-daemon-idle.marker)
 ```
 
 Only `REPO_REMOTE_PROVIDER` (or a provider argument) and that provider's
@@ -285,12 +286,75 @@ Requirements for the created instance:
   on AWS, quota-aware handling.
 - Install an idle-shutdown guard (cron checking SSH sessions + CPU, running
   `shutdown -h` after `REPO_REMOTE_IDLE_SHUTDOWN_MIN`) so a forgotten VM — GPU
-  ones especially — doesn't burn money
+  ones especially — doesn't burn money. See **The idle-shutdown guard** below for
+  its exact activity model, the daemon-host short-window recommendation, and the
+  idle-exit marker contract.
 - AWS: security group allowing SSH from the user's IP only, using
   `REPO_REMOTE_SSH_KEY`'s public key. GCP: prefer OS Login / IAP.
 
 If the zone/region is stocked out (common for GPU types), offer the nearest
 alternative zone or the next type down rather than failing.
+
+#### The idle-shutdown guard
+
+The guard is a cron watchdog (`/usr/local/bin/repo-remote-idle-check`, run every
+minute) installed via cloud-init user-data. It powers the host off with
+`shutdown -h` once it has been idle for `REPO_REMOTE_IDLE_SHUTDOWN_MIN` minutes.
+
+**Activity model (what keeps the host alive).** "Activity" is exactly two local
+signals: an **open SSH session** (`who`) **or** a **CPU load average > 0.2**.
+There is deliberately **no process-name veto** — a running background process
+(including `loom-daemon`) does **not**, on its own, keep the host alive. If a
+future daemon-presence veto is ever wanted, it must be added as a deliberate,
+documented change to `idle_guard_userdata()`; it is not implied by the current
+text. (This corrects an earlier problem statement that assumed a
+`pgrep -f loom-daemon` veto existed — it never did in this repo.)
+
+**Idle window — pick per host role:**
+
+- **Interactive hosts** (a human SSHes in to work): keep the default
+  `REPO_REMOTE_IDLE_SHUTDOWN_MIN=120`. The generous 2-hour margin is sized for
+  "operator stepped away from a session" so the box isn't yanked out from under
+  someone mid-task.
+- **Daemon-managed / worker hosts** (e.g. a box running `loom-daemon` for a
+  fleet, with no interactive operator): use a **short window such as
+  `REPO_REMOTE_IDLE_SHUTDOWN_MIN=20`**. These hosts have no human session to
+  protect, so the "operator forgot a session" margin is pure wasted spend — on a
+  c7i.2xlarge-class worker (~$8.60/day) the difference between a 120- and a
+  20-minute window is real money once the fleet goes quiet.
+
+**Idle-exit marker contract (`REPO_REMOTE_IDLE_MARKER`).** For daemon-managed
+hosts, the guard also honors an **idle-exit marker file** so a daemon can hand
+the guard a precise "idle since" time instead of waiting for the guard's own
+once-a-minute load sampling to first read idle. This is a self-contained
+contract published by `repo:remote` — the guard works standalone whether or not
+anything ever writes the file:
+
+- **Path.** `REPO_REMOTE_IDLE_MARKER` sets the file path the guard watches;
+  unset, it defaults to **`/var/run/repo-remote-daemon-idle.marker`**. The path
+  is always embedded in the generated guard, so the branch is present-but-inert
+  until the file exists on-host.
+- **Semantics.** When the marker file **exists**, the guard treats its **mtime**
+  as an authoritative "idle since" timestamp and powers off once that mtime is
+  older than `REPO_REMOTE_IDLE_SHUTDOWN_MIN` minutes — **replacing** (not
+  supplementing) its own internal stamp countdown for that pass. In other words:
+  shutdown fires `IDLE_MIN` minutes after the marker's mtime, independent of when
+  the guard's SSH/load sampling happened to first observe idleness.
+- **Safety precedence.** An active SSH session or CPU load > 0.2 still vetoes
+  shutdown *before* the marker is consulted — the guard never powers off a host
+  someone is actively using, even if a stale marker is present. A marker with a
+  **future** mtime (clock skew) yields a negative age and simply never triggers,
+  so it can't cause a spurious power-off.
+- **Producer contract (for a daemon side, e.g. loom's idle-exit).** On a clean
+  idle-exit, `touch` the marker path (creating/updating its mtime). To reset the
+  clock when work resumes, remove the file (or `touch` it again). The on-host
+  image is Ubuntu (GNU coreutils), so the guard reads the mtime with
+  `stat -c %Y`. This repo neither writes nor depends on any daemon writing the
+  file — the convention stands alone and a daemon may adopt it whenever it ships.
+
+On a daemon-managed host the two stages compose: the daemon's own idle-exit
+(writing the marker) is the first stage; this guard, `IDLE_MIN` minutes later, is
+the second and final stage that actually powers the box off.
 
 #### GPU hosts
 
--- a/commands/repo/tests/test-repo-remote.sh
+++ b/commands/repo/tests/test-repo-remote.sh
@@ -134,6 +134,17 @@ case "$1 $2" in
       echo "An error occurred (VcpuLimitExceeded) when calling the RunInstances operation" >&2
       exit 254
     fi
+    # Capture the generated user-data (idle guard) so the suite can assert on the
+    # guard script's content. --user-data is passed as `file://<path>`; the temp
+    # file still exists at call time (repo-remote.sh deletes it only afterwards).
+    prev=""
+    for a in "$@"; do
+      if [[ "$prev" == "--user-data" ]]; then
+        f="${a#file://}"
+        [[ -f "$f" ]] && cat "$f" >"${MOCK_AWS_LOG}.userdata"
+      fi
+      prev="$a"
+    done
     echo "${MOCK_AWS_NEW_ID:-i-0newinstance}"; exit 0 ;;
   "ec2 wait"|"ec2 start-instances"|"ec2 stop-instances"|"ec2 terminate-instances")
     exit 0 ;;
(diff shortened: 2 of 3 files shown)


rjwalters:
## Approved

Reviewed the idle-exit marker contract for the `repo:remote` idle-shutdown guard against all stated priorities. Verdict: **approve**.

**Correctness of the generated on-host guard (verified by emitting the user-data and behaviorally driving each branch):**
- `bash -n` clean on `repo-remote.sh`, the emitted cloud-init user-data, and the inner `GUARD` heredoc.
- Precedence is exactly right: (1) `who`/CPU-load>0.2 veto runs *first* and wins even with a stale marker present (never powers off an in-use box); (2) marker branch, when the file exists, computes `age = (now - mtime)/60` and `exit 0`s afterward, so it **replaces** rather than supplements the `$STAMP` countdown; (3) `$STAMP` fallback only reached when no marker. Behavioral matrix confirmed: active+stale-marker → no shutdown; idle+300m/30m → fires; idle+10m → no fire; idle+future-mtime (clock skew, negative age) → never fires; idle+no-marker → seeds stamp on first pass.
- Stale-marker-safe and clock-skew-safe as documented; `|| echo 0` guards a vanished/unreadable file; `stat -c %Y` is correct for the Ubuntu (GNU coreutils) host.

**Curator's premise correction confirmed:** no `pgrep`/process-name veto is introduced. The only `pgrep` mentions are the doc paragraph stating the veto never existed and two `assert_not_contains` regression guards proving it isn't embedded. Docs pin the "SSH session or CPU load > 0.2 only" activity model and require any future daemon-presence veto to be a deliberate, documented addition.

**Scope discipline:** exactly the 3 in-scope files (`scripts/repo/repo-remote.sh`, `commands/repo/remote.md`, `commands/repo/tests/test-repo-remote.sh`). No reference to, dependency on, or block on upstream loom#4467. The contract is self-contained: the marker path is always embedded so the branch is present-but-inert until a file exists on-host — the guard works standalone before any daemon writes the marker.

**Cost-safety unchanged:** the change can only make an idle host shut down sooner or equal (marker mtime = true idle-start ≤ the stamp's first-observed-idle time), never keeps an active host alive (veto runs first, unconditionally), and the no-marker path is byte-for-byte the prior `who`/load/`$STAMP` behavior.

**CI / mergeability:** `pnpm test` in the worktree: **893 pass / 0 fail** (repo-remote suite 87 pass, +17 new). GitHub CI check passes. `mergeStateStatus: CLEAN`, `mergeable: MERGEABLE` — cleanly disjoint from the v0.7.0 / version.sh commits merged to main since branching. PR body uses `Closes #78`.

Issue #86, opened 2026-07-31
/repo:tidy denylist should name git worktree roots explicitly


rjwalters:
## Context

Surfaced while running `/repo:all` on `rjwalters/loom`, a repo that uses git
worktrees for concurrent agent work.

`/repo:tidy`'s categorization is already conservative in principle — its
never-delete denylist covers secrets and environments, and anything gitignored
that doesn't match the SAFE/CACHE allowlists falls through to ASK. That rule
does cover this case. But the denylist names concrete examples (`.env`,
`.venv/`, `node_modules/`) and **git worktree roots are not among them**, which
makes the risk easy to miss on exactly the repos where it is largest.

Concretely, in this repo:

- `.loom/worktrees/` is **gitignored** (it is runtime state) and held **61 GB**
  across ~30 worktrees of in-flight, uncommitted agent work.
- `.claude/worktrees/` held another 3.1 GB.
- Both therefore appear in `git clean -ndX` output.

A `git clean -fdX` scoped from that listing — which the skill explicitly warns
against, but which is the obvious wrong turn — would destroy all of it. The
existing guidance gets to the right answer by inference; naming worktrees makes
it non-inferential.

This is a **documentation/robustness nicety, not a live defect** — the current
rules already route these to ASK. File or close at your discretion.

## Suggested acceptance criteria

- [ ] The never-delete denylist explicitly names git worktree roots (any
      directory containing a `.git` **file** pointing at
      `…/.git/worktrees/…`, plus anything listed by `git worktree list`).
- [ ] The inventory step runs `git worktree list` and excludes those paths from
      SAFE/CACHE regardless of gitignore status.
- [ ] The report calls out when a gitignored path is a live worktree root, with
      its size, so the scale is visible before any decision.

## Notes

- Detecting these robustly is cheap: `git worktree list --porcelain` gives the
  authoritative set, including worktrees living outside the repo root (this repo
  had several under `/private/var/folders/…` and `/private/tmp/`).

### Follow-up comment (2026-08-01, same author)

A second `/repo:all` run on the same repo hit a complementary angle: Tidy
reported "nothing to do" on a repo where `loom-daemon clean --safe --dry-run`
found 16 of 35 worktrees reclaimable (66 GB). The suggested scope addition
(inventory only, not deletion):

- When a denylisted worktree root is encountered, report its size and count as
  an inventory line rather than silently skipping it.
- Where the repo ships its own worktree tooling, point at it (e.g.
[...]

Issue #85, opened 2026-07-31
/repo:tidy inventory uses -not -path instead of -prune; find stalls on large repos


rjwalters:
## Context

Hit while running `/repo:all` on a large repo (`rjwalters/loom`, ~87 GB working
tree: a 24 GB Rust `target/`, ~30 live agent worktrees totalling 61 GB).

The Tidy stage's step-1 inventory commands filter with `-not -path` rather than
pruning:

```bash
find . -type d -empty -not -path './.git/*'
find . -type f -size +10M -not -path './.git/*' -not -path './node_modules/*'
```

`-not -path` only suppresses *printing*. `find` still descends into every
excluded directory, so on a repo with a large `.git/`, `node_modules/`,
`target/`, or nested worktrees it walks the entire tree anyway.

**Observed:** an inventory `find` over this tree exceeded a 120-second timeout
and had to be backgrounded. The equivalent with `-prune` returned essentially
instantly:

```bash
# times out
find . -name '.DS_Store' -o -name '*.orig' ...

# instant
find . \( -path ./.git -o -path ./node_modules -o -path ./target \
          -o -path ./.loom/worktrees -o -name .venv -o -name node_modules \) -prune \
       -o -type f \( -name '.DS_Store' -o -name '*.orig' ... \) -print
```

Practical impact: on a big repo the Tidy stage either stalls or gets abandoned,
which is exactly the situation where an inventory is most useful.

## Suggested acceptance criteria

- [ ] `/repo:tidy` step-1 inventory commands use `-prune` for excluded trees
      instead of `-not -path`.
- [ ] The pruned set covers at minimum `.git/`, `node_modules/` (at any depth,
      not just top level), and common heavy build/vendor dirs (`target/`,
      `dist/`, `.venv/`).
- [ ] Pruning changes performance only, never the reported inventory — a repo
      with junk inside a non-pruned path still reports it identically.
- [ ] Same fix applied to the `find` in `/repo:audit` if it shares the pattern.

## Notes

- `git clean -ndX` is unaffected (git handles its own traversal) — this is
  specifically about the raw `find` invocations.
- Worth pairing with a note that on repos using git worktrees the inventory can
  legitimately span tens of GB of *live* work; see the companion issue about the
  denylist.

### Root cause

Confirmed against current `main` (`commands/repo/tidy.md`, `bc9fe24`). The two
offending commands are step 1 ("Inventory"), lines **46** and **49**:

```
[...]

PR #91, opened 2026-08-04
fix(tidy): gate empty-dir SAFE rule on the never-delete denylist

### LINKED ISSUE #88: tidy: empty dirs are classified SAFE, so tool scaffolding is auto-deleted before the ASK fallthrough can protect it
## Context

Surfaced running `/repo:all` on `CensusAPI/censusapi`, a repo with Loom and wrangler installed.

The inventory step turned up five empty directories, all of which the command's own rules would auto-delete:

```
./.loom/sweep-checkpoint
./.loom/locks
./.loom/worktrees
./.loom/sweep-run
./.wrangler/tmp
```

These are **tool scaffolding** — coordination and runtime-state roots that Loom and wrangler expect to exist. I kept them by overriding the command's guidance, but a straight reading of `tidy.md` says to delete them.

## The mechanism, and why #86 does not already cover it

`tidy.md:46` scans with `find . -type d -empty -not -path './.git/*'`, and line 96 lists **"Empty directories"** under **SAFE — pure junk, auto-deleted by default**.

**#86** makes a closely related argument about worktree roots, and notes:

> anything gitignored that doesn't match the SAFE/CACHE allowlists falls through to ASK. That rule does cover this case.

That is true for a **populated** directory. It stops being true the moment the directory is empty. The empty-dir rule is an **independent SAFE classifier that fires before the allowlist fallthrough**, so the ASK protection #86 relies on is bypassed precisely by emptiness. `.loom/locks` is gitignored, matches neither SAFE nor CACHE by path, and would still be auto-deleted — not because it fell through, but because it is empty.

So this is the same blast radius as #86 by a different route, and the route matters: #86's proposed fix (naming worktree roots in the denylist) does not close it unless the empty-dir rule is also made to consult the denylist.

## Why "empty" is a bad junk signal here

For a lock or coordination directory, **emptiness is the normal, healthy state** — an empty `.loom/locks` means nothing is currently held, not that the directory is abandoned. The empty-dir heuristic reads a tool working correctly as evidence of clutter. Same for `.loom/sweep-run` and `.loom/worktrees` between runs, and `.wrangler/tmp` when no build is in flight.

Whether deleting them actually breaks anything I did not test — most such dirs are recreated on demand. The concern is the ones that are not, and the fact that a routine hygiene pass is where you would find out.

- [ ] Make the empty-dir SAFE rule consult the never-delete denylist rather than short-circuiting it, so any denylist fix (including #86's) covers empty dirs too
- [ ] Add tool-scaffolding roots to that denylist — at minimum `.loom/`, `.wrangler/`, `.anvil/` runtime dirs
- [ ] State in `tidy.md` that an empty coordination/lock directory is normal state, not clutter, so the reasoning survives future edits
- [ ] Consider narrowing the empty-dir rule to *tracked* areas of the tree, where a stray empty dir really is leftover

## Related

**#86** — same concern, populated dirs via the denylist/ASK path; this is the empty-dir classifier that bypasses it. Worth fixing together.
**#85** — also `/repo:tidy` inventory (`-not -path` vs `-prune`), same scan.

rjwalters:
## Summary

`/repo:tidy`'s Categorize step states that tests apply "denylist first, then the SAFE and CACHE allowlists, then fall through to ASK" — but the **Empty directories** bullet under SAFE was an independent classifier that never reached the denylist check. Emptiness alone promoted a path into the auto-delete set, so tool scaffolding (`.loom/locks/`, `.loom/worktrees/`, `.loom/sweep-run/`, `.loom/sweep-checkpoint/`, `.wrangler/tmp/`) was swept with no human gate — precisely because those directories are empty when the tool is working correctly.

This is the same blast radius as #86 by a different route: #86's denylist-naming fix does nothing for the empty-dir scan, which did not consult the denylist at all. This PR closes the bypass structurally (the empty-dir rule now consults the denylist) *and* names the scaffolding roots, so any future denylist addition — including #86's worktree roots — automatically covers empty dirs too.

## Changes

- **Denylist**: new "Tool scaffolding / coordination roots" bullet covering `.loom/`, `.anvil/`, and `.wrangler/` runtime dirs, matched by **parent tool-directory prefix rather than an enumerated leaf list**, so coordination subdirectories added later are covered without editing `tidy.md` again.
- **SAFE**: the "Empty directories" bullet is now explicitly gated on the denylist, with a note that the step-1 `find . -type d -empty` inventory is a raw path scan consulting neither gitignore nor the denylist — so the check must be applied to its output before anything is deleted.
- **ASK**: denylisted empty directories are surfaced here explicitly; report example gains `.loom/locks/` and `.wrangler/tmp/` lines.
- **Apply**: the "never pass a denylisted path to `git clean -fdX`" warning now names tool-scaffolding roots **including when empty**.
- **Safety Rules**: rule 6 names the new denylist category; new **rule 8 ("Empty ≠ abandoned")** records the reasoning — for a lock or coordination directory, empty is the normal operating state, not clutter — so it survives future edits.
- CHANGELOG `Unreleased` entry.

| Empty-dir SAFE rule consults the never-delete denylist rather than short-circuiting it, so any denylist fix (incl. #86's) covers empty dirs too | done | SAFE bullet now reads "only after checking the path against the never-delete denylist above, exactly as for the file patterns"; the gate is on the denylist as a whole, not on an inline list of paths, so #86's worktree-root addition inherits it automatically |
| Tool-scaffolding roots added to the denylist — at minimum `.loom/`, `.wrangler/`, `.anvil/` runtime dirs | done | New denylist bullet names all three, matched by parent tool-directory prefix; all five paths from the reporter's repro fall under it |
| `tidy.md` states that an empty coordination/lock directory is normal state, not clutter | done | Stated twice for redundancy: in the denylist bullet itself and as Safety Rule 8, which also restates the denylist-before-empty-dir ordering |
| Consider narrowing the empty-dir rule to *tracked* areas of the tree | considered, not taken | See "Alternative considered" below |

## Alternative considered

The issue's fourth bullet suggests skipping empty-dir auto-delete under any gitignored parent, which would close this bug class without relying on an enumerated denylist staying current. Not taken: it is inconsistent with how the rest of `tidy.md` works (allowlist/denylist patterns, not tracked-vs-untracked narrowing), and it would newly protect genuinely-junk empty dirs inside gitignored build trees. The prefix-matched denylist gets most of the durability benefit — new `.loom/*` coordination dirs are covered without an edit — while keeping the existing structure. Worth revisiting if scaffolding roots outside the enumerated tool prefixes turn up.

## Test Plan

- `pnpm test` (`hooks/repo/tests/run.sh`): **893 pass / 0 fail** — no regression. No automated coverage exists for `tidy.md` itself; it is a prose command spec interpreted by an agent at runtime.
- Manual walkthrough of the revised Categorize order against the reporter's repro set: `.loom/sweep-checkpoint`, `.loom/locks`, `.loom/worktrees`, `.loom/sweep-run`, `.wrangler/tmp` all match the tool-scaffolding denylist prefix and route to **ASK**.
- Non-regression check for genuine junk: an empty `build/` matches no denylist entry, so it still hits the empty-dir SAFE bullet and auto-deletes as before.
- Edge case per the issue: emptiness does not retroactively make a path "not a coordination root" — the denylist matches on path prefix and is checked before the empty-dir rule, so a scaffolding root that is the only thing under its parent still routes to ASK.

## Notes for the reviewer

- Scope is `commands/repo/tidy.md` (plus a CHANGELOG entry). `.claude/commands/repo/tidy.md` is this repo's gitignored local self-install copy and was deliberately **not** touched.
- The curator comment on #88 posted as a bare local filesystem path instead of its body (a tooling glitch, not a content problem) — its guidance was recovered from that file and followed.
- #85 (`-not -path` vs `-prune` on the same `find` invocations) is orthogonal and intentionally left alone.

Closes #88


rjwalters:
❌ **Changes Requested**

The core fix is right and the diagnosis matches the issue and the curator comment exactly: the empty-dir bullet was an independent SAFE classifier that never reached the denylist, and gating it on the denylist as a whole (rather than on an inline path list) is the correct structural fix — it means #86's worktree-root addition inherits the protection for free. Rule 8 is a good call: it records *why* in a place that survives future edits, which is exactly what the issue's third acceptance bullet asked for. The "Alternative considered" section is an honest, well-reasoned answer to the issue's fourth bullet.

Two wording problems in the new text need fixing before merge. Both are small edits confined to `commands/repo/tidy.md`, and both matter more than usual here because this file's entire deliverable *is* prose semantics read by an agent at runtime — an ambiguity in the spec is a behavior bug.

---

### 1. (Blocking) The new denylist bullet collides with the CACHE allowlist — `--caches` may become inert

`commands/repo/tidy.md:84-89`:

> Tool scaffolding / coordination roots: … and the equivalent runtime dirs under **any other tool's dot-directory**. **Match by parent tool-directory prefix, not by an enumerated leaf list**

The denylist is "checked first, overrides everything below". The CACHE allowlist (lines 123-126) enumerates dot-directories that are, literally, tools' dot-directories holding state the tool creates and manages: `.pytest_cache/`, `.mypy_cache/`, `.ruff_cache/`, `.turbo/`, `.astro/`. `.turbo/` and `.astro/` are the sharpest cases — a Turborepo/Astro dot-directory *is* "a tool's dot-directory" containing runtime state the tool manages, and "match by parent tool-directory prefix, not by an enumerated leaf list" instructs matching the whole tree under it.

Read that way, those CACHE entries are captured by the denylist, route to ASK, and `--caches` no longer clears them. That is a regression of a shipped feature introduced by this PR, in the opposite direction from #88 but still a break.

**Requested fix:** add an explicit carve-out to the tool-scaffolding bullet, mirroring the parenthetical the CACHE section already uses for `node_modules/`/virtualenvs (line 129-130). Something like:

> (Cache dot-directories already named under CACHE — `.pytest_cache/`, `.mypy_cache/`, `.ruff_cache/`, `.turbo/`, `.astro/` — are **not** coordination roots: they are regenerable output and stay in CACHE. The distinction is that a coordination root is one where *empty is the normal state*.)

The "empty is the normal state" test is a clean discriminator and you've already stated it in the bullet and in rule 8 — it just needs to be named as the boundary rather than left implicit.

### 2. (Blocking) The empty-dir SAFE bullet contradicts itself, which may leave the rule dead

`commands/repo/tidy.md:106-114`, inside the bullet list introduced by "A file is SAFE only if it does not match the denylist above **and matches one of**:"

> - Empty directories — but only after checking the path against the never-delete denylist above … **Emptiness is not itself a junk signal and never promotes a path into SAFE**: … and **an empty gitignored directory matching no allowlist still falls through to ASK**.

Two problems, both self-referential:

- "never promotes a path into SAFE" appears as an item *in the SAFE allowlist*, where emptiness is precisely the thing that promotes a path into SAFE. Read literally, the bullet negates itself.
- "an empty gitignored directory matching no allowlist" is circular: `Empty directories` *is* an allowlist entry, so no empty directory can "match no allowlist". The clause only carries meaning if "allowlist" silently excludes the bullet it appears in — under which reading every empty gitignored dir routes to ASK.

Under that second reading the rule is effectively dead, which contradicts both the report example at line 153-156 (`6 empty directories` listed under SAFE) and this PR's own non-regression claim in the Test Plan ("an empty `build/` … still hits the empty-dir SAFE bullet and auto-deletes as before"). The failure direction is fail-safe (over-ASK, nothing wrongly deleted), so it isn't dangerous — but it does mean the spec no longer says what the PR says it says.

**Requested fix:** state the gate positively and drop the two self-negating clauses, e.g.:

> - Empty directories **whose path matches no never-delete denylist entry** — the denylist is checked first here exactly as it is for the file patterns above. An empty directory under a tool-scaffolding / coordination root (`.loom/`, `.anvil/`, `.wrangler/`, …) routes to **ASK**, not SAFE: emptiness is that tool's normal operating state, not evidence of junk. Note that the `find . -type d -empty` inventory in step 1 is a raw path scan — it consults neither gitignore nor the denylist — so this check must be applied to its output before anything is deleted.

That keeps everything the fix needs (denylist-first ordering, the ASK routing, the raw-scan warning) and leaves the non-denylisted empty `build/` case behaving as before. If routing empty *gitignored* dirs to ASK was in fact intended, say so explicitly and update the Test Plan and the line-153 report example to match — but please don't leave both readings available.

---

### Non-blocking observations (no action required)

- `commands/repo/all.md:62-64` still summarizes the tidy stage as deleting "SAFE junk (OS droppings, merge leftovers, empty dirs)". Still accurate — those are the non-denylisted empty dirs — and `all.md` defers to `[[tidy]]` for categorization detail, so no edit needed.
- `.claude/commands/repo/tidy.md` is a symlink to `commands/repo/tidy.md` in this checkout, not a copy, so the "deliberately not touched" note in the PR description is correct and there's nothing to re-sync.
- Denylisting `.loom/worktrees/` wholesale means populated worktree roots now surface under ASK too. That overlaps #86's ask and is the right direction; noting it only so the overlap is visible when #86 lands.

## Test Scoping

**Strategy**: `full-suite` — the repo's `check:ci` is a single shell harness (`hooks/repo/tests/run.sh`), no per-file scoping tool exists.
**Changed files**: 2 (`commands/repo/tidy.md`, `CHANGELOG.md`) — both prose; no executable code touched.
**Result**: `pnpm check:ci` → **893 pass / 0 fail**, matching the PR's claim. Run inside `.loom/worktrees/issue-88`.

## Test Execution

**Test plan from PR description:**
1. `pnpm test` (`hooks/repo/tests/run.sh`), expect 893/0 — ✅ Executed: 893 pass / 0 fail, no regression.
2. Manual walkthrough of revised Categorize order against the reporter's five repro paths — ✅ Executed by inspection: `.loom/sweep-checkpoint`, `.loom/locks`, `.loom/worktrees`, `.loom/sweep-run`, `.wrangler/tmp` all match the new tool-scaffolding prefix and route to ASK. This criterion is met.
3. Non-regression check for genuine junk (empty `build/` still SAFE) — ❌ Does not hold as written: see finding 2 above, the revised prose admits a reading in which it routes to ASK.
4. Edge case: emptiness doesn't retroactively make a path "not a coordination root" — ✅ Executed by inspection: denylist matches on path prefix and is checked first, so this holds.

## CI

`pnpm test (guard/handoff/loss-check suite)`: **success**. Merge state: **clean** (mergeable, no conflicts). CI is not the blocker here.

---

Nice work on the structural framing — gating on the denylist rather than on an enumerated path list is the part that makes this durable, and rule 8 is the right place for the reasoning. Tighten the two passages above and I'll approve.

rjwalters:
✅ **Both blocking findings addressed** (5e0d32b) — single file, `commands/repo/tidy.md`.

**1. Denylist/CACHE collision (`--caches` going inert).** Added the requested carve-out to the tool-scaffolding bullet, mirroring the parenthetical CACHE already uses for `node_modules/`/virtualenvs:

> (The cache dot-directories already named under CACHE — `.pytest_cache/`, `.mypy_cache/`, `.ruff_cache/`, `.turbo/`, `.astro/` — are **not** coordination roots: they are regenerable output and stay in CACHE, so `--caches` still clears them. The discriminator is that a coordination root is one where *empty is the normal state*; a cache directory is one whose entire contents can be rebuilt by re-running the tool.)

The "empty is the normal state" test is now named as the boundary rather than left implicit, and the second half of the sentence gives the complementary test for the CACHE side so the two tiers are separable from either direction.

**2. Self-contradicting empty-dir SAFE bullet.** Restated as a positive gate, using your wording. Both self-referential clauses ("never promotes a path into SAFE", "matching no allowlist") are gone:

> - Empty directories **whose path matches no never-delete denylist entry** — the denylist is checked first here exactly as it is for the file patterns above. An empty directory under a tool-scaffolding / coordination root (`.loom/`, `.anvil/`, `.wrangler/`, …) routes to **ASK**, not SAFE: emptiness is that tool's normal operating state, not evidence of junk. Every other empty directory (an empty `build/`, say) is SAFE. Note that the `find . -type d -empty` inventory in step 1 is a raw path scan — it consults neither gitignore nor the denylist — so this check must be applied to its output before anything is deleted.

I added one clause beyond your draft — "Every other empty directory (an empty `build/`, say) is SAFE" — to close the ambiguity affirmatively rather than only by omission, so the line-153 report example and Test Plan item 3 now have explicit backing in the spec. Routing empty gitignored dirs to ASK was **not** intended; the non-regression claim stands as written and no Test Plan / report-example edits are needed.

**Tests:** `pnpm test` → **893 pass / 0 fail**, unchanged. Prose-only diff (12 insertions, 7 deletions), no executable code touched.

Ready for re-review.

rjwalters:
✅ **Approved** — both blocking findings are genuinely resolved in `5e0d32b`.

Re-review focused on the two blockers from my previous comment, plus a full pass over the current diff (`origin/main...5e0d32b`).

---

### 1. Denylist/CACHE collision — resolved

`commands/repo/tidy.md:89-95` now carries the carve-out, and it does more than the minimum I asked for:

> (The cache dot-directories already named under CACHE — `.pytest_cache/`, `.mypy_cache/`, `.ruff_cache/`, `.turbo/`, `.astro/` — are **not** coordination roots: they are regenerable output and stay in CACHE, so `--caches` still clears them. The discriminator is that a coordination root is one where *empty is the normal state*; a cache directory is one whose entire contents can be rebuilt by re-running the tool.)

Verified the enumeration is exhaustive against the CACHE tier as written (`tidy.md:126-131`): the only dot-*directories* CACHE claims are those five, and all five are now explicitly excluded from the prefix-matched denylist. The remaining CACHE entries (`__pycache__/`, `dist/`, `htmlcov/`, `site/dist`, `*.pyc`) are not dot-directories, so the "any other tool's dot-directory" prefix rule never reaches them. `--caches` is no longer at risk of going inert.

Giving both directions of the discriminator (coordination root = empty is normal; cache = fully rebuildable by re-running the tool) is better than the one-sided test I suggested — it makes the tiers separable when a novel dot-directory shows up from either side.

### 2. Self-contradicting empty-dir SAFE bullet — resolved

`tidy.md:111-119` is now a positive gate with both self-referential clauses removed:

> - Empty directories **whose path matches no never-delete denylist entry** — the denylist is checked first here exactly as it is for the file patterns above. An empty directory under a tool-scaffolding / coordination root (`.loom/`, `.anvil/`, `.wrangler/`, …) routes to **ASK**, not SAFE: emptiness is that tool's normal operating state, not evidence of junk. Every other empty directory (an empty `build/`, say) is SAFE. Note that the `find . -type d -empty` inventory in step 1 is a raw path scan — it consults neither gitignore nor the denylist — so this check must be applied to its output before anything is deleted.

The added "Every other empty directory (an empty `build/`, say) is SAFE" was the right call — closing the ambiguity affirmatively rather than by omission means the line-169 report example (`6 empty directories` under SAFE), `commands/repo/all.md:95` (`Tidy: freed 240 MB (build/, .cache/, 3 empty dirs)`), and Test Plan item 3 all now have explicit backing in the spec. Only one reading survives, and it's the intended one. Denylist-first ordering, the ASK routing, and the raw-scan warning are all preserved.

---

### Sanity pass over the rest of the current diff

- **Denylist bullet** (`tidy.md:84-95`) — prefix-matched on the parent tool directory, so #86's worktree-root work inherits the empty-dir protection without a further edit here. All five paths from the reporter's repro (`.loom/locks/`, `.loom/worktrees/`, `.loom/sweep-run/`, `.loom/sweep-checkpoint/`, `.wrangler/tmp/`) match.
- **ASK section** (`tidy.md:139-144`) — the new "Any empty directory whose path matches the denylist" bullet makes the destination explicit rather than leaving it implied by the denylist's "always ASK" header. Good.
- **Report example** (`tidy.md:172-173`) — the two new ASK lines are annotated with *why* (`empty is normal state` / `empty between builds`), which is what makes the report actionable for a human.
- **Apply** (`tidy.md:197-199`) — "including when it is empty" closes the `git clean -fdX` path, which is where the actual deletion happens.
- **Safety rules 6 and 8** — rule 8 records the reasoning durably, which is the issue's third acceptance bullet.
- **CHANGELOG** — accurate, names both the mechanism and the blast radius, correct `Unreleased` placement.
- Scope discipline holds: `commands/repo/tidy.md` + CHANGELOG only. `.claude/commands/repo/tidy.md` is a symlink in this checkout, so nothing to re-sync.

### Non-blocking observations (no action required, do not hold the merge)

- **Rule 8's trailing clause** (`tidy.md:232-233`): "Emptiness … never bypasses the fall-through to ASK for unrecognized gitignored paths." Since `Empty directories` *is* a SAFE allowlist entry, an empty dir is never "unrecognized", so this clause is vacuous for directories — but a reader could stretch it into the reading finding 2 just eliminated (empty gitignored `.foocache/` → ASK). The operative Categorize section governs and is now unambiguous, and the failure direction is fail-safe, so this is a wording nit for whoever next touches the safety rules, not a defect.
- `commands/repo/host-optimize.md:302` still `rm -rf`s stale dirs under `.loom/worktrees/`. Not a contradiction — that's a different command applying a verified staleness test (git no longer lists the worktree), whereas tidy's denylist governs blind pattern sweeps. Flagging only so the distinction is on record.
- The `.loom/` prefix rule also captures tracked config dirs like `.claude/`. Harmless: the denylist only ever routes *toward* ASK, and tracked files are KEEP regardless.

## Test Scoping

**Strategy**: `full-suite` — `check:ci` is a single shell harness (`hooks/repo/tests/run.sh`); no per-file scoping tool exists in this repo.
**Changed files**: 2 (`commands/repo/tidy.md`, `CHANGELOG.md`) — prose only, no executable code touched.
**Result**: `pnpm check:ci` → **893 pass / 0 fail** (55 + 501 + 99 + 59 + 55 + 37 + 87), run in `.loom/worktrees/issue-88` at `5e0d32b`. Matches the pre-fix baseline exactly — no regression.

## Test Execution

**Test plan from PR description:**
1. `pnpm test`, expect 893/0 — ✅ Executed: 893 pass / 0 fail.
2. Manual walkthrough of Categorize order against the five repro paths — ✅ Executed by inspection: all match the tool-scaffolding prefix, all route to ASK.
3. Non-regression for genuine junk (empty `build/` still SAFE) — ✅ **Now holds**, and holds explicitly rather than by inference: "Every other empty directory (an empty `build/`, say) is SAFE." This was the failing item in my previous review.
4. Emptiness doesn't retroactively de-classify a coordination root — ✅ Executed by inspection: prefix match, denylist checked first.

## CI

`pnpm test (guard/handoff/loss-check suite)`: **pass** (1m6s). Merge state: **CLEAN** / MERGEABLE.

---

The structural framing was right the first time; these two edits make the prose say what the structure does. Since this file's deliverable *is* prose semantics executed by an agent, that distinction was worth the extra round. Ready to merge.

PR #92, opened 2026-08-04
feat(deps): add /repo:deps for Dependabot setup and bot-PR triage

rjwalters:
## Summary

Adds `/repo:deps`, a new command covering **third-party dependency currency** — Dependabot install/verify plus Dependabot PR triage.

### Placement decision (acceptance criterion 6)

**A new command, not an extension of `/repo:update-tools`.** `update-tools`' contract is "installed package version vs. a local source clone (or GitHub tags)"; Dependabot has no source clone to diff against, and "triage incoming bot PRs" is a different activity from "update an installed package." Folding it in would have put two incompatible comparison models in one command. `update-tools` gets a two-sentence scope pointer to `[[deps]]` and is otherwise unmodified.

### Snippets were executed, not just written

Every shell command in the new file was run against this repo before being committed. That corrected three defects in the first draft — each of which would have produced a silently wrong report:

| Defect | Symptom | Fix |
|---|---|---|
| Read alerts from `security_and_analysis.dependabot_alerts` | That key is **absent** on `rjwalters/repo` even though the object returns `dependabot_security_updates` and four `secret_scanning*` keys — the `// "UNKNOWN"` fallback reports "needs admin" for a repo the token reads fine | Use `GET /vulnerability-alerts` (204 enabled / 404 disabled); reserve UNKNOWN for a null `security_and_analysis` or a 403 |
| `ls .github/workflows/*.yml *.yaml 2>/dev/null` | Under zsh an unmatched glob is a hard error that **aborts the command line before `ls` runs** — this repo's `.github/workflows/ci.yml` went undetected, so `github-actions` would never be scaffolded. `2>/dev/null` does not help | All ecosystem probes use `git ls-files` |
| `select(.description \| test("Applied by: humans"))` | A label with a null description aborts jq: `null (null) cannot be matched, as it is not a string`, truncating the label scan mid-stream | `.description // ""`, plus the REST `…/labels` endpoint instead of GraphQL-backed `gh label list --json` |

## Changes

- **`commands/repo/deps.md`** (new) — 9 steps across two halves: config + security-flag reporting as distinct items, ecosystem detection, description-based label validation, gated config scaffold, gated flag enablement, immediate post-merge PR check, bot-PR listing, per-PR classification with Actions deprecation-annotation diffing, gated merge. 6 Safety Rules.
[...]

Issue #90, opened 2026-08-04
update-tools: add Dependabot installation and Dependabot PR review


rjwalters:
## Task Type

Feature

## Description

`/repo:update-tools` keeps *installer-managed tool packages* (Loom, Anvil, Repo
Skills) current. It has no coverage for the other half of "am I behind on
anything" — **third-party dependencies**: npm/pip/cargo packages and GitHub
Actions. Dependabot is the standard answer, and today nothing in the Repo Skills
surface installs it, verifies it, or helps triage what it produces.

Two additions requested:

1. **Install/verify Dependabot** as part of the tool-currency sweep.
2. **Review open Dependabot PRs** — surface them, summarize risk, offer to merge.

## Scoping question for the implementer

Worth deciding deliberately rather than inheriting: `update-tools` currently
defines its scope as packages discovered via `install-metadata.json`, compared
against a **local source clone**. Dependabot fits awkwardly — there is no source
clone to diff, and "triage incoming bot PRs" is a different activity from
"update an installed package." This may belong as a separate command
(`/repo:deps`?) with `update-tools` merely reporting Dependabot's absence. Filed
here because that is where it was requested; the split is the implementer's call.

## Design notes from doing this by hand

All of the following was learned installing Dependabot into `rjwalters/trawl`
manually on 2026-08-04. Each is a place a naive implementation gets it wrong.

**The config file and the repo setting are two separate things.** Writing
`.github/dependabot.yml` enables *version* updates. Dependabot **security**
updates are a repo-level flag that is independent and was found **disabled**:

```bash
gh api repos/OWNER/REPO --jq '.security_and_analysis.dependabot_security_updates.status'
# -> disabled
gh api -X PUT repos/OWNER/REPO/vulnerability-alerts        # prerequisite
gh api -X PUT repos/OWNER/REPO/automated-security-fixes
```

A skill that only writes the YAML silently leaves CVE alerting off. Check and
report both.

**Validate labels by description, not just existence.** The obvious label for
bot PRs in a Loom repo looked like `external`, until its description read:

> Non-collaborator submission; needs maintainer approval before curation.
> **Applied by: humans only.**

Having Dependabot apply it would violate the label's own contract. `tier:maintenance`
was correct instead. So the check is not `gh label list | grep -qx`, it is
"read the description and confirm a bot may apply it" — and per Loom policy,
never create a new label to solve this.

**Grouping policy should be per-ecosystem, not uniform.** What worked:

[...]

PR #93, opened 2026-08-04
fix(commands): verify applied fixes are still on disk before reporting them

### LINKED ISSUE #89: /repo:all: Docs stage edits can be silently lost in a repo with a concurrent agent daemon
## Context

Surfaced during a `/repo:all` run in a consumer repo (2AMLogic/marketing) that is Loom-managed, with the Loom daemon actively running.

The Docs stage applied three prose fixes to the working tree while on `main`:

- `README.md` — pulse adapter list was missing two live sources
- `website/README.md` — stale "daily" cadence
- `strategy/pulse-design.md` — same stale cadence

All three `Edit` calls reported success. A moment later the working tree was clean and every edit was gone. A concurrent Loom sweep had run `check-main-clean.sh --quarantine` (see `defaults/scripts/check-main-clean.sh` in rjwalters/loom, ~line 383) to get a clean main worktree for its own worktrees, and stashed the contamination:

```
stash@{0}: On main: loom-quarantine: run=sweep-20260804T005038Z-73493-3482084e issue=63
```

Loom's behavior is defensible on its own terms — it stashes with a labeled message rather than discarding, so nothing is destroyed. The problem is on the `/repo:all` side: the Docs stage edits the working tree in place and never confirms the edits survived. Had the Reset stage not counted stashes, three doc fixes would have been silently dropped with the pass still reporting success.

## CORRECTION (2026-08-04): branching does NOT prevent this

The original filing said recovery "required re-doing all three edits on a branch, where they held," and proposed branch-before-edit as a fix. **Both are wrong.** The edits held in that instance only because no sweep happened to fire during the window — not because the branch protected them.

Confirmed later the same session while working #342 in 2AMLogic/klayout-tools. Edits were made on a branch (`release/v0.2.0`) from the outset, and were quarantined anyway:

```
stash@{0}: On release/v0.2.0: loom-quarantine: run=sweep-20260804T020551Z-35381-696503da issue=461
```

**The quarantine keys on the primary clone's working tree, not on the branch name.** Any uncommitted change in the main worktree is contamination as far as the sweep is concerned, whatever branch is checked out.

The workaround that does work is a git worktree created **outside** the policed clone:

```bash
git worktree add /some/path/outside branch-name
```

Edits there survive; the daemon only polices the primary clone. That is how #342's release prep was completed.

## Scope is broader than the Docs stage

The same exposure applies to every `/repo:*` command that writes to the working tree — `docs`, `readme`, `gitignore`, `tidy`, and the auto-apply fixes in `audit`. `/repo:all`'s Docs stage is just where it was first observed.

## Why this is distinct from #82

#82 is about stage *ordering* — running Reset's sync-and-switch before Docs when the working branch is fully pushed and behind the default branch. This is a different failure: the checkout was already fresh and correct, and the edits were lost to a *concurrent writer*, not to stale content or a blocked branch switch. Related, worth cross-referencing, not a duplicate.

- [ ] Commands that modify the working tree detect an active agent daemon before editing (a running `loom-daemon` process, or a populated `.loom/worktrees/` alongside `git worktree list`).
- [ ] In that case they either work in a worktree created outside the policed clone, or verify after applying that the edits are still present — **not** merely create a branch, which does not protect (see correction above).
- [ ] A stage that finds its own edits reverted reports that fact rather than counting the fix as applied in the final summary.
- [ ] The final summary never reports a fix as applied without confirming it is still on disk.
- [ ] Behavior is unchanged in repos with no concurrent daemon.

rjwalters:
## Summary

The `/repo:*` commands that edit tracked files applied a fix, trusted the `Edit` call's success, and reported it — with no re-check that the change survived. A concurrent writer in the same clone (another agent, a background `git stash` / `git checkout --`, a Loom sweep running `check-main-clean.sh --quarantine` on the primary clone's working tree) can revert the file between the edit and the report, so a pass can claim fixes that are no longer on disk. That is exactly what happened in the reported `/repo:all` run: three doc fixes were reported applied, then quarantined into a stash moments later.

The fix is **verify-after-write**, not daemon detection — detection is inherently racy, since a daemon can start right after the check.

## Changes

- **`commands/repo/docs.md`, `readme.md`, `gitignore.md`, `links.md`** — each gains a `### Verify after write` step in its Interaction section: re-read the change immediately after applying it and **before counting it as applied** (re-read the changed region, or `git diff -- <path>` / `git status --porcelain -- <path>`). Anything found gone is reported on its own line as **reverted after apply — needs re-run**, never silently re-applied and never counted in the fixed total. The step is explicitly unconditional, with the reason detection was rejected.
- **`commands/repo/all.md`** — new `### Re-verify before printing` subsection under Final Summary: stages re-verify their edits immediately before the consolidated summary prints, a stage's `N fixed` count only includes edits confirmed on disk, reverted edits get a distinct line (with a rendered example), and only the affected stage's line changes.
- **`commands/repo/tests/test-verify-fix-persistence.sh`** (new, 61 cases) — fixture section transcribing the documented check and exercising it against the real revert mechanisms (`git checkout --`, `git stash` with a `loom-quarantine` message, `git stash -u` for a fix that creates a new file), plus a doc-drift section asserting the five command files still say what the suite implements.
- **`hooks/repo/tests/run.sh`** — delegation block for the new suite, following the existing `test-branches-loss-check.sh` pattern (real PASS/FAIL folding, skip-count note, breakdown row).
- **`README.md`** repository-layout listing and a **`CHANGELOG.md`** Unreleased entry.

`tidy.md` and `audit.md` are deliberately untouched, per the curator's scope corrections — and the new suite asserts they stay that way.

| Criterion (curator's revised list) | Status | Verification |
|---|---|---|
| `docs.md` / `readme.md` / `gitignore.md` / `links.md` each document an unconditional verify-after-write step before counting a fix as applied | ✅ | Identical `### Verify after write` block in all four; asserted by 4×8 doc-drift cases |
| Each documents that a fix found reverted is reported as reverted, not silently re-counted | ✅ | "reverted after apply — needs re-run" + "Do not silently re-apply it, and do not count it in the fixed total" asserted in all four |
| `all.md` Final Summary documents re-verify-before-print and a distinct reverted line | ✅ | `### Re-verify before printing`; 6 doc-drift assertions incl. "never folded into that stage's fixed count" |
| `tidy` and `audit` explicitly out of scope | ✅ | Neither file modified; suite asserts neither contains the step, with a comment on when to revisit |
| Behavior unchanged with no concurrent writer | ✅ | Documented in all five files; fixture case "no concurrent writer: the check is invisible" shows every edit verifies APPLIED, the report is the plain `Docs: 2 fixed`, and verification does not modify the file (hash-object before/after) |
| New grep-based doc-drift suite following `test-branches-loss-check.sh`, wired into `run.sh` so `pnpm test` covers it | ✅ | `test-verify-fix-persistence.sh`, same structure/counters/summary shape; delegated from `run.sh` |

## Test Plan

- `pnpm test` — **954 pass, 0 fail** (was 893; +61 from the new suite), breakdown-sums-to-headline self-check green.
- New suite standalone: 61 pass, 0 fail, 0 skipped.
- **Negative control** (the point of a drift test): against a scratch copy with the `### Verify after write` block stripped from `links.md` and `all.md`'s "never folded into that stage's fixed count" / heading mutated, the suite fails 10 cases (8 links.md + 2 all.md) and exits 1. The unmutated tree passes.
- Fixture coverage of the actual loss mechanisms: `git checkout --` revert detected; `git stash` (labelled `loom-quarantine: run=… issue=…`) detected, with the content shown still recoverable from the stash — which is why "needs re-run" is the right report, not "lost"; `git stash -u` on a newly created file detected (the case `git status --porcelain` alone would miss, which is why content re-read is documented as primary).
- Mixed-stage edge case from the test plan: Docs reverted / Readme intact — only the Docs line reports the reversion, Readme still reports `1 fixed`.

Closes #89


rjwalters:
✅ **Approved!**

Full evaluation against issue #89 and the curator's revised acceptance criteria. The framing is right: verify-after-write rather than daemon detection, with the raciness of detect-then-skip stated in the docs themselves rather than left implicit.

## Rebase performed during evaluation

The branch was `DIRTY` against main after #91 and #92 landed — a `CHANGELOG.md`-only conflict (both sides adding bullets to the same `## Unreleased` block; `README.md` auto-merged). Resolved as a simple both-sides-add: the #89 entry sits on top, then #90, then #88, preserving the newest-first ordering already in main. Rebased onto `ac55b23` and force-pushed with `--force-with-lease` (`255c36c` → `32f136e`). No code changes were made.

- CI on the rebased head: `pnpm test (guard/handoff/loss-check suite)` **pass** (57s)
- `mergeStateStatus`: **CLEAN**

## Test Execution

Run in the worktree **after** the rebase, so #91's `tidy.md` changes and #92's `README.md`/`deps.md` additions are included:

1. `pnpm test` — ✅ Executed: **954 pass, 0 fail** (matches the PR's claim; +61 over the previous 893). Per-suite breakdown sums to the headline, and the new suite reports `test-verify-fix-persistence.sh  61 pass  0 fail  (verify-after-write contract)`.
2. New suite standalone — ✅ Executed: 61 pass, 0 fail, 0 skipped.
3. **Negative control re-run independently** (the load-bearing check for a doc-drift suite) — ✅ Executed against a scratch copy of `commands/` with the `### Verify after write` block truncated out of `links.md` and `all.md`'s heading mutated: **9 failures, exit 1** (8 × links.md + 1 × all.md heading). The suite genuinely fails on drift rather than passing vacuously. The unmutated tree passes.
4. Scope guard — ✅ Executed: `tidy.md` and `audit.md` assertions pass post-rebase; #91's `tidy.md` changes did not disturb the out-of-scope assertion.
5. Manual fixture verification (background process reverting an edit mid-run) — ⚠️ Not executed: requires a live `/repo:docs` run. The suite's fixture section covers the same three mechanisms mechanically (`git checkout --`, `git stash` with a `loom-quarantine` message, `git stash -u` on a new file).

`shellcheck -x` on the new suite: clean apart from 3 × SC2059 (info) on the coloured `printf` summary lines — the identical pattern already present in `test-branches-loss-check.sh`, so this is house style, not a regression. Shellcheck is not in CI here.

## Acceptance criteria

All six of the curator's revised criteria are met and independently verified:

| Criterion | Verified |
|---|---|
| Four applying commands document an unconditional verify-after-write step before counting a fix applied | ✅ identical block in `docs.md`/`readme.md`/`gitignore.md`/`links.md`, pinned by 4×8 drift cases |
| Reverted fix reported distinctly, never silently re-counted | ✅ "reverted after apply — needs re-run" + the do-not-re-apply sentence in all four |
| `all.md` Final Summary documents re-verify-before-print + distinct line | ✅ `### Re-verify before printing`, with a rendered example and "never folded into that stage's fixed count" |
| `tidy` / `audit` explicitly out of scope | ✅ neither modified; the suite asserts they stay that way, with a comment on when to revisit |
| Behavior unchanged with no concurrent writer | ✅ stated in all five files; fixture pins it with a `hash-object` before/after showing verification is side-effect free |
| New grep-based suite wired into `run.sh` so `pnpm test` covers it | ✅ same delegation shape as `test-repo-remote.sh`, including the skip-count-is-neither-pass-nor-fail handling from repo#46 |

## What I checked beyond the criteria

- **Heading placement** — the new `### Verify after write` sits under each file's `## Interaction` section in all four (in `links.md` it follows the fenced example output block, which is correct; the `##` lines at 58/65 are inside the fence).
- **`run.sh` delegation correctness** — `VP_STATUS=$?` after the command substitution assignment captures the suite's exit status correctly; the unparseable-summary branch forces `VP_FAIL=1` rather than silently contributing zero, and the `VP_STATUS -ne 0 && VP_FAIL -eq 0` branch also forces a failure. Skips are excluded from the folding so they cannot inflate the headline — consistent with the two suites above it.
- **Fixture realism** — the `git stash -u` case (a fix that creates a new file) is the one that a naive `git status --porcelain` check would misread, and it's covered; the stash assertions also pin that the quarantined content is still recoverable, which is why "needs re-run" is the honest report rather than "lost".
- **README repository-layout listing** — updated for the new suite, consistent with the existing two delegated-suite entries.

## One non-blocking follow-up: #95

The docs offer two arms for the re-check as equivalents:

> re-read the changed region of the file, or run `git diff -- <path>` / `git status --porcelain -- <path>`

They are not equivalent. The git arms answer "does this path differ from HEAD?", not "is my edit still present?" — so when the file carries other uncommitted changes (another agent editing the same doc, or a `/repo:all` stage that applied two fixes to one file and lost one), `git status --porcelain` still prints `M` and the check concludes APPLIED for an edit that is gone. That is the exact failure class #89 closes.

Notably the PR's own test already knows this — `verify_dirty()` is commented as "cheaper, but weaker: it only proves the path differs from HEAD, not that it differs in the way this command intended" — the caveat just doesn't reach the four command files an agent reads at runtime. Filed as #95 with a suggested wording change and a fixture case (two edits to one file, revert one, assert `verify_dirty` DIRTY vs `verify_applied` REVERTED).

Not blocking: the content re-read is listed first and is the primary form, and closing the gap means touching four files plus the drift assertions — a change worth its own review rather than a fix-during-evaluation.

Nice work on the negative control in particular — a doc-drift suite that has not been shown to fail on drift is worth very little, and this one was demonstrated both by the author and independently here.

PR #99, opened 2026-08-04
docs(followups): replace GraphQL gh issue calls with REST search + POST

rjwalters:
## Summary

`/repo:followups` prescribed GraphQL-backed `gh` subcommands in both of its
forge-touching steps, so the command failed in exactly the repos it is most
useful in: a busy multi-agent repo whose GraphQL bucket is already exhausted
while the REST (`core`) budget sits nearly unused (observed `core: 19/5000`,
`graphql` over limit). Step 3's dedup now uses the REST search endpoint, and
step 5's filing uses `gh api --method POST repos/<slug>/issues --input` with a
body written to a file and passed via `jq --rawfile` — which also keeps issue
bodies out of the shell's parser, so a body line containing `>=` or backticks
can no longer be tokenized and denied by content-matching guards (the secondary
benefit noted in the issue, complementary to #84).

## Changes

- `commands/repo/followups.md` step 3: `gh issue list --repo … --search … --json number,title,url`
  becomes `gh api "search/issues?q=repo:<slug>+state:open+<terms>&per_page=30" --jq '.items[] | "#\(.number) \(.title) \(.html_url)"'`.
  Search items already carry `number`/`title`/`html_url`, so the output shape is
  a straight replacement. Adds a note on the three distinct rate-limit buckets
  (`core` / `graphql` / `search`) and `gh api rate_limit --jq .resources`.
- `commands/repo/followups.md` step 5: `gh issue create --body "$(cat <<'EOF' … EOF)"`
  becomes: write body to a scratch file (agent file-write, not a heredoc), build
  the payload with `jq -n --arg t … --rawfile b … '{title,body,labels:[]}'`, then
  `gh api --method POST "repos/<slug>/issues" --input "$PAYLOAD" --jq '.html_url'`.
  Documents both motivations (rate-limit pool; body never re-enters the shell).
- Near-match comment path moved off GraphQL-backed `gh issue comment` to
  `POST repos/<slug>/issues/<n>/comments` with the same `--input` shape.
- New Safety Rule 5 records the REST-only constraint so the recipe cannot drift
  back to the `gh issue` forms.
- CHANGELOG Unreleased entry.

## Acceptance Criteria Verification

| Criterion | Status | Verification |
|-----------|--------|--------------|
[...]

PR #100, opened 2026-08-04
perf(tidy): prune heavy trees in inventory find walks instead of -not -path

### LINKED ISSUE #85: /repo:tidy inventory uses -not -path instead of -prune; find stalls on large repos

rjwalters:
## Summary

`/repo:tidy`'s step-1 inventory filtered heavy trees out of its two `find`
walks with `-not -path`, which only suppresses *printing* — `find` still
`readdir()`s into `.git/`, `node_modules/`, `target/`, and every other
"excluded" directory. On a large working tree that descent is the entire cost:
the reported case (~87 GB, a 24 GB Rust `target/`, ~30 live agent worktrees)
blew past a 120-second timeout, so the Tidy stage stalled exactly where an
inventory is most useful. Both walks now prune the heavy trees before the print
term so `find` never descends into them.

Secondary fix folded in per the Curator Enhancement: the pruned entries are
matched by `-name` rather than `-path`, which closes a latent coverage gap —
`-not -path './node_modules/*'` only ever matched a *top-level* `node_modules`,
never `packages/foo/node_modules/`.

## Changes

- `commands/repo/tidy.md` — rewrote the two step-1 `find` invocations (empty
  directories, >10 MB files) into the `\( … \) -prune -o … -print` form,
  pruning `./.git`, `node_modules`, `target`, `dist`, `.venv`.
- `commands/repo/tidy.md` — added a short prose block directly under the
  inventory commands recording that the prune list is a performance-only device
  that must not change what is reported, must stay identical across both walks,
  must draw from the denylist/CACHE categories already named in step 2, and
  deliberately excludes coordination roots (`.loom/`, `.anvil/`, `.wrangler/`)
  so step 2 still sees their empty directories to route to ASK.

Scope note: this PR touches **only** the two `find` lines and the prose
immediately following them. Line 52's `git worktree list` and the step-2/step-3
denylist and report sections (issue #86's span, being built concurrently
against the same file) are untouched. Expect a trivial rebase if #86 lands
first.

| Line 46 (`find . -type d -empty -not -path './.git/*'`) rewritten to use `-prune` | ✅ | See diff; executed against a fixture and this repo |
| Line 49 (`find . -type f -size +10M …`) rewritten the same way | ✅ | Same, identical prune list |
| Pruned set covers `.git/`, `node_modules/` at any depth, `target/`, `dist/`, `.venv/` | ✅ | `-path './.git'` for the root `.git`; `-name` for the rest, so nested copies prune too |
| Pruning does not change *what* is reported outside the pruned set | ✅ | Fixture run below: `./src/keepme-empty` and `./src/bigfile.bin` (11 MB) reported identically by old and new forms |
| Prune list documented as performance-only and told to track step 2's denylist/CACHE names | ✅ | New prose block under the inventory commands |
| No change to `commands/repo/audit.md` (N/A per Curator) | ✅ | `git diff --stat` shows only `CHANGELOG.md` + `commands/repo/tidy.md` |
| No rewrite of step-2 categorization logic | ✅ | Diff is confined to lines 46–49 plus the following prose |

## Test Plan

Automated: `pnpm test` → **PASS: 954  FAIL: 0** (matches baseline; `tidy.md` is
a prompt document with no executing harness — `test-verify-fix-persistence.sh`
explicitly excludes it from its scope guard).

Manual verification on a synthetic fixture (`.git/`, top-level `node_modules/`,
nested `packages/foo/node_modules/deep/`, `target/`, `dist/`, `.venv/`, plus
real junk under `src/`):

```
=== OLD empty-dir form ===          === NEW empty-dir form ===
./.venv/lib                         ./src/keepme-empty
./dist/out
./node_modules/top
./src/keepme-empty

=== OLD large-file form ===         === NEW large-file form ===
./packages/foo/node_modules/deep/blob   ./src/bigfile.bin
./src/bigfile.bin
./target/big/blob
```

- **Reporting outside the pruned set is unchanged**: `./src/keepme-empty`
  (empty dir) and `./src/bigfile.bin` (11 MB) appear in both forms.
- **Nested `node_modules/` is now pruned** — `packages/foo/node_modules/deep/blob`
  was reported by the old form and is not by the new one. This is a deliberate
  *coverage* change, not just a perf one: the old `-not -path './node_modules/*'`
  never matched below the top level. Nested `node_modules` is a denylisted
  environment that `git clean -ndX` surfaces on its own, so it is not lost from
  the inventory as a whole.
- **Traversal is actually avoided, not just filtered**: with 20,204 nodes under
  `target/`, directories visited drops from 217 to 5, and wall time from
  `real 0.03` to `real 0.00`. The pruned directories' contents are never
  visited.

Closes #85


rjwalters:
✅ **Approved!**

Reviewed the full diff against issue #85's Curator Enhancement. The change matches the curator's suggested form almost exactly, the prose block is the strongest part of the PR, and I independently reproduced the builder's fixture verification rather than taking it on faith.

## Independent verification

I rebuilt an equivalent fixture (`.git/`, top-level `node_modules/`, nested `packages/foo/node_modules/deep/`, `target/`, `dist/`, `.venv/`, an empty `.loom/locks/`, a nested `sub/.git/`, plus real junk under `src/`) and ran both the old and new forms. Results:

**Empty-directory walk**
```
=== OLD ===                    === NEW ===
./.loom/locks                  ./.loom/locks
./.venv/lib                    ./build
./build                        ./packages/foo/src
./dist/out                     ./src/keepme-empty
./node_modules/top
./packages/foo/src
./src/keepme-empty
```

**Large-file walk**
```
=== OLD ===                                     === NEW ===
./.loom/worktrees/issue-9/node_modules/x/blob   ./src/bigfile.bin
./packages/foo/node_modules/deep/blob           ./sub/.git/objects/pack
./src/bigfile.bin
./sub/.git/objects/pack
./target/big/blob
```

Three things I specifically wanted to confirm, all of which hold:

1. **Reporting outside the pruned trees is genuinely unchanged.** `./build`, `./packages/foo/src`, `./src/keepme-empty`, `./src/bigfile.bin` are all reported identically. Run against **this repo's actual working tree**, both walks `diff`'d clean — byte-identical output old vs. new for both the empty-dir and large-file forms.
2. **The `.loom/` carve-out works.** An empty `./.loom/locks` survives pruning and is still reported, so step 2 can route it to ASK — which is exactly what #88/PR #91 just established and would have been silently undone by a naive prune list. Good catch by the builder to call this out explicitly in the prose.
3. **The real-world win is the nested case, not the top-level one.** `./.loom/worktrees/issue-9/node_modules/x/blob` is pruned by the new form and was not by the old one. That is precisely the reported 87 GB / ~30-worktree scenario: the worktree roots stay unpruned (correctly), but the heavy trees *inside* them no longer get walked. `-name` rather than `-path` is what makes that work, and the PR is right to flag it as a coverage change rather than a pure perf change.

Traversal is actually avoided, not just filtered: directory visits on the fixture drop from 27 to 14, and the pruned subtrees' contents are never `readdir()`'d.

## Notes on the deliberate behavior changes

Two things the new form no longer reports, both of which I consider correct and within the AC:

- **Empty directories inside pruned trees** (`./dist/out`, `./node_modules/top`, `./.venv/lib`) are no longer listed. Impact is negligible — these are CACHE/denylist trees that `git clean -ndX` surfaces wholesale, and an orphaned empty dir inside `dist/` is not clutter worth a traversal budget.
- **Large files inside `target/`/`dist/`** are no longer listed as large files. This one is a real (small) loss of signal for "what's eating my disk", but the curator explicitly reasoned about it and asked for `dist`/`target` to be pruned anyway, and gitignored build trees still surface via `git clean -ndX`. Correct call.

The prose's framing — "traversal optimization only and must never change what is reported" — is scoped correctly by the very next clause ("junk *outside the pruned trees*"), so it isn't overclaiming. The four-bullet editing contract (keep the lists identical / draw from step 2's names / `-name` not `-path` / coordination roots stay unpruned) is the right thing to write down; the "keep both lists identical" bullet in particular is the invariant most likely to rot, and naming it is what stops a future edit from reintroducing the stall in one of the two commands.

## Checks

- **CI**: ✅ `pnpm test (guard/handoff/loss-check suite)` — success (the repo's only workflow). Verified via REST on head `f05bafd`.
- **Local `pnpm test` in the worktree**: **PASS: 954, FAIL: 0** — matches the builder's claim and the baseline exactly.
- **Merge state**: `mergeable: true`, `mergeable_state: clean`. #99 landed on main between PR creation and review and also touched `CHANGELOG.md`, but the two entries sit at different offsets in the Unreleased list — `git merge-tree` against current `origin/main` is conflict-free. **No rebase needed**; leaving the branch as-is so the passing CI run stays valid.
- **Scope**: diff is confined to `CHANGELOG.md` and `commands/repo/tidy.md` lines 46–79. Line 56's `git worktree list` is untouched, as are step 2/3 — so #86 / PR #101's span is clear and that PR should rebase cleanly on top of this one. `commands/repo/audit.md` correctly untouched (N/A per the Curator). Grepped the tree for other copies of the `-not -path` inventory pattern: none outside `tidy.md`.
- **Issue linking**: "Closes #85" present.
- **CHANGELOG**: entry is accurate and honest about the nested-`node_modules` coverage change rather than selling it as pure perf.

## Follow-up filed

Filed #104 for the one residual gap I found: `-path './.git'` prunes only the **root** `.git`, so a nested `.git` *directory* (a vendored clone, a fixture checkout) is still fully descended into — visible above as `./sub/.git/objects/pack` surviving into the new form's output. Not a regression (the old `-not -path './.git/*'` had the identical blind spot) and not blocking, but it's the single case where the prune list doesn't do what this PR's own "`-name`, not `-path`" rule says, and `.git` is usually the largest tree in a checkout. `-name .git` would close it with no downside — for worktrees and submodules `.git` is a file, and `-prune` on a non-directory is a harmless no-op. Deliberately kept out of this PR to preserve its tight scope.

Nice work — the empirical verification in the PR description held up under independent reproduction, which is not something I get to say often.

PR #101, opened 2026-08-04
docs(tidy): denylist git worktree roots and report their size

### LINKED ISSUE #86: /repo:tidy denylist should name git worktree roots explicitly

rjwalters:
## Summary

`/repo:tidy`'s never-delete denylist protected `.loom/worktrees/` only as a side effect of the `.loom/` tool-scaffolding prefix added in #88/PR #91. Nothing was worktree-*specific*, so `.claude/worktrees/` (3.1 GB in the reporter's repo), sibling worktrees (`git worktree add ../repo-wt-fix123`), and worktrees under `/private/tmp/` or `/private/var/folders/` had no protection of their own — and worktrees outside the repo root never appeared in the inventory at all, so tidy could report "nothing to do" on a 94 GB tree holding 66 GB of worktrees.

This adds tool-agnostic worktree-root detection wired into both categorization and reporting. Strictly report-only: deciding which worktrees are *stale* still belongs to `/repo:reset`, and deciding which are *reclaimable* needs merge state and agent liveness that tidy deliberately does not acquire.

## Changes

- **Step 1 (Inventory)**: `git worktree list` → `git worktree list --porcelain`, with the output explicitly **retained** as a path set for step 2 rather than printed for the `/repo:reset` pointer and discarded. Adds a per-root `du -sh` for the report, a note to drop the first (main working tree) entry, and the `.git`-**file** `gitdir: …/.git/worktrees/…` fallback — with porcelain documented as authoritative when the two disagree (stale pointer edge case).
- **Step 2 (Categorize)**: new "Git worktree roots" bullet in the never-delete denylist, routing to ASK, never SAFE/CACHE, independent of gitignore status and of which dot-directory (if any) the worktree lives under. Notes that `git clean -ndX` shows these as `Would skip repository <path>` and that `-ffdX` deletes them, so the force flag must never be widened.
- **Step 3 (Report)**: new `WORKTREES` inventory block, distinct from the generic denylist ASK lines, with per-root sizes summed **separately** from SAFE/CACHE/ASK totals and printed even on an otherwise-clean run. Optional soft pointer to repo-owned worktree tooling (e.g. `loom-daemon clean --safe --dry-run`) when the repo documents one — omitted rather than guessed.
- **Step 4 (Apply)** and **safety rule 6** name worktree roots alongside the other denylisted paths; **new safety rule 9** records the reasoning.
- CHANGELOG `Unreleased` entry.

| Step 1 uses `git worktree list --porcelain` and its output is referenced later | ✅ | Command changed; inline comment says "RETAIN this output as a path set for step 2's denylist"; the step-2 bullet and step-3 report both reference it |
| Denylist gains explicit "git worktree roots" bullet via porcelain path set and/or `.git`-file pointer | ✅ | Both mechanisms documented in the new bullet; verified both work against a scratch repo (see Test Plan) |
| Documented as routing to ASK (never SAFE/CACHE), consistent with denylist-first ordering and Safety Rule 8 | ✅ | Bullet states "never SAFE and never CACHE, regardless of gitignore status" and cites rule 8 for the empty-directory carve-out; also added to the ASK category list |
| Report gains a worktree-specific line item with path and size, distinct from generic denylist ASK lines | ✅ | New `WORKTREES` block in the step-3 example, with bytes summed separately and printed even when nothing is deleted |
| `.claude/worktrees/`-style worktrees (outside #91's prefix list) explicitly covered | ✅ | Named in the bullet and in rule 9; scratch-repo run confirms both detection paths fire on `.claude/worktrees/scratch-wt` |
| Existing behavior unchanged for non-worktree paths | ✅ | Additive bullet + report block; SAFE/CACHE allowlists, `--caches` semantics, and the #91 tool-scaffolding rules are untouched |

## Test Plan

Detection verified against a scratch repo with `.claude/` gitignored and two worktrees (`.claude/worktrees/scratch-wt`, sibling `../sibling-wt`):

- `git worktree list --porcelain` lists the main tree plus both worktrees, including the one outside the repo root (confirming the "drop the first entry" note is needed).
- `git worktree list --porcelain | awk '/^worktree /{print $2}' | xargs -I{} du -sh {}` produces the per-root sizes the report block consumes.
- The `.git`-file test (`[ -f "$d/.git" ] && grep -Eq 'gitdir: .*/\.git/worktrees/.*'`) fired on **both** worktrees, including the sibling that no prefix rule reaches.
- `git clean -ndX` printed `Would skip repository .claude/worktrees/scratch-wt`, confirming the `-ffdX` caveat now documented in the bullet.

`pnpm test`: **954 pass / 0 fail** (baseline unchanged — `commands/repo/tidy.md` is a skill prompt and is deliberately out of scope for `commands/repo/tests/`, per the scope guard at `test-verify-fix-persistence.sh:321-327`).

Closes #86


rjwalters:
❌ **Changes Requested**

Substantively this is the right shape and it clears every acceptance criterion in #86's Curator Enhancement section. The detection claims all reproduce. But three things in step 1 are demonstrably wrong or self-defeating, and one of them regresses the fix that merged an hour ago in #100 — in the same step of the same file.

## Rebase (done by me)

`feature/issue-86` was based on `bc9fe24` and went DIRTY when #100 (`ce0e0b1`) and #99 landed. Both conflicts were pure both-adds, so I rebased in the worktree and force-pushed (`68f1862` → `cbf2867`):

- **CHANGELOG.md** — #85's and #86's `Unreleased` bullets, both kept, #85 first.
- **commands/repo/tidy.md** — #100's prune-list prose kept ahead of this PR's worktree prose (matches the command order inside the fence); the two `find` invocations auto-merged to #100's `-prune` form. No content from either side dropped (`grep 'Coordination roots'` still present).

`pnpm test` in the worktree: **954 pass / 0 fail**. CI on `cbf2867`: green. `mergeable_state`: `clean`. Nothing below is a merge-conflict artifact.

## Verification I ran

Scratch repo, `.claude/` gitignored, worktrees at `.claude/worktrees/scratch-wt` and a **sibling `../sibling wt`** (name deliberately containing a space):

| Claim | Result |
|---|---|
| porcelain lists out-of-root worktrees | ✅ all three entries, including the sibling |
| `.git`-file regex `gitdir: .*/\.git/worktrees/.*` | ✅ fired on **both**, including the sibling no prefix rule reaches |
| `git clean -ndX` → `Would skip repository <path>` | ✅ exact wording reproduced |
| the `du` pipeline in the fence | ❌ **broke** — see 2 below |
| "first entry == `git rev-parse --show-toplevel`" | ❌ **false from a linked worktree** — see 3 below |

## 1. The per-root `du -sh` reintroduces exactly the traversal #100 just removed (blocking)

`commands/repo/tidy.md:62-65`:

> Bounded by the number of worktrees (typically a handful), so one `du -sh` per root is cheap enough — unlike a `du` over the whole tree, which is not.

That rationale does not hold on the repo this PR is motivated by. Your own CHANGELOG entry says the tree is 94 GB with **66 GB sitting in worktrees** — so `du` over the worktree roots *is* ~70% of a `du` over the whole tree. "Bounded by the number of worktrees" is the wrong bound; the cost is bounded by inode count under those roots, and #100's entry records ~30 worktrees plus a 24 GB Rust `target/`.

Worse, `du` has no prune. Sitting eight lines below it are two `find` walks that #100 rewrote specifically so they'd stop descending into `node_modules`, `target`, `dist`, and `.venv` — because that descent blew past a **120-second timeout** and stalled the Tidy stage on this exact repo. The new command re-enters every one of those trees, once per worktree, in the same step. I did not benchmark a 66 GB tree, but the structural argument doesn't need a benchmark: this is the traversal #100 measured as fatal, multiplied by N worktrees.

It also silently violates #100's adjacent prose, which is now directly above it: *"Keep both invocations' lists identical"* and *"Draw entries from the denylist and CACHE categories already named in step 2 rather than growing a second, inconsistent list."* This is a third walk with no list at all.

There's a genuine design tension here — pruning `node_modules` out of `du` would understate the footprint, and the whole point of the block is showing the operator where 66 GB went. So I'm not prescribing the fix, but it needs an explicit answer rather than a "cheap enough" claim that isn't true. Options that seem reasonable:

- Bound it: `timeout 20 du -sh <root>` per root, printing `size unavailable (timed out)` rather than stalling the step.
- Or state the cost honestly and make the size column best-effort/skippable, with the count and paths always printed (the count alone already fixes the "nothing to do" failure the follow-up comment on #86 describes).
- Or reuse the prune list and label the number as "excluding regenerable trees", which is defensible but changes what the number means and should say so.

Whichever you pick, please replace the "cheap enough" sentence — as written it will teach an agent to run an unbounded walk in the step that already has a recorded timeout.

## 2. The `du` pipeline drops any worktree whose path contains a space (blocking, one-line fix)

`commands/repo/tidy.md:65`:

```bash
git worktree list --porcelain | awk '/^worktree /{print $2}' | xargs -I{} du -sh {}
```

`awk '{print $2}'` truncates at the first space, and the unquoted `{}` re-splits. Observed in the scratch repo:

```
du: /…/scratchpad/wtcheck/sibling: No such file or directory
```

The sibling worktree is `…/sibling wt`. It silently vanishes from the size listing — and sibling/ad-hoc worktrees are this PR's headline case, so the failure lands squarely on the paths the rest of the doc is written to protect. Porcelain's `worktree ` prefix is fixed-width, so:

```bash
git worktree list --porcelain | sed -n 's/^worktree //p' | while IFS= read -r wt; do du -sh "$wt"; done
```

## 3. "The first entry is the main working tree (`git rev-parse --show-toplevel`)" is false from inside a linked worktree (blocking, small)

`commands/repo/tidy.md:89-91` equates two things that diverge. From inside `.claude/worktrees/scratch-wt`:

```
show-toplevel: /…/wtcheck/main/.claude/worktrees/scratch-wt
first porcelain entry: /…/wtcheck/main
```

The first porcelain entry is always the **main** working tree, never the cwd. `git rev-parse --show-toplevel` is the cwd's root. This matters concretely here: Loom agents run commands from inside `.loom/worktrees/issue-N`, so `/repo:tidy` invoked by an agent drops the main repo root from the list and then reports its **own cwd** as a linked worktree to be protected. The protection direction is safe-fail (over-protective, not under-protective), so this isn't a data-loss risk — but the report will be confusing and the stated identity is wrong.

Suggest dropping the entry whose path equals `git rev-parse --show-toplevel` (i.e. "the tree being tidied"), and noting separately that the first porcelain entry is the main working tree, which may or may not be the same thing.

## Non-blocking nits

- `commands/repo/tidy.md:223-228` — the `WORKTREES` example columns don't line up (`32 GB` / `3.1 GB` / `12 GB` / `19 GB`), unlike the neatly aligned SAFE/CACHE/ASK blocks right above. Cosmetic, but this block is the template an agent will imitate.
- `CHANGELOG.md` — "the tool-scaffolding prefixes added in #88" where the adjacent bullet uses the fuller "(#88, PR #91)" form. Worth matching.

## What's good

Genuinely: the `-ffdX` caveat in the denylist bullet is the most valuable line in the diff and I verified the `Would skip repository` wording it rests on. Rule 9's "do not infer protection from a dot-directory prefix" is the correct framing and is exactly the gap #91 left. Keeping `/repo:reset` as the owner of the stale/reclaimable decision — and saying so twice, in the ASK section and the report block — is the right scope discipline, and the "omit the tooling pointer rather than guessing" instruction is a nice piece of restraint. Fix 1–3 and this should go straight through.

The branch is rebased, pushed, and green — please build on `cbf2867`.

rjwalters:
✅ **Feedback addressed** — all three blocking findings fixed in `91d61b0` (on top of your rebase `cbf2867`).

Thanks for the scratch-repo verification; I reproduced both failures before changing anything (`.../sibling wt` → `awk` prints `.../sibling`; `--show-toplevel` from inside a linked worktree ≠ first porcelain entry).

## 1. Sizing is now opt-in behind `--sizes`, and bounded

**Option picked: eager `du` removed from the default inventory; count + paths always printed; sizes opt-in via a new `--sizes` flag, and best-effort/bounded when it is passed.**

Why this one, out of the three you listed:

- **`--caches` is the existing precedent.** tidy already has a tier whose *result* is useful but whose *cost* is high, and the command's answer there is "present it, don't perform it, until asked." Cache deletion is cheap and safe but forces a rebuild → opt-in. A worktree `du` is safe but costs a full traversal → same shape, same answer. Bounding alone (option A) would have left an unbounded-by-default walk gated on a timeout, which is a mitigation, not a design.
- **The count and paths carry the signal the block exists for.** The failure this PR is fixing is `/repo:tidy` reporting "nothing to tidy" on a 94 GB tree. `git worktree list` reads `.git/worktrees/` — it does not walk anything — so `WORKTREES (4 roots …)` plus every path is free and still fixes that failure completely. The sizes quantify it; they are not the point. That is what made dropping them from the default cheap rather than a regression of the PR's own goal.
- **I did not take the prune option.** As you noted, an "excluding regenerable trees" number understates exactly the footprint the operator is hunting for, and it would also make the WORKTREES total mean something different from the SAFE/CACHE/ASK totals next to it. The doc now says so explicitly: *"The honest choices are a bounded full number or none."*
- **Belt and braces:** when `--sizes` *is* passed, each root is wrapped in `timeout 20` and prints `size unavailable` instead of hanging the step; unavailable roots are excluded from the total and the total is marked a lower bound. Noted that `timeout` is GNU coreutils (`gtimeout` on macOS) and that the correct fallback is "report unavailable," never "run it unbounded."

The "cheap enough / bounded by the number of worktrees" sentence is gone, replaced by the inode-count argument and an explicit pointer back to the 120-second timeout #100 recorded — so the doc now argues *against* the walk in the same place it used to invite it.

## 2. Space-safe path extraction

```bash
git worktree list --porcelain | sed -n 's/^worktree //p' \
  | while IFS= read -r wt; do
      timeout 20 du -sh "$wt" 2>/dev/null \
        || printf '%s\t%s\n' 'size unavailable' "$wt"
    done
```

Verified against a scratch repo with a sibling `…/sibling wt`: `sed` + `while IFS= read -r` keeps all three roots, `awk '{print $2}'` still truncates to `…/sibling`. The plain path listing (used by the default, size-free report) uses the same `sed` form, and I noted that paths containing *newlines* need `git worktree list --porcelain -z`.

## 3. The `--show-toplevel` identity

Removed. The doc now states that the first porcelain entry is the **main** working tree *always, regardless of where the command runs from*, and that this is **not** necessarily `git rev-parse --show-toplevel` — with the Loom case (`.loom/worktrees/issue-N`) named as the concrete divergence. Per your suggestion, what gets dropped from the list is **the tree being tidied** (`--show-toplevel`), so every remaining entry — including the main working tree when tidy runs from a linked worktree — is reported and protected.

## Non-blocking nits

- WORKTREES example columns aligned (arrows now at a fixed column in both the default block and the `--sizes` variant; sizes right-aligned so `3.1 GB` lines up under `32 GB`).
- CHANGELOG: inline `#88` → `#88 (PR #91)` to match the adjacent bullet; the bullet's own header is now `(#86, PR #101)`, and its second half rewritten to describe the `--sizes` design rather than the removed eager `du`.

`pnpm test`: **954 pass / 0 fail**. Pushed to `feature/issue-86` (no force — built directly on `cbf2867`). Re-labeled `loom:review-requested`.

rjwalters:
✅ **Approved!**

All three blocking findings are genuinely resolved, and the fix for #1 is better than what I suggested. Re-verified `91d61b0` against the same fixtures I used to find the bugs.

## Finding 1 — eager `du` — resolved, and resolved well

The `--sizes` flag is the right call, and the reasoning in `commands/repo/tidy.md:104-127` now states the cost accurately instead of asserting it away: cost bounded by inode count rather than worktree count, `du` has no `-prune`, sizing the roots is most of a whole-tree walk on the motivating repo, in the step with the recorded 120 s timeout.

Two judgments in there I want to endorse explicitly, because both were live options I left open and the Doctor picked correctly:

- **Refusing to prune inside the `du`.** *"An 'excluding regenerable trees' number would understate exactly the footprint the operator is looking for. The honest choices are a bounded full number or none."* That is right — I had listed the pruned variant as acceptable-if-labelled, and declining it is the better answer.
- **Keeping count + paths unconditional.** This is what preserves the actual fix for #86's follow-up complaint. The failure being prevented is "tidy says nothing to do on a 94 GB tree" — the count and paths carry that signal and cost nothing (`git worktree list` reads `.git/worktrees/`, it does not walk the trees; confirmed). The size only quantifies it.

The `--caches` precedent framing — present the expensive thing, don't perform it until asked — is exactly the right analogy and makes the flag feel native rather than bolted on. The `timeout 20` bound with a `size unavailable` fallback, unavailable roots excluded from the total, and the total marked as a lower bound all hang together. The `gtimeout`/`brew install coreutils` note is warranted: `timeout` on this box is `/opt/homebrew/bin/timeout`, i.e. absent on a stock macOS install, and the documented degradation (report unavailable rather than run unbounded) is the safe direction.

## Finding 2 — space-safe paths — resolved, verified

Same `"…/sibling wt"` fixture that broke the old command:

```
$ git worktree list --porcelain | sed -n 's/^worktree //p'
…/wt2/main
…/wt2/main/.claude/worktrees/scratch-wt
…/wt2/sibling wt          ← intact

$ …| while IFS= read -r wt; do timeout 20 du -sh "$wt" …
 12K  …/wt2/sibling wt     ← sized correctly

# old command, for contrast:
du: …/wt2/sibling: No such file or directory
```

`^worktree ` can't collide with the other porcelain keys (`HEAD`, `branch`, `bare`, `detached`, `locked`, `prunable`), so the `sed` is safe as well as correct. The `--porcelain -z` pointer for newline-containing paths checks out on git 2.55.

## Finding 3 — `--show-toplevel` identity — resolved, verified

The prose now names the divergence instead of asserting the identity, and the drop rule changed from positional to "the entry whose path equals `git rev-parse --show-toplevel`". Run from inside `.claude/worktrees/scratch-wt`:

```
show-toplevel = …/main/.claude/worktrees/scratch-wt
KEEP …/wt2/main
KEEP …/wt2/sibling wt
```

The cwd is dropped, the main working tree is kept — matching the doc's own claim that the main tree "is a live checkout to report and protect" when tidy runs from a linked worktree. That's the correct behavior for the Loom case and it's now what the doc describes.

## Nits — both fixed

Arrow columns are consistent within each block now (col 42 default, col 50 with sizes) and the size column is right-aligned across `32 GB` / `3.1 GB` / `12 GB` / `19 GB`. CHANGELOG uses `(#86, PR #101)` and `#88 (PR #91)` consistently.

## One thing for the record, not a blocker

Issue #86's acceptance criterion reads *"the report format (step 3) gains a worktree-specific line item showing path **and size**."* With `91d61b0` the size is opt-in, so the default report satisfies that criterion in spirit but not literally. I'm calling it satisfied, and flagging it so the narrowing is on the record rather than discovered later:

- The narrowing is a direct consequence of my own blocking finding — the literal reading requires an unbounded `du` in a step with a recorded timeout, which would have regressed #85/#100 the same day it merged.
- The failure mode the criterion exists to prevent is fully fixed: the operator sees `WORKTREES (4 roots …; --sizes to measure)` on an otherwise-clean run, so the space is never invisible.
- The number is one bounded flag away, and the flag is discoverable from the block's own header.

Also noting deliberately: `--sizes` is **not** plumbed through `/repo:all` the way `--caches` is (`commands/repo/all.md:28`). I think that's correct — an omnibus run is the last place you want an unbounded `du` — so no change requested, just confirming it reads as intentional rather than forgotten.

## Verification

- `pnpm test` in the worktree on `91d61b0`: **954 pass / 0 fail**
- CI: `pnpm test (guard/handoff/loss-check suite)` — success
- `mergeable_state`: `clean` (rebase onto `ce0e0b1` still holds; no new drift)
- Scratch-repo re-checks of findings 2 and 3 as shown above

Nice work on the turnaround — the `--sizes` rationale is now the clearest part of the file.

Issue #81, opened 2026-07-31
release: no detection path for versions kept in source constants; scaffold package.json misdirects to npm


rjwalters:
## Context

Cutting a release in rjwalters/claude-monitor with /repo:release, Phase 2 detection provisionally landed on `npm` solely because a `package.json` exists at the repo root — but that file is Loom workspace scaffolding (`"name": "loom-workspace"`, no `version` field at all). The repo's real version-bearing files are source constants: `AppVersion.current` in a Swift file plus two CFBundle strings in the build script, as documented in its CLAUDE.md.

The existing cross-source reconciliation only covers the `package.json` + root `VERSION` coexistence case, so nothing in the command flags this misdirection; the operator had to notice from CLAUDE.md context and take the manual [m] path.

## Suggested acceptance criteria

- [ ] A `package.json` without a `version` field (or otherwise clearly a workspace scaffold) does not win npm detection
- [ ] Repos can declare arbitrary version-bearing files (e.g. via `.repo/release-policy.md` — a `pre-apply (replace)` seam works today but detection/drift-gate/verify still assume a known tool; a first-class declaration would cover all three)
- [ ] The manual path documents how the version was found (CLAUDE.md context) so the next release doesn't rediscover it

---

### AC item 1 is already resolved on `main` — no work remains here

PR #105 (`aa65cb7`, merged for issue #83) tightened the Phase 2 `npm` branch in
`commands/repo/release.md` while adding PEP 621 `pyproject.toml` detection.
Current `main`:

```bash
elif command -v npm >/dev/null 2>&1 && [ -f package.json ] && grep -q '"version"' package.json; then
  VERSION_TOOL="npm"; WHY="npm + package.json"
```

A version-less scaffold `package.json` (`{"name": "loom-workspace"}`, no
`"version"` key) now fails this branch and falls through the `elif` chain to
`VERSION_TOOL="version-file"` (if a root `VERSION` exists) or to the `[none
detected]` → manual-path prompt — it can no longer misdirect detection to
`npm`. The `## Cross-source reconciliation` section and the accompanying prose
were updated to match ("`npm` is matched on any `package.json` carrying a
`version` field…"). **This issue's scope narrows to AC items 2 and 3 below —
do not re-open AC item 1.**

See the curator cross-reference already on this issue
(2026-08-04, from the #83 curation pass) for the original reasoning.

### Remaining scope: a first-class declaration mechanism for source-constant versions

The claude-monitor repro repo keeps its real version in source constants no
[...]

PR #105, opened 2026-08-04
feat(release): detect PEP 621 [project].version ahead of the npm fallback

### LINKED ISSUE #83: release: detect PEP 621 pyproject [project].version before the npm fallback
## Context

Hit during a real `/repo:release` run in a Python repo managed with uv/hatchling (PEP 621 `[project] version = "..."` in `pyproject.toml`, no poetry). The detection chain has no branch for PEP 621 pyproject versions — only `[tool.poetry]` — so detection fell through to `npm` because a `package.json` existed. That `package.json` was a Loom workspace stub: `"private": true`, **no `version` field at all**. A blind `npm version` would have written a version into the wrong file. The operator had to override to a manual pyproject bump (plus `uv lock` refresh so `uv sync --locked` CI didn't break).

Two gaps, either of which would have prevented the false positive:

1. No detection branch for PEP 621: `pyproject.toml` with a `[project]` `version` key (the uv/hatchling/flit/setuptools-modern standard) should be detected ahead of the npm fallback — e.g. as a `pyproject` tool whose apply path edits `[project].version`, commits, and tags (and, when a `uv.lock` exists, runs `uv lock` so locked CI installs don't go stale).
2. The npm match is presence-only: a `package.json` **without a `version` field** cannot be the version source and should be a non-match, letting detection continue down the chain.

- [ ] Detection recognizes `pyproject.toml` `[project].version` (non-poetry) as a version tool before the npm fallback.
- [ ] `package.json` lacking a `version` field no longer matches as `npm`.
- [ ] The pyproject apply path regenerates `uv.lock` when present (mirroring the cross-source care the VERSION/package.json reconciliation already takes).

---

### Problem statement

`/repo:release` Phase 2 (`commands/repo/release.md`) walks a fixed `if/elif` chain
to pick `VERSION_TOOL`, in this order: `version.sh` → `cargo-release` →
`cargo-set-version` → `cargo-workspace` → `bumpversion` → `bump2version` →
`poetry` (`[tool.poetry]` only) → `npm` (`package.json` presence only) →
`version-file` (root `VERSION`).

Two related bugs compound in PEP 621 repos (uv/hatchling/flit/setuptools —
i.e. the modern non-poetry standard `[project] version = "..."` key):

1. **No PEP 621 branch exists at all.** The only pyproject-aware branch checks
   for `[tool.poetry]`. A `pyproject.toml` with `[project].version` and no
   `[tool.poetry]` section falls through it entirely.
2. **`npm` is matched on file presence, not on having a version.** `[ -f
   package.json ]` alone triggers the `npm` branch (`commands/repo/release.md:160-161`),
   even when `package.json` has no `"version"` key — which is exactly the
   shape of a Loom workspace-scaffold `package.json` (`"private": true`, no
   version field). Because the PEP 621 branch is missing (bug 1) and doesn't
   exist to be tried first, this npm-presence-only bug is what actually fires
   in the field: an unrelated scaffold file wins detection over the real
   version source. The existing *Cross-source reconciliation* section (lines
   172–210) only reconciles `npm` against a root `VERSION` file — it never
   even considers pyproject, so it does not catch this case.

A blind `npm version` in this state would write a version into the
scaffold `package.json` (starting from `0.0.1` since it has no existing
version) while the real `pyproject.toml` version stays stale — the exact
mis-bump the operator had to catch and route around manually, including a
follow-up `uv lock` to keep `uv sync --locked` CI from breaking.

### Implementation guidance

**Where**: `commands/repo/release.md`, Phase 2 detection chain (lines
135–166) and the "Cross-source reconciliation" subsection (lines 172–226),
plus the Phase 5 apply dispatch (lines 298–345).

**1. Add a PEP 621 detection branch**, ordered directly before the `npm`
branch (so it wins whenever both a `pyproject.toml` `[project].version` and
an unrelated `package.json` coexist — the exact scenario reported here). Keep
the existing `poetry` branch checking `[tool.poetry]` first (poetry files
also have a `[project]` table in modern releases, so branch order matters —
poetry must still win when both markers are present, since `poetry version`
is the correct apply path for those repos, not a raw TOML edit):

```bash
elif command -v poetry >/dev/null 2>&1 && [ -f pyproject.toml ] && grep -q '\[tool.poetry\]' pyproject.toml; then
  VERSION_TOOL="poetry"; WHY="poetry + [tool.poetry]"
elif [ -f pyproject.toml ] && grep -qE '^\[project\]' pyproject.toml && \
     awk '/^\[project\]/{f=1;next} /^\[/{f=0} f' pyproject.toml | grep -qE '^version[[:space:]]*='; then
  VERSION_TOOL="pyproject"; WHY="pyproject.toml [project].version (PEP 621)"
elif command -v npm >/dev/null 2>&1 && [ -f package.json ] && grep -q '"version"' package.json; then
  VERSION_TOOL="npm"; WHY="npm + package.json"
elif [ -f VERSION ]; then
  VERSION_TOOL="version-file"; WHY="plain VERSION file at repo root"
fi
```

Notes on the `[project].version` check: it deliberately scopes the
`version[[:space:]]*=` grep to lines *within* the `[project]` table (via the
`awk` block-extractor) rather than matching anywhere in the file, so a
`[tool.poetry.dependencies]` block or similar with an unrelated `version =`
line doesn't false-positive the branch. This mirrors the same
false-positive risk the issue calls out for `npm`.

**2. Fix the npm presence-only match** (item 2 above) — the diff already
shown folds this in: add `&& grep -q '"version"' package.json` to the `npm`
elif condition. This is the general fix; a `package.json` without a
`version` field never wins `npm` detection regardless of what other tool
does or doesn't match, so it correctly falls through toward `version-file`
or "no tool detected" instead of silently misdirecting.

**3. Add a `pyproject` apply-path branch** to the Phase 5 dispatch
(`commands/repo/release.md:315-325`):

```bash
pyproject)
  # Edit [project].version in place, then commit + tag (mirrors version-file).
  python3 - "$NEW" <<'PY'
import re, sys
path = "pyproject.toml"
new = sys.argv[1]
text = open(path).read()
text = re.sub(
    r'(\[project\][^\[]*?\nversion\s*=\s*)"[^"]+"',
    lambda m: m.group(1) + f'"{new}"',
    text, count=1,
)
open(path, "w").write(text)
PY
  [ -f uv.lock ] && command -v uv >/dev/null 2>&1 && uv lock
  ;;  # then stage pyproject.toml (+ uv.lock if regenerated) + CHANGELOG.md, commit, tag v$NEW
```

Add `pyproject` to the "tools that don't self-commit" list alongside
`cargo-set-version`, `cargo-workspace`, `poetry`, `version-file` (line
328-329), since this branch, like `version-file`, only edits the file — it
does not commit or tag itself.

Also add a "read the current version" example for `pyproject` next to the
existing per-tool read examples in Phase 3 (line 235-236), e.g. a `python3 -c
"import tomllib; print(tomllib.load(open('pyproject.toml','rb'))['project']['version'])"`
one-liner (Python 3.11+; repos on older Python can fall back to a `grep`/`sed`
one-liner scoped to the `[project]` block, matching the same style as the
existing `poetry version -s` / `node -p …` examples).

**4. `uv.lock` regeneration** (item 3 in the suggested AC) is folded into the
apply-path snippet above via `[ -f uv.lock ] && uv lock` — gated on `uv`
being on `PATH` so a non-uv PEP 621 repo (flit/setuptools-only, no lock file)
doesn't fail Phase 5 apply.

### Relationship to #81

#81 ("no detection path for versions kept in source constants; scaffold
package.json misdirects to npm") reports the **same root failure mode** —
detection landing on `npm` solely because a scaffold `package.json` exists —
but from a **different, disjoint direction**:

- **#83 (this issue)**: the *correct* answer is discoverable via a **known,
  parseable file format** — PEP 621 `pyproject.toml` `[project].version`.
  Fix = add a new first-class detection branch for that format, ordered ahead
  of `npm`, plus tighten the `npm` match to require a `version` field.
- **#81**: the *correct* answer lives in **arbitrary source constants**
  (e.g. `AppVersion.current` in a Swift file, CFBundle strings in a build
  script) that no generic parser can discover — #81's own suggested AC
  explicitly asks for a **declaration mechanism** (e.g. via
  `.repo/release-policy.md`) rather than a new autodetection branch, plus
  documenting how the manual path found the version so it isn't rediscovered
  each release.

**Shared fix, independent scopes**: both issues independently motivate
tightening the `npm` presence-only match (`[ -f package.json ]` →
`[ -f package.json ] && grep -q '"version"' package.json`) — that one-line
change is included in #83's implementation guidance above since it's a
direct, minimal, already-scoped fix. If #83 lands first, #81's remaining
scope narrows to just the declaration-mechanism ask (its AC item 2) and the
manual-path documentation (its AC item 3); its AC item 1 ("package.json
without a version field does not win npm detection") will already be
satisfied. Do not duplicate the `npm`-match fix in both PRs — whichever
lands first should make the other's AC item 1 a no-op verification rather
than a second edit.

No sequencing dependency: #83 does not block on #81's declaration mechanism
(PEP 621 files are always parseable, so #83 needs no new declaration seam),
and #81's source-constants problem has no bearing on the pyproject-specific
branch #83 adds.

### Affected Files

- `commands/repo/release.md` - Phase 2 detection chain: add `pyproject` (PEP 621) branch before `npm`, tighten `npm` branch to require a `version` field (lines ~135-166); Phase 3: add a `pyproject` "read current version" example (~line 235-236); Phase 5 apply dispatch: add `pyproject)` case + add it to the "tools that don't self-commit" list (~lines 298-331).

### Test Plan

- [ ] Manual verification (PEP 621, no poetry): in a scratch dir, create `pyproject.toml` with `[project]\nname = "x"\nversion = "1.2.3"` and no `[tool.poetry]`; also drop in a scaffold `package.json` with `{"private": true}` (no `version` key) to simulate the reported false positive. Run the Phase 2 detection snippet standalone and confirm it prints `pyproject — pyproject.toml [project].version (PEP 621)`, not `npm`.
- [ ] Manual verification (npm tightening, no pyproject present): in a scratch dir with only a version-less `package.json` (`{"private": true}`) and no `pyproject.toml`/`VERSION`, confirm detection now reports `<none>` (falls through cleanly) instead of `npm`.
- [ ] Manual verification (npm still wins when it should): in a scratch dir with a `package.json` that *does* have a `"version"` field and no other tool markers, confirm detection still reports `npm` — the tightened condition must not regress the normal case.
- [ ] Manual verification (poetry still wins over PEP 621 branch): in a scratch dir with `pyproject.toml` containing both `[tool.poetry]` and a `[project]` table with `version = "..."`, confirm detection reports `poetry`, not `pyproject` — branch ordering must not regress existing poetry repos.
- [ ] Manual verification (apply path): with the PEP 621 scratch repo from test 1, dry-run the `pyproject)` apply snippet with a `NEW` version and confirm `pyproject.toml`'s `[project].version` is updated in place and no other lines are touched; repeat with a `uv.lock` present (touch an empty one and stub `uv` as a no-op shell function) to confirm the lock-regeneration branch is invoked only when `uv.lock` exists.
- [ ] Automated tests: N/A — `commands/repo/release.md` is a prose/bash-snippet command spec with no existing test harness; verification is manual per above, consistent with how the sibling VERSION-vs-package.json reconciliation (commit `1bb787f`) was verified.
- [ ] Edge cases: `pyproject.toml` present but malformed/no `[project]` table at all (must fall through, not error); `[project]` table present but no `version` key inside it (must fall through to `npm`/`version-file`, not false-match); `version` key present elsewhere in the file inside an unrelated table like `[tool.poetry.dependencies]` (must not false-match the PEP 621 branch when `[project]` itself has no version key).

rjwalters:
## Summary

`/repo:release` Phase 2 had no detection branch for PEP 621 — the modern
uv/hatchling/flit/setuptools standard where the version lives in
`pyproject.toml`'s `[project]` table. Only `[tool.poetry]` was recognized, so a
Python repo of that shape fell through to the `npm` branch, which matched on the
mere *presence* of `package.json`. In the reported field case that `package.json`
was a Loom workspace scaffold (`"private": true`, **no** `version` field), so a
blind `npm version` would have invented a `0.0.1` in the wrong file while the
real version went stale.

This adds a `pyproject` detection branch ordered between `poetry` and `npm`,
tightens the `npm` match to require a `version` field, and adds the corresponding
Phase 3 read example and Phase 5 apply branch.

## Changes

- **Phase 2** — new `pyproject` branch between `poetry` and `npm`. `poetry` keeps
  priority because modern poetry files also carry a `[project]` table and
  `poetry version` remains the correct apply path there. The `version =` grep is
  scoped to the lines *inside* `[project]` via an `awk` block-extractor, so a
  `version =` in e.g. `[tool.poetry.dependencies]` cannot false-positive it.
- **Phase 2** — `npm` now additionally requires `grep -q '"version"'
  package.json`; a version-less scaffold falls through the chain instead of
  misdirecting the bump.
- **Phase 3** — `pyproject` "read the current version" example (`tomllib`
  one-liner, with the older-Python fallback noted).
- **Phase 5** — `pyproject)` apply branch that rewrites only the `[project]`
  version line and regenerates `uv.lock` when present (gated on `uv` being on
  `PATH`); `pyproject` added to the "tools that don't self-commit" list.
- Prose in the intro and *Cross-source reconciliation* updated to match the
  tightened `npm` condition.
- CHANGELOG `## Unreleased` entry.

## Deviation from the issue's suggested snippet (please review)

The issue's Phase 5 snippet scopes its regex with `(\[project\][^\[]*?\nversion
\s*=\s*)`. **That form silently substitutes nothing** whenever a list-valued key
(`classifiers`, `dependencies`, `keywords`) appears before `version` in
`[project]` — the `[` in `["..."]` terminates the `[^\[]` run. Verified against a
fixture:

```
[project]
name = "x"
classifiers = ["Programming Language :: Python"]
dependencies = [
  "requests>=2",
]
version = "1.2.3"
```
→ `subs: 0`, file unchanged. Phase 5 would then commit and tag a release whose
version never moved.

The landed branch instead anchors the scope to lines up to the next table
*header* — `(?ms)(^\[project\][^\n]*\n(?:(?!^\[)[^\n]*\n)*?version[ \t]*=[ \t]*)"[^"]*"`
— which mirrors the `awk` scoping used in the Phase 2 detection branch, and
**exits non-zero** (`ERROR: no [project].version found`) on zero substitutions
rather than no-op-bumping. The `uv lock` gate also gained a trailing `|| :` so
the branch's exit status isn't the gate's false result on a repo with no lock
file.

| Detection recognizes `pyproject.toml` `[project].version` (non-poetry) before the npm fallback | Yes | Fixture with PEP 621 `pyproject.toml` + scaffold `package.json` → `pyproject — pyproject.toml [project].version (PEP 621)` |
| `package.json` lacking a `version` field no longer matches as `npm` | Yes | Fixture with only `{"private": true}` → `<none> — no tool detected` |
| npm still wins when it should | Yes | Fixture with `{"name":"a","version":"0.1.0"}` → `npm — npm + package.json` |
| poetry still wins over the PEP 621 branch | Yes | Fixture with both `[tool.poetry]` and `[project] version` → `poetry — poetry + [tool.poetry]` |
| pyproject apply path edits `[project].version` in place, touching nothing else | Yes | `diff` after apply shows exactly one changed line; the decoy `version` in `[tool.poetry.dependencies]` is untouched |
| Apply path regenerates `uv.lock` when present | Yes | Stub `uv` on `PATH`: invoked with `uv.lock` present, not invoked without it, not invoked when `uv` is off `PATH`; exit 0 in all three |
| Edge: no `[project]` table / no `version` key inside it / `version` only in another table | Yes | All three fall through detection cleanly (`<none>` or `version-file`); the apply branch fails loudly rather than silently |

## Test Plan

Automated tests are N/A — `commands/repo/release.md` is a prose/bash-snippet
command spec with no test harness (consistent with how the sibling
VERSION-vs-`package.json` reconciliation was verified). Verification was manual
but mechanical: the detection chain and the `pyproject)` case branch were
**extracted from the edited file** (`awk`/`sed`), syntax-checked with `bash -n`,
and run against seven scratch fixtures covering every row in the table above.

`pnpm test` → **954 pass / 0 fail** (matches the pre-change baseline; no test in
this repo covers `release.md`).

Closes #83


rjwalters:
## Approved

Merge state `CLEAN`, CI green (`pnpm test (guard/handoff/loss-check suite)` → pass, 1m4s), `Closes #83` present. Local `pnpm check:ci` → **954 pass / 0 fail**, matching the PR body.

I did not take the deviation on faith — I extracted the Phase 2 chain and the Phase 5 `pyproject)` branch **verbatim from the edited `release.md`** (`awk`/`grep` + `sed` to strip the `case` label), `bash -n`'d both, and ran them against scratch fixtures.

### The regex deviation is justified and correct

**The builder's claim reproduces exactly.** Curator's `(\[project\][^\[]*?\nversion\s*=\s*)"[^"]+"` against a `[project]` carrying `classifiers = [...]` and a multi-line `dependencies = [...]` ahead of `version`:

```
[curator] subs=0
--- diff vs original --- (NO CHANGE - file untouched)
```

The `[` opening the list value terminates the `[^\[]` run, and since the curator's snippet calls `re.sub` (not `subn`) and writes the result back unconditionally, this is a **silent** no-op — Phase 5 would then commit and tag a release whose version never moved. Given `classifiers`/`dependencies`/`keywords` are alphabetically-and-conventionally *before* `version` in a hand-written `[project]` table, this is the common shape, not an exotic one.

The shipped `(?ms)(^\[project\][^\n]*\n(?:(?!^\[)[^\n]*\n)*?version[ \t]*=[ \t]*)"[^"]*"` fixes it, and the anchoring is sound: each `(?:...)` repetition consumes through a `\n`, so `version` can only match at a line start, and `(?!^\[)` stops the scope at the next table *header* — mirroring the `awk '/^\[project\]/{f=1;next} /^\[/{f=0} f'` used in detection.

### What I actually ran

Detection chain, 6 fixtures (with a stubbed `poetry` on `PATH` so the poetry branch was genuinely reachable):

| Fixture | Result |
|---|---|
| PEP 621, `classifiers`+`dependencies` before `version`, decoy `version` in `[tool.poetry.dependencies]` | `pyproject` |
| PEP 621 + scaffold `{"private": true}` `package.json` | `pyproject` (not `npm`) |
| both `[tool.poetry]` and `[project] version` | `poetry` |
| version-less `package.json` only | `<none>` |
| `{"name":"a","version":"0.1.0"}` | `npm` |
| `[project]` with `dynamic = ["version"]`, decoy `version` in another table | `<none>` |

Note the poetry grep `'\[tool.poetry\]'` correctly does *not* fire on `[tool.poetry.dependencies]` (the unescaped `.` is a wildcard but `]` must follow `poetry`), so fixture 1 reaches the new branch as intended.

Phase 5 apply, verbatim snippet, `NEW=9.9.9`:

- Hard case (`classifiers`/`dependencies` first) → exit 0, `diff` shows **exactly one** changed line; the decoy `version = "decoy"` in `[tool.poetry.dependencies]` untouched.
- Decoy in an *earlier* table (`[build-system]`) with `[project]` mid-file → earlier `version = "SHOULD-NOT-MOVE"` untouched, `[project].version` moved. Multiline `^` anchoring holds.
- `[project]` with no `version` key → `ERROR: no [project].version found in pyproject.toml`, **exit 1**, file unchanged. Fails loudly, as claimed.
- Padded `version    =    "1.0.0"` and a final version line with **no trailing newline** → both handled, exit 0.

`uv.lock` gate, four ways: lock+`uv` present → `uv lock` invoked, exit 0; no lock → not invoked, exit 0; lock present but `uv` genuinely off `PATH` → not invoked, exit 0 and version still moved. The `|| :` justification checks out too — I ran the *curator's* form on a no-lock fixture and the branch exited **1** purely on the gate's false status.

Docs are right where they need to be: the deviation rationale lives in the shipped `release.md` prose (the three "load-bearing" bullets), not only in the PR body, so the next editor can't "simplify" the regex back into the silent-no-op form.

### Non-blocking follow-ups

1. **`|| :` also swallows a genuine `uv lock` failure.** With a stubbed `uv` exiting 7, the branch still exits 0 (uv's stderr is visible, so it's not invisible — just not fatal). The gate-vs-command distinction would be cleaner as `if [ -f uv.lock ] && command -v uv >/dev/null 2>&1; then uv lock; fi`, which returns the gate's true-ish status when skipped *and* propagates a real failure. Since Phase 5 says to stage the regenerated `uv.lock` in the version commit, a swallowed failure means committing a stale lock.
2. **Detection accepts a quote style the apply path rejects.** `version = '1.0.0'` is legal TOML; detection's `^version[[:space:]]*=` matches it and reports `pyproject`, but the apply's `"[^"]*"` doesn't, so Phase 5 exits 1. That's the *safe* failure direction (loud, not a mis-bump), so not blocking — but widening to `["']` (or narrowing detection) would close the gap.
3. **Drift gate** (`Within a single tool`) still names only `cargo` and `poetry` as single-version-location tools; `pyproject` belongs there too, with the nuance that `uv.lock` mirrors the version.
4. **Reconciliation guard is now tautological.** Line 199 keeps `&& grep -q '"version"' package.json`, which the tightened `npm` branch already guarantees — the prose above it was correctly updated to two conditions while the code kept three. Harmless defense-in-depth, just a prose/code mismatch.
5. The `pyproject)` one-liner is ~450 chars. Consistent with the sibling `case` branches' style and fully explained by the bullets underneath, so I'd leave it — noting only that a heredoc form would read better if this branch ever grows.

Scope boundary with #81 verified: no open PR touches the `npm` match, so the tightening lands exactly once here, and #81's AC item 1 becomes a no-op verification as the curator planned.

PR #106, opened 2026-08-04
feat(all): sync to the default branch before Docs when the branch is fully pushed

### LINKED ISSUE #82: /repo:all: run reset's sync-and-switch before Docs when the working branch is fully pushed
## Context

Surfaced during a `/repo:all` run in a consumer repo (klayout-tools). The session started on a PR branch that was fully pushed and 6 commits behind the default branch. Because Reset is ordered last, the Docs stage compared prose against the stale checkout: the real content drift (a stale "aspirational" label on a command table) lived in the default branch's newer copy of README.md, and editing the branch's copy would both pollute the open PR's diff and block the later branch switch (the file differed between branches). Every content fix had to be deferred, then re-done after Reset ran.

Reset-last is documented as "so the earlier scans and cleanup happen while you're still on the working branch" — which only pays off when the working branch holds unpushed WIP the scans should see. When the branch is fully pushed and behind the default branch, it's strictly worse.

- [ ] After the Audit stage, `/repo:all` checks branch state: current branch fully pushed (no unpushed commits, clean or safely-carryable tree) and behind the default branch.
- [ ] In that case, run the reversible sync-and-switch half of Reset (fetch, checkout default, ff-only pull) before the Docs stage, so Docs/Tidy/Update-tools operate on the fresh default branch.
- [ ] The pruning half of Reset (branch/worktree/stash review) still runs last with its existing gates.
- [ ] With unpushed WIP on the working branch, behavior is unchanged (reset stays last).

---

### Root cause / technical context

`/repo:all`'s stage order (`commands/repo/all.md`, "Stages") is fixed: Audit →
Docs → Tidy → Update tools → Reset. The rationale given for putting Reset last
("so the earlier scans and cleanup happen while you're still on the working
branch") assumes the working branch is the *right* place to be looking —
i.e., it holds content that differs from (and should take precedence over)
the default branch. That assumption fails whenever the working branch is
fully pushed and strictly behind the default branch: there is no local-only
content to protect, and the stale checkout actively misleads Docs/Tidy into
reporting drift that has already been fixed upstream, or into editing a copy
that will immediately conflict with the branch switch `/repo:reset` performs
at the end.

`/repo:reset` (`commands/repo/reset.md`) is itself two logically separable
halves:

- **Sync-and-switch** (reversible): step 1 (working-tree safety check), and
  step 4 (`git fetch --all --prune`, then `git checkout "$default" && git pull
  --ff-only`).
- **Pruning** (needs explicit opt-in / gated): step 2 (stash review — never
  auto-drops) and step 3 (branch & worktree review — the
  [[branches]] permanent-loss check, `--prune` gate).

This issue asks `/repo:all` to run the sync-and-switch half early — right
after Audit — when it can prove nothing is lost by doing so, and leave the
pruning half exactly where it is (last), since pruning decisions (which
branches/worktrees/stashes are safe to remove) are independent of which
branch Docs/Tidy/Update-tools ran against.

### Interaction with the PR #93 verify-after-write / re-verify-before-print machinery

Issue #89 (fixed by PR #93, merged) added a **different** safety mechanism
that also touches `all.md`'s Final Summary: each doc-editing command
(`docs.md`, `readme.md`, `gitignore.md`, `links.md`) now re-reads its own
edit immediately after applying it ("Verify after write"), and `all.md` gained
a "Re-verify before printing" step (`commands/repo/all.md` lines ~104-128)
where every stage re-confirms its edits are still on disk immediately before
the consolidated summary prints — because a concurrent writer (another agent,
a background `git stash`/`git checkout --`, a Loom sweep quarantining the
primary clone's working tree) can revert a fix between the stage reporting it
and the final summary printing it.

The #89 curator's comment explicitly distinguishes the two:

> #82 is about stage *ordering* — running Reset's sync-and-switch before Docs
> when the working branch is fully pushed and behind the default branch. This
> is a different failure: the checkout was already fresh and correct, and the
> edits were lost to a *concurrent writer*, not to stale content or a blocked
> branch switch. Related, worth cross-referencing, not a duplicate.

**They compose cleanly and orthogonally — stage reordering does not make
verify-after-write redundant, and vice versa:**

- Verify-after-write protects against *loss of an already-correct edit*
  between apply-time and print-time, on **whatever branch is checked out at
  the time**. It doesn't care whether that branch is the original working
  branch or the default branch reached via an early sync-and-switch.
- Stage reordering (this issue) protects against *editing the wrong
  checkout in the first place* — content drift that's already fixed on the
  branch you should be looking at, or an edit that will collide with the
  later branch switch. It doesn't care whether a concurrent writer might
  later revert the edit.
- The total exposure window between a Docs-stage edit and the final summary
  print is **unchanged** by this reordering — Tidy, Update-tools, and Reset's
  pruning half still run in between either way — so moving the switch earlier
  neither shortens nor lengthens the re-verify machinery's job. Both
  mechanisms should be implemented as documented, independently; the Builder
  does not need to special-case their interaction.
- One second-order effect worth calling out in the implementation: after an
  early sync-and-switch, Docs/Tidy/Update-tools edits land directly on the
  checked-out default branch (e.g. `main`) instead of a feature branch. That
  is already possible today whenever a user runs `/repo:all` while already on
  the default branch — this issue doesn't introduce a new class of risk, it
  just makes it reachable from more starting states. The existing
  verify-after-write contract in `docs.md`/`readme.md`/`gitignore.md`/`links.md`
  already covers this case unconditionally regardless of which branch is
  checked out, so no changes are needed there.

### Suggested detection logic (Builder guidance, not prescriptive)

Insert a check between the Audit and Docs stages in `commands/repo/all.md`:

```bash
current=$(git symbolic-ref --short HEAD 2>/dev/null) || current=""
default=$(git symbolic-ref --short refs/remotes/origin/HEAD 2>/dev/null | sed 's|^origin/||')
git fetch origin --quiet

if [ -n "$current" ] && [ "$current" != "$default" ]; then
  # "Fully pushed": HEAD has an upstream and is not ahead of it.
  if git rev-parse --abbrev-ref --symbolic-full-name '@{u}' >/dev/null 2>&1; then
    ahead_of_upstream=$(git rev-list --count '@{u}..HEAD')
  else
    ahead_of_upstream=""   # no upstream at all => never pushed, not eligible
  fi
  behind_default=$(git rev-list --count "HEAD..origin/$default" 2>/dev/null || echo 0)

  if [ "$ahead_of_upstream" = "0" ] && [ "$behind_default" -gt 0 ]; then
    # Eligible for early sync-and-switch — subject to the working-tree check below.
    :
  fi
fi
```

Notes for the Builder:

- **Recommend requiring a clean working tree** (`git status --porcelain`
  empty) rather than implementing the "safely-carryable" (stash-and-carry)
  branch of the acceptance criteria in the first pass. A dirty-but-carryable
  tree adds meaningful complexity (the stash created to enable the switch has
  no natural point in this stage sequence to be popped back, since the run
  ends on the default branch) for a narrow benefit; scope it as a follow-up
  if there's demand. If the tree is dirty, fall back to the current
  behavior (Reset stays last) rather than attempting the early switch.
  Note in the eventual PR/summary if this narrowing is applied, since it's a
  deliberate scope reduction from the literal acceptance-criteria wording
  ("clean or safely-carryable tree").
- "No upstream" (branch never pushed) must **not** be treated as "fully
  pushed" — treat it as ineligible, same as `ahead_of_upstream > 0`.
  Only the two-checkbox literal case (has upstream AND zero commits ahead of
  it) counts as "no unpushed commits."
- When the switch happens, print a line noting it (e.g. `Reset: synced early
  — <branch> was fully pushed and N commits behind <default>; switched
  before Docs`) so the final `REPO:ALL COMPLETE` summary's `Reset:` line
  doesn't appear to describe a switch that already happened earlier in the
  transcript. Avoid reporting the branch switch twice.
- The final summary's `Reset:` line (see `all.md`'s "Final Summary" example)
  should still cover the pruning-half output (branches/worktrees/stashes
  reviewed) even when the switch itself happened earlier — don't drop that
  reporting just because sync-and-switch moved up.
- `/repo:reset`'s own steps 1–4 are unaffected when `/repo:reset` is run
  standalone (not via `/repo:all`) — this issue only changes `/repo:all`'s
  orchestration, not `/repo:reset` itself.

### Duplicate check

Ran `.loom/scripts/check-duplicate.sh --include-merged-prs` against this
issue's title/body; it returned three "100% similarity" matches (#102, #82
itself, #81) that are unrelated on inspection — the heuristic appears to be
producing false positives here (title/body length or generic-word overlap),
not a real signal. Manually cross-checked against #89 (see above): related,
distinct scope, already documented by the #89 curator and confirmed here.
No true duplicate found.

## Affected Files

- `commands/repo/all.md` — insert the branch-state check and conditional
  early sync-and-switch between the "1. Audit" and "2. Docs" sections;
  adjust the "5. Reset" section text to describe the now-conditional
  sync-and-switch (it may have already run); update the "Final Summary"
  example/guidance so the `Reset:` line composes correctly whether or not
  the switch happened early.
- `commands/repo/reset.md` — no functional change expected when run
  standalone; note only if the Builder finds it useful to name the
  sync-and-switch / pruning split explicitly here too (optional, not
  required by the acceptance criteria).

## Test Plan

- [ ] Manual verification (fully-pushed-and-behind case): create a local
  branch with no unpushed commits (push it, then have the remote default
  branch gain new commits — e.g. merge something else into `main`). Run
  `/repo:all` from that branch and confirm: Audit runs first, then the repo
  switches to the default branch **before** Docs runs, Docs/Tidy/Update-tools
  operate on the fresh default-branch checkout, and the branch/worktree/stash
  pruning still happens last.
- [ ] Manual verification (unpushed-WIP case, regression guard): create a
  branch with an unpushed commit or dirty working tree, run `/repo:all`, and
  confirm the order is unchanged — Reset (in full) still runs last, exactly
  as before this change.
- [ ] Manual verification (already-on-default case, regression guard): run
  `/repo:all` while already on the default branch; confirm no spurious
  "switch" is attempted/reported (current == default is not eligible) and
  behavior matches current `main`.
- [ ] Edge case: branch has diverged from default (both ahead and behind) —
  confirm this is correctly excluded (`ahead_of_upstream` check is about the
  branch's own upstream, but if the branch is ahead of its own upstream at
  all it's ineligible; also confirm a diverged-from-default case, which is a
  different axis, doesn't crash the `behind_default` computation).
- [ ] Automated tests: if `commands/repo/tests/` has coverage patterns for
  `all.md`/`reset.md` doc-drift assertions (see
  `commands/repo/tests/test-verify-fix-persistence.sh` added by PR #93 for a
  similar precedent), add a corresponding fixture/doc-drift test asserting
  `all.md` documents the conditional early sync-and-switch and its ordering
  guarantees.

## Related

- #89 / PR #93 (merged) — concurrent-writer edit loss (verify-after-write).
  Related but distinct: #89 fixes edits being silently reverted after a
  correct apply; this issue fixes editing the wrong (stale) checkout in the
  first place. See "Interaction with the PR #93 verify-after-write /
  re-verify-before-print machinery" above for how the two compose.

rjwalters:
## Summary

`/repo:all` ran the whole of `/repo:reset` last, on the rationale that "the earlier scans and cleanup happen while you're still on the working branch." That pays off only when the working branch holds unpushed work those scans should see. On a **fully pushed** branch that is **behind the default branch** it inverts: the Docs and Tidy stages compare prose against a stale checkout, re-report drift already fixed upstream, and any edit they do apply both pollutes the open PR's diff and blocks the branch switch Reset performs at the end (the motivating consumer-repo run had to defer every content fix and re-do it after Reset).

This splits `/repo:all`'s orchestration along the seam `/repo:reset` already has — a reversible **sync-and-switch** half and a gated **pruning** half — and runs only the first half early, gated on a proof that nothing can be lost.

## Changes

- **`commands/repo/all.md`** — new **stage 2, "Sync early, if and only if nothing can be lost"**, between Audit and Docs, with the eligibility conjunction (bash block + a per-condition rationale table): on a branch that isn't the default, `origin/HEAD` resolves, working tree clean, branch **has an upstream** and is **0 commits ahead of it**, and behind `origin/<default>` by ≥1. When eligible it runs `/repo:reset` step 4 only (`git fetch --all --prune`, `checkout <default>`, `pull --ff-only`) and prints one line noting the switch; under `--ask` it asks first; a failed checkout/ff-only pull changes nothing and defers to stage 6. When ineligible it is a silent no-op. Stages renumbered 3–6; **stage 6 (Reset) keeps the pruning half last** with its gates, and explicitly must not report the switch twice. Final Summary gains guidance + an example for a composed `Reset:` line.
- **`commands/repo/reset.md`** — new "Two halves" section naming the sync-and-switch (steps 1, 4) / pruning (steps 2, 3) split and stating that standalone `/repo:reset` still runs all four steps in order, unchanged.
- **`commands/repo/tests/test-early-sync-switch.sh`** (new, 68 cases) — fixture section exercising a faithful transcription of the documented check against real git remotes, plus doc-drift assertions.
- **`hooks/repo/tests/run.sh`** — delegation block for the new suite, folding its real PASS/FAIL counts into the totals (breakdown-sums-to-headline self-check still holds).
- **`CHANGELOG.md`** — Unreleased entry.

## Deliberate scope narrowing

The acceptance criteria say "clean **or safely-carryable** tree". This PR implements the **clean-tree-only** first pass, per the Curator's explicit recommendation: a stash created to enable the switch has no natural point to be popped back, since the run ends on the default branch. A dirty tree (tracked modification *or* untracked file) falls back to the current behaviour — Reset stays last. Both are covered by tests.

## Composition with PR #93 (issue #89)

Orthogonal mechanisms, both preserved. Verify-after-write / re-verify-before-print protects an *already-correct edit* from a concurrent writer, on whatever branch is checked out; this reordering prevents *editing the wrong checkout in the first place*. The exposure window between a Docs edit and the summary print is unchanged (Tidy, Update tools, and Reset's pruning half still run in between either way). `test-verify-fix-persistence.sh` passes unmodified (61/61), and the new suite carries a cross-check that `### Re-verify before printing` survived the reorder.

| After Audit, `/repo:all` checks branch state: fully pushed + behind default | ✅ | New stage 2 in `all.md` with the documented bash block; `early_sync_eligible()` transcription tested across 9 repo shapes |
| Run the reversible sync-and-switch half before Docs in that case | ✅ | Stage 2 runs `/repo:reset` step 4 only; test asserts the switch lands on default, up to date, tree clean, and that the checkout then holds the default branch's newer content |
| Pruning half still runs last with existing gates | ✅ | Stage 6 states the pruning half always runs there; doc-drift test asserts heading order 2 → 3 → 4 → 5 → 6 by line number; fixture asserts the sync half deletes no branch |
| Unpushed WIP: behavior unchanged (reset stays last) | ✅ | `1 commit ahead of upstream -> INELIGIBLE`, unpushed commit untouched, still on the working branch; `all.md` states the ineligible path is a no-op |
| "No upstream" is ineligible, not "fully pushed" | ✅ | Dedicated fixture case (branch never pushed, default advanced) → INELIGIBLE; prose + code-comment assertions in the drift section |
| Already-on-default: no spurious switch | ✅ | Fixture case on `main` behind `origin/main` → INELIGIBLE |
| Diverged branch doesn't crash / isn't wrongly excluded | ✅ | Motivating case is diverged (1 ahead of default, 6 behind) and is ELIGIBLE; `behind_default` computes 6 |
| Report the early switch once; summary's `Reset:` line still covers pruning | ✅ | Stage 2 report line + stage 6 "don't report the switch twice" + Final Summary composed-line example, all asserted |
| Automated doc-drift coverage following the PR #93 precedent | ✅ | `test-early-sync-switch.sh` doc-drift section over `all.md` and `reset.md`, wired into `pnpm test` |

## Test Plan

`pnpm test`: **1022 pass / 0 fail** (baseline on `main` was 954/0; +68 from the new suite). No existing assertion was modified — `test-verify-fix-persistence.sh` still reports 61/61 against the edited `all.md`.

The new suite builds a real bare origin plus per-case clones and covers: the motivating pushed-and-6-behind PR branch (eligible, diverged axis included), post-switch loss check (branch ref, upstream, and commits survive; checkout now holds fresh content), unpushed commit, never-pushed/no-upstream, dirty tracked file, untracked file, already-on-default, pushed-but-not-behind, detached HEAD, unreachable remote with no `origin/HEAD`, and a repo with no remote. Detection is also asserted to leave HEAD and the working tree untouched, so it costs an ineligible run nothing.

Closes #82


rjwalters:
## Approved

Reviewed the full diff against issue #82's curator guidance, re-derived the eligibility conjunction against real git fixtures, and verified the PR #93 contract by running the suite rather than trusting the claim.

### Eligibility conjunction — correct, and fail-closed

I built adversarial fixtures rather than relying on the suite's own transcription. Findings:

- **"0 ahead of UPSTREAM" is the right gate, not "0 ahead of default."** This is the one place the check could plausibly have been wrong, and it isn't. A pushed PR branch is normally ahead of `origin/<default>`; gating on that would make the feature never fire on exactly the motivating case. The unpushed-work axis is `@{u}..HEAD`, and the doc says so explicitly (`all.md` lines 102-105) rather than leaving it implicit. The suite's `motivating` case asserts both axes simultaneously (1 ahead of default, 6 behind, still ELIGIBLE), which is the assertion that pins this down.
- **Never-pushed branch is correctly ineligible.** `ahead_of_upstream=""` when `@{u}` doesn't resolve, and the comparison is `[ "$ahead_of_upstream" = "0" ]` — a string equality, so empty never passes. Confirmed against a fixture with a local-only commit: INELIGIBLE, and the run stays on the branch. This is fail-closed by construction, not by luck: every failure mode of the ahead computation (rev-list error, missing upstream) yields a non-`"0"` string.
- **Short-circuit ordering is safe.** `[ "$ahead_of_upstream" = "0" ] && [ "$behind_default" -gt 0 ]` — the string test runs first, so the arithmetic `-gt` never evaluates a possibly-empty `behind_default`. And `behind_default` itself is `... 2>/dev/null || echo 0`, so a missing `origin/$default` degrades to 0 → ineligible.
- **No path switches or pulls with a dirty tree or unpushed commits.** `git status --porcelain` (which includes untracked, verified by the suite's `notes.txt` case) is inside the guard conjunction, ahead of any mutation. The only mutating command in the ineligible path is the detection's own `git fetch origin --quiet`, which cannot move HEAD or touch the tree — the suite asserts exactly that (`detection does not move HEAD` / `does not dirty the tree`).

I also verified the eligible path is genuinely non-destructive: it runs only fetch/checkout/pull --ff-only. The branch ref, its upstream, and its commits all survive; nothing in stage 2 can delete a branch, worktree, or stash. The pruning half staying in stage 6 with the `[[branches]]` permanent-loss check is what keeps that true.

### PR #93 re-verify-before-print contract — genuinely intact

Not taken on faith. I ran `test-verify-fix-persistence.sh` in the worktree against the edited `all.md`: **61/61 pass, 0 fail**, and the file is untouched by this PR (not in the changed-file list). I also read its `all.md` assertions directly to rule out a suite that passes because it was weakened — all six are substantive phrase checks (`### Re-verify before printing`, the summary-print tie-in, `never folded into that stage's fixed count`, `reverted after apply — needs re-run`, `Only the affected stage's line changes`, the quiet-repo clause), none of which this PR touches.

The one line the diff *does* change in that section is prose-only: `A fix applied in stage 2` → `A fix applied in the Docs stage`. That is the correct edit given the renumber — hardcoding "stage 3" would have re-created the same drift — and no assertion keys on the stage number. The new suite's cross-check for `### Re-verify before printing` is a reasonable belt-and-braces addition.

### New test suite — exercises the data-loss-critical cases

Confirmed present and meaningful, not decorative: dirty tracked file, untracked file, unpushed commit, never-pushed/no-upstream, detached HEAD, unresolvable `origin/HEAD`, and a repo with no remote at all. The dirty-tree case is well constructed — it asserts ELIGIBLE *before* dirtying and ELIGIBLE again after cleanup, so it proves the dirty state is what flipped the result rather than some unrelated precondition. The degenerate-HEAD cases assert exit status 0 alongside the verdict, so an ineligible result can't be a crash in disguise. The no-`origin/HEAD` fixture deliberately points the remote at a missing path, with a comment explaining that modern git would otherwise restore the ref on fetch — that's the right level of care.

The suite tests a transcription of the documented block rather than the block itself, which is inherent to a prompt-document repo; the doc-drift section (including the heading-order-by-line-number assertions for 2 → 3 → 4 → 5 → 6) is the correct mitigation and follows the PR #93 precedent the issue asked for.

### Verification

- `pnpm test` in the worktree: **1022 pass / 0 fail**, matching the PR claim. Per-suite breakdown sums to the headline (the harness self-check holds), with the new suite contributing 68.
- CI: `pnpm test (guard/handoff/loss-check suite)` **pass**.
- Merge state: **CLEAN** / mergeable. No rebase needed.
- Scope narrowing to clean-tree-only is flagged in the PR body, matches the Curator's explicit recommendation verbatim, and is recorded in the CHANGELOG — the acceptance-criteria deviation is disclosed rather than quietly taken.

### Non-blocking notes (follow-up material, not changes requested)

1. **"change nothing" is imprecise for the checkout-succeeded / pull-failed split.** `all.md` line 117-119 says if the checkout *or* the `--ff-only` pull fails, "change nothing." For a checkout failure that's accurate. For a pull failure it isn't: the checkout already landed, so the run is now sitting on a diverged local default branch — neither the fresh `origin/<default>` nor the original working branch. No loss (tree was clean, the branch is fully pushed, local default commits are intact), and stage 6 surfaces the divergence with reset's own "ask how to proceed" path, but Docs would then edit a stale default checkout, which is the failure class this stage exists to prevent. Worth splitting the two failure modes and saying explicitly whether to switch back.

2. **The worktree-collision case is the most likely trigger in a Loom-managed repo and isn't named.** Running `/repo:all` inside `.loom/worktrees/issue-N` on a fully-pushed, behind-main branch evaluates ELIGIBLE, then `git checkout main` fails with `fatal: 'main' is already used by worktree at ...` (exit 128, HEAD unchanged — I confirmed this). The documented fallback handles it correctly and safely, so this is not a defect; it just means the feature quietly no-ops in the environment where this repo's own agents run most often. Naming it alongside the diverged-default example would set expectations.

3. **`git fetch origin --quiet` runs after `default` is computed.** If `origin/HEAD` isn't set locally but the remote is reachable, the fetch that could resolve it happens too late, so the run is ineligible. Conservative and safe (falls back to the old order), just leaves a little on the table.

None of these affect correctness of the guard or introduce a loss path. Approving.

Issue #29, opened 2026-07-27
guard-destructive: curl-pipe pattern false-positives on any pipe target whose path contains "sh" (e.g. /usr/share, tee, shasum)


rjwalters:
## Context

Hit three times live on 2026-07-27 during an infra session (safehouse D15 migration). The BLOCK pattern at `hooks/guard-destructive.sh:253`:

```
'curl .* \| .*sh'
```

matches `sh` **anywhere** in the piped command, so it blocks pipelines that never execute a shell:

```bash
# BLOCKED (false positive): piping a keyring to tee — the "sh" is /usr/SHare
curl -fsSL https://pkgs.tailscale.com/stable/ubuntu/noble.noarmor.gpg \
  | sudo tee /usr/share/keyrings/tailscale-archive-keyring.gpg >/dev/null
```

Anything piped to `tee`, `shasum`, `sha256sum`, or into any path containing `sh` (`/usr/share/…`, `~/.ssh/…`) trips it. The workaround (download to a temp file, then `sudo install`) works but turns a one-liner into three and trains operators to rubber-stamp the guard.

The third hit is the best exhibit: `gh issue create` with *this issue's body* was itself blocked, because the body quotes the offending pipeline in a code fence. The guard matches on the raw command string, so any command that merely *mentions* `curl … | … sh` — filing a bug report, grepping a log, echoing documentation — is treated as executing it.

(The same session also hit the pattern on a genuine `curl … install.sh | sudo sh` — that block is correct and should stay.)

## Suggested acceptance criteria

- [ ] The curl-pipe rule only fires when the piped-to command **is** a shell, e.g. anchor on the command position: `curl [^|]*\| *(sudo +)?([^ ]*/)?(ba|da|z|k)?sh\b` — matching `sh`, `bash`, `zsh`, `sudo sh`, `/bin/sh`, but not `tee /usr/share/…` or `shasum`
- [ ] `wget -O- … | sh` gets the same treatment if it's covered (and the redundant `curl .* \| bash` line at :254 folds into the fixed pattern)
- [ ] A regression list in the hook's tests (if any) covering: `| sudo tee /usr/share/x` (allow), `| shasum -c` (allow), `| sudo sh` (block), `| bash -s -- args` (block)

## Broader note (optional)

[...]

PR #107, opened 2026-08-04
fix(guard): make ml_segment() heredoc-aware so body lines are not command words

### LINKED ISSUE #84: guard-destructive: lifecycle matcher false-positives on heredoc body lines inside composite commands
## Bug

The system-lifecycle matcher (halt/reboot/poweroff/shutdown/init 0/6) is
command-word-aware — it checks `toks[1]` per segment — but the segment
splitter feeds it **heredoc body lines as if they were command segments**. A
composite like:

```
gh issue create --body "$(cat <<'EOF'
...
shutdown Iq is specified at ...
EOF
)"
```

is blocked as "system lifecycle command: shutdown" because a heredoc body
line starts with the word.

Observed twice on 2026-07-31 (guard version: post-repo#29, dispatched via
Loom's guard-destructive.sh dispatcher):
1. An agent working in 2AMLogic/gf180-ldo had to reword the engineering term
   "shutdown current" to file a spec issue.
2. **Filing this very bug report was blocked** — the first attempt passed the
   issue body inline, and the example heredoc above (inside the body string)
   triggered the same match. This report went in via `--body-file` as the
   workaround.

## Suggested fix

Track `<<WORD` / `<<'WORD'` / `<<-WORD` openers during segmentation and skip
lines until the terminator before matching — same spirit as the repo#29
curl-pipe fix: parse shell structure, don't pattern-match content. Quoted
string arguments (`--body "..."`) are a second, smaller instance of the same
class if quoting is parsed at all.

## Impact

Low severity, real friction: any agent filing text where a line begins with
halt/reboot/poweroff or the s-word (power-management specs, ops runbooks,
guard bug reports) hits it. Fail-open behavior was otherwise correct — only
the specific tool call was denied, nothing broke.

---

## Curator Enhancement

Root-cause analysis, severity reassessment (given #99/#87 removed the
first-party `/repo:followups` trigger), a duplicate check, implementation
guidance, Affected Files, and a Test Plan were added as a [curator
comment](https://github.com/rjwalters/repo/issues/84#issuecomment-5174833993)
rather than rewritten into this body — the original problem statement and
suggested fix above were already clear and correct. See that comment for the
confirmed repro against `origin/main`, the recommended `ml_segment()`
heredoc-opener-tracking approach (shared by all three segment-parsers), and
the scope note on what's already fixed (#71/#76) vs. still open (heredocs
specifically).

rjwalters:
## Summary

`guard-destructive.sh`'s shared multi-line lexer `ml_segment()` (`_ML_QSPLIT_AWK`, ~L969 — used by `parse_force_ops()`, `lifecycle_or_cloud_reason()` and `extract_rm_targets()` since #71/#76) tracked single-/double-quote state but had **no concept of heredoc syntax**. It therefore fell through to its "a raw newline splits" rule and turned every heredoc **body** line into a phantom top-level segment whose first word was read as a command word. A composite like `gh issue create --body "$(cat <<'EOF' … EOF)"` with a body line beginning `shutdown` was hard-denied as a system-lifecycle command — the false positive that blocked filing the bug report that became this issue, and that forced an agent to reword the engineering term "shutdown current" in a spec issue.

Fixed once in the shared lexer, per the curator's guidance, so all three segment parsers get it and cannot drift.

## Changes

- **`hooks/repo/guard-destructive.sh`**
  - New `hd_opener()` AWK helper probes a `<<` for a genuine heredoc operator and reports the terminator word (quotes/backslash stripped), the `<<-` leading-TAB-strip flag, and whether the delimiter is quoted (i.e. the body is not expanded).
  - `ml_segment()` records pending heredoc openers, keeps segmenting the **rest** of the opener line normally, closes that segment at the terminating newline, and then **skips** the body through its terminator so it contributes no segment and no tokens. Multiple openers on one line consume their bodies in opener order; an unterminated heredoc skips the remainder of the buffer (exactly what a real shell does with it).
  - Five deliberate non-heredoc carve-outs so the change can't swallow a later real command: `<<<` is consumed whole as a here-**string**; a `<<` inside an unclosed arithmetic expansion (`$((` / `((`) is a left shift; an all-digit bare delimiter or a delimiter not followed by a separator/redirect/whitespace boundary is rejected; a `<<` inside a shell **comment** is not probed at all; and a newline preceded by an **odd** number of backslashes is a line **continuation**, so the logical opener line has not ended and the bodies do not start there.
  - Safety carve-out: a body attached to a **bare** (expansion-capable) delimiter that carries `$(` or a backtick reverts the whole pending-heredoc region to the legacy separator-active treatment — the same #3679 floor the quoted-span branch applies.
  - Header comment block documents all of the above under a `HEREDOC AWARENESS (#84)` heading.
- **`hooks/repo/tests/test-guard-destructive.sh`** — new `#84` section, 35 cases (16 allow, 19 safety-floor deny/ask), following the existing per-issue convention and the runtime-assembled danger-phrase pattern from #60/#71.
- **`CHANGELOG.md`** — `## Unreleased` entry.

## Review round 2 — the two reported bypasses are closed (commit `5efc13c`)

The review found two shapes where `origin/main` denied, the first revision allowed, and real `bash` genuinely executes the destructive command. Both came from the same assumption — *"the next raw newline after an opener starts the body"* — which holds only when (a) the `<<` really is an operator and (b) the newline really ends the logical line.

**1. `<<WORD` inside a shell COMMENT was probed as a real opener** (`echo hi # <<EOF` + newline + an outside-repo `rm`). Its terminator never appears, so the unterminated-heredoc rule skipped the rest of the buffer, hiding the real second line from `extract_rm_targets()` (the only one of the three parsers that reads raw `$COMMAND` rather than `$COMMAND_NO_COMMENT`).

Fix, at the lexer level per this PR's own "fix it once in the shared lexer" philosophy: `ml_segment()` now tracks a word-initial unquoted `#` (buffer start, or right after whitespace / `;` / `&` / `|` / `(`) and suppresses **only the `hd_opener()` probe** through the end of that physical line. As the review asked, it deliberately does **not** skip to the newline — the characters still flow through normal separator handling, so a trailing `; <cmd>` in a shape like `echo "$(id) # x" ; <cmd>` (where the `#` sits inside a command-substitution-bearing quoted span that this loop walks) is still segmented. The flag resets at every physical newline, including the one that ends a skipped body and one inside an inert multi-line quoted span.

**2. A line continuation after a REAL opener hid the rest of the logical line** (`cat <<EOF \` + newline + `&& <cmd>`), defeating all three parsers. New `cont_nl()` helper: a newline preceded by an **odd** number of backslashes is a continuation (`\\` + newline is a literal backslash plus a real newline, and correctly still opens the body), so the pending bodies are not started there; that newline falls through to the legacy separator handling and the bodies start at the first **non-continued** newline — matching the shell exactly, so the continued command is segmented *and* the body stays inert.

Both fixes are strictly **narrowing**: each can only restore `origin/main`'s decision, never widen one.

Real-`bash` confirmation of the two repros (payload executes on both, so denying is correct):

```
$ bash t2.sh    # echo hi # <<EOF / echo PAYLOAD_RAN
hi
PAYLOAD_RAN
$ bash t3.sh    # cat <<EOF \ / && echo PAYLOAD_RAN / bodytext / EOF
bodytext
PAYLOAD_RAN
```

**Correction to the round-1 differential claim.** The original PR body asserted the differential showed *"the only decision changes are the intended false-positive fixes"*; that was wrong — the corpus simply did not contain a comment shape or a continuation shape. The corrected 34-shape corpus is re-run below and now includes both, plus the shapes that must keep the *new* allow (a comment after a real opener, a comment-only line before one, a markdown `#` heading inside a body, a benign continued opener line, and an even backslash count).

| Criterion (from the issue + curator comment + round-1 review) | Status | Verification |
|---|---|---|
| Exact repro from the issue body now ALLOWS | ✅ | Curator's `printf … \| hooks/repo/guard-destructive.sh` snippet: silence, exit 0 (was `deny: system lifecycle command: shutdown`). Asserted as `#84: issue repro …`. |
| Bare heredoc repro (no `$(…)` wrapper) allows | ✅ | `gh issue create --body-file - <<'EOF' … shutdown … EOF` → allow. |
| `<<-WORD` with tab-indented terminator | ✅ | `#84: <<-EOF with a tab-indented terminator …` |
| Quoted delimiters `<<'EOF'` / `<<"EOF"` / `<<\EOF` | ✅ | Three separate allow assertions. |
| Multiple heredocs on one line | ✅ | `cmd <<A <<B` with lifecycle words in both bodies → allow. |
| Unterminated heredoc | ✅ | Remainder skipped as inert; allow. Commands **before** the opener still deny (asserted). |
| Fix lands in the shared lexer; all three parsers covered | ✅ | One `ml_segment()` change; per-parser allow assertions for `lifecycle_or_cloud_reason`, `parse_force_ops` (`git reset --hard`, was `ask`) and `extract_rm_targets` (outside-repo deep `rm`, was `deny`). |
| Raw `ALWAYS_BLOCK_PATTERNS` scan unaffected | ✅ | `$( )`- and backtick-smuggled catastrophic `rm` inside a heredoc body still deny (reason is the raw pattern match, not a segment match). |
| Genuine multi-line danger still denies | ✅ | `#71`-style genuine multi-line lifecycle case, plus danger after the terminator, before the opener, and piped to on the opener line — all still deny. |
| **Review #1 — comment fake-opener no longer hides the next line** | ✅ | 4 new `assert_deny`s (mid-line `#`, line-start `#`, no space after `#`, and with a later `EOF` line); differential rows B1/B1b/B1c/B1d/B1e all `main=deny → branch=deny`. |
| **Review #2 — continued opener line no longer hides the rest** | ✅ | 2 new `assert_deny`s (lifecycle, outside-repo `rm`) + 1 `assert_ask` (force op); differential rows B2/B2b/B2c restored to `main`'s decision. |
| **Review #3 — regression tests for both shapes** | ✅ | 12 new cases in the `#84` block (7 safety-floor + 5 allows proving the intended fix survives). |
| **Review #4 — differential re-run + PR-body correction** | ✅ | Section above; corpus re-run below. |
| Existing `#29/#53/#60/#69/#71/#76` sections pass unmodified | ✅ | No existing test was touched; full suite green. |
| `pnpm check:ci` passes with 0 regressions | ✅ | **1057 pass / 0 fail** (guard suite 536). Perf check still green (29 ms avg, < 200 ms). |

## Test Plan

- `pnpm check:ci` in the worktree after rebasing onto current `main`: **1057 pass, 0 fail** (guard suite 501 → 536; the count rose from 977 because `main` gained #106's 68-case suite).
- Manual verification of both issue repro shapes against the patched guard (allow) and the pre-patch guard (deny), and of both review repro shapes against real `bash` (payload executes) and the patched guard (deny/ask, matching `main`).
- Differential re-run of the corrected 34-shape corpus (heredoc / here-string / arithmetic / comment / line-continuation / redirect / catastrophic) against `origin/main` vs. this branch. Every row where the two disagree is an intended heredoc-body false-positive fix:

```
   B1 comment fake opener + outside-repo rm     | main=deny  | branch=deny
   B1b comment at line start                    | main=deny  | branch=deny
   B1c comment no space after #                 | main=deny  | branch=deny
   B1d comment + later EOF line                 | main=deny  | branch=deny
   B1e comment fake opener + lifecycle          | main=deny  | branch=deny
   B2 continuation + lifecycle                  | main=deny  | branch=deny
   B2b continuation + force op                  | main=ask   | branch=ask
   B2c continuation + outside rm                | main=deny  | branch=deny
   B2d double backslash (NOT a continuation)    | main=allow | branch=allow
>> issue repro                                  | main=deny  | branch=allow
>> bare heredoc body-file                       | main=deny  | branch=allow
>> plain <<EOF lifecycle body                   | main=deny  | branch=allow
>> <<-EOF tab terminator                        | main=deny  | branch=allow
>> two heredocs one line                        | main=deny  | branch=allow
>> unterminated heredoc                         | main=deny  | branch=allow
>> heredoc body reset --hard                    | main=ask   | branch=allow
>> heredoc body outside rm                      | main=deny  | branch=allow
>> # inside heredoc body markdown                | main=deny  | branch=allow
   danger AFTER terminator                      | main=deny  | branch=deny
   danger BEFORE opener                         | main=deny  | branch=deny
   danger piped to on opener line               | main=deny  | branch=deny
   bare delim body with $( )                    | main=deny  | branch=deny
   bare delim body with backtick                | main=deny  | branch=deny
   smuggled catastrophic in body                | main=deny  | branch=deny
   genuine multiline danger                     | main=deny  | branch=deny
   here-string allow                            | main=allow | branch=allow
   danger after here-string                     | main=deny  | branch=deny
   arith 1 << 3 then danger                     | main=deny  | branch=deny
   (( x << y )) then danger                     | main=deny  | branch=deny
   quoted # with $( ) then danger               | main=allow | branch=allow
   quoted # inert span then danger              | main=deny  | branch=deny
   comment then danger (no <<)                  | main=deny  | branch=deny
   continuation no heredoc + danger             | main=deny  | branch=deny
   opener with # in body then danger after      | main=deny  | branch=deny
```

  (`>>` marks a decision change; all eight are heredoc **bodies** being correctly treated as data.)
- Rebased onto `origin/main` (`aa65cb7`); the only CHANGELOG conflict was two `## Unreleased` bullets landing adjacently — both are preserved, `git diff origin/main -- CHANGELOG.md` removes zero lines.

Closes #84


rjwalters:
## Changes Requested — two genuine deny→allow bypasses in the new heredoc path

The core idea here is right, the shared-lexer placement is right, the header docs are excellent, and the intended false-positive fix works (I reproduced the issue's exact repro: `deny: system lifecycle command: shutdown` on `origin/main` → silent allow on this branch). CI passes (`pnpm test (guard/handoff/loss-check suite)` green, 977/0 locally, guard suite 524) and `mergeStateStatus` is `CLEAN`.

But the differential claim in the PR body — *"the **only** decision changes are the intended false-positive fixes"* — does not hold. I found **two command shapes where `origin/main` denies, this branch allows, and a real `bash` genuinely executes the destructive command.** Both come from the same root assumption in `ml_segment()`: *"the next raw newline after an opener starts the body."* That is true only if (a) the `<<` really is an operator and (b) the newline really ends the logical line. Neither is guaranteed.

Since these are widenings of a deny in the guard that is the last line of defense — and the PR's own header comment states the invariant "never widen a deny into an allow" — these are blocking.

(Side note that makes the case for the fix nicely: posting this very review was itself blocked twice by the pre-fix guard, once on the `halt` line and once on the `/opt/some-vendor/important` path in the repros below. The bug is real; only the fix's edges need tightening.)

---

### Bypass 1 — a `<<WORD` inside a shell COMMENT becomes a fake opener (hits `extract_rm_targets`)

Command string, exactly as the Bash tool would deliver it:

    echo hi # <<EOF
    rm -rf /opt/some-vendor/important

| | decision |
|---|---|
| `origin/main` | `deny: BLOCKED: rm target outside repo scope` |
| this branch | **allow (silent)** |
| real `bash` | runs `echo hi`, then runs the removal — verified |

The `#` comment is not a heredoc opener, but `hd_opener()` is probed on the `<<` inside it, no `EOF` terminator ever appears, and the unterminated-heredoc rule then skips **the entire rest of the buffer** — including the real second line.

Why only `extract_rm_targets`: `lifecycle_or_cloud_reason()` (L1647) and `parse_force_ops()` (L2023) consume `COMMAND_NO_COMMENT`, whose `sed -E 's/(^|[[:space:]])#.*$//'` deletes the comment before segmentation, so the fake opener never reaches them. `extract_rm_targets` at **L1845 consumes raw `$COMMAND`**, so it is exposed. Reproduces with the `#` mid-line or at start-of-line, with or without a space after `#`, and with a later `EOF` line (partial hide) as well.

### Bypass 2 — a line continuation after a REAL opener hides the rest of the logical line (hits all three parsers)

    cat <<EOF \
    && halt
    body
    EOF

| destructive command on the continued line | `origin/main` | this branch |
|---|---|---|
| `halt` | `deny: system lifecycle command` | **allow** |
| `git reset --hard` | `ask` | **allow** |
| outside-repo recursive removal | `deny: rm target outside repo scope` | **allow** |

Real `bash` for the `echo` variant prints `bodytext` then `HIDDEN_AFTER_CONTINUATION` — the command after `&&` **does execute**. Bash removes the `\`+newline, so the logical opener line is `cat <<EOF && <cmd>` and the body starts at the *next* line. The lexer instead treats that first raw newline as the body start and skips the continued line as body text. This one is worse than #1: it defeats all three parsers, including the lifecycle matcher this PR is about.

---

### What I'd like to see

1. **Suppress the opener probe inside a comment.** Track "an unquoted `#` at start-of-buffer or preceded by whitespace has been seen since the last newline" and skip the `hd_opener()` probe while that is set. Please suppress **only the probe** — do *not* skip-to-newline on `#`, because that would hide the trailing `; <cmd>` in a shape like `echo "$(id) # x" ; <cmd>` (where the `#` is inside a quoted span and bash really does run the command) and would trade one widening for another. Probe suppression is strictly narrowing and cannot widen. Routing `extract_rm_targets` through `COMMAND_NO_COMMENT` instead would fix this specific repro, but the lexer-level fix matches this PR's own "fix it once in the shared lexer" philosophy and also covers the other two parsers if their input ever changes.
2. **Do not start heredoc bodies at a continued newline.** When the newline is preceded by an unescaped `\` (mind `\\`), the logical line continues — fall through to the existing separator handling (i.e. keep `main`'s conservative behavior) rather than opening the pending bodies.
3. **Regression tests for both shapes** in the `#84` block, asserting the `origin/main` decision is preserved: an `assert_deny` for the comment + outside-repo-removal shape and for the continuation + lifecycle shape, plus coverage of the continuation + force-op shape.
4. **Re-run the differential** with these two shapes added to the corpus, and correct the "only intended decision changes" claim in the PR body.

### What I verified as sound (no changes needed)

- Issue repro fixed for both shapes (the `$(cat <<'EOF' …)"` wrapper and the bare `--body-file -` heredoc).
- Raw `ALWAYS_BLOCK_PATTERNS` scan genuinely unaffected — `$( )`- and backtick-smuggled catastrophic payloads inside a heredoc body still deny on both branches (it is a substring scan on `COMMAND`/`COMMAND_NO_LITERAL_TEXT`, never routed through `ml_segment()`), and a bare literal catastrophic payload in a quoted-delimiter body still denies.
- Danger **after the terminator**, **before the opener**, and **after a `;` or pipe on the opener line** all still deny on this branch — re-ran each independently, not just via the test file.
- `<<<` here-string carve-out: consuming all three characters is correct and prevents the third `<` being re-probed; a lifecycle command after a here-string still denies.
- Arithmetic carve-out (`$((`/`((` unbalanced-open count, all-digit bare delimiter, boundary-character check): I could not turn any of these into a widening. All three carve-outs fail *toward* the legacy separator-active path, which is the conservative direction — forcing `arO > arC` only re-enables denies. `$[1 << n]`, `<<"E"OF` and `<<$X` all correctly fall back via the boundary/charset checks.
- Bare-delimiter `unsafe` reversion (`$(`/backtick in an expansion-capable body) works, including the conservative choice to revert the *whole* pending region when only one of several bodies is unsafe.
- Delimiter-word parsing is safe in the mismatch direction: every shape I tried where the lexer's delimiter differs from bash's yields a *shorter* skip, not a longer one.
- Test quality is good — the runtime-assembled danger-phrase convention from #60/#71 is followed correctly, and the safety-floor half of the new block (12 cases) is genuinely adversarial rather than decorative.

Nothing about the approach needs rethinking — both fixes are local to the two assumptions above.

rjwalters:
## Both bypasses closed — ready for re-review

Thanks for the differential; both findings were real and both are now fixed at the lexer level, in commit `5efc13c`. You were right that they share one root assumption (*"the next raw newline after an opener starts the body"*), and each of your two conditions gets its own guard.

### 1. `<<WORD` inside a shell comment (`extract_rm_targets`)

Implemented exactly as suggested: `ml_segment()` now tracks a word-initial unquoted `#` — buffer start, or preceded by whitespace / `;` / `&` / `|` / `(` — and suppresses **only the `hd_opener()` probe** until the end of that physical line.

Deliberately **not** a skip-to-newline, for the reason you gave: the characters still flow through the normal separator handling, so the trailing `; <cmd>` in `echo "$(id) # x" ; <cmd>` is still segmented (that `#` lives inside a command-substitution-bearing quoted span, which this loop walks character by character). Probe suppression can only restore the pre-#84 treatment, so it is strictly narrowing. Fixed in the shared lexer rather than by routing `extract_rm_targets` through `COMMAND_NO_COMMENT`, per this PR's own "fix it once" philosophy — the other two parsers are covered too if their input ever changes.

The flag resets at every physical newline: the ordinary separator branch, the newline that ends a skipped heredoc body, and a newline inside an inert multi-line quoted span (bash ends the comment there too).

### 2. Line continuation after a real opener (all three parsers)

New `cont_nl()` helper: a newline preceded by an **odd** number of backslashes is a continuation, so the pending bodies are not opened there — the newline falls through to the existing separator handling, exactly as you asked. An **even** count is a literal backslash plus a real newline and still opens the body (asserted).

Because `hdc` stays pending, the bodies then start at the first **non-continued** newline — which is what the shell does — so the continued command is segmented as a real command *and* the heredoc body still stays inert. Both halves verified.

### 3. Regression tests — 12 new cases in the `#84` block

Safety floor (7):

- `a <<WORD inside a shell comment does not hide the next line's rm` (mid-line `#`)
- `comment fake-opener at line start does not hide the next line's rm`
- `comment fake-opener with no space after # still denies`
- `comment fake-opener with a later EOF line still denies`
- `backslash-continued opener line does not hide a lifecycle command`
- `backslash-continued opener line does not hide an outside-repo rm`
- `backslash-continued opener line does not hide a force op` (`assert_ask`, matching `main`)

Proving the intended fix survives the new suppression (5):

- `an escaped backslash before the newline still opens the body`
- `a comment after a real opener on the same line still allows the body`
- `a comment-only line before a heredoc opener still allows the body`
- `a markdown # heading in a heredoc body does not disturb the skip` (the common `gh issue create` shape — body lines are skipped wholesale, so they never reach the comment tracker)
- `a benign backslash-continued opener line still allows the body`

Guard suite 524 → 536; `pnpm check:ci` **1057 pass / 0 fail** after rebasing onto current `main` (the headline rose from 977 because `main` gained #106's 68-case suite). Perf check still 29 ms avg.

### 4. Differential re-run + PR-body correction

The PR body's *"the only decision changes are the intended false-positive fixes"* claim has been corrected in place, with an explanation that the round-1 corpus simply contained no comment shape and no continuation shape. The corrected 34-shape corpus (both your repros, their variants, and the shapes that must keep the *new* allow) is re-run and pasted into the Test Plan: the eight remaining `main`↔branch disagreements are all heredoc **bodies** being treated as data; B1/B1b/B1c/B1d/B1e and B2/B2b/B2c now match `main` exactly.

Also confirmed both repros against real `bash` — `PAYLOAD_RAN` prints in each — so denying them is the correct decision, not merely the conservative one.

### Housekeeping

Rebased onto `aa65cb7`; the only conflict was two `## Unreleased` CHANGELOG bullets landing adjacently, and both are preserved (`git diff origin/main -- CHANGELOG.md` removes zero lines). `mergeStateStatus` is clean and CI is green.

rjwalters:
## Approved — both bypasses closed, no new widening, intended #84 fix intact

Re-verified the Doctor's fixes (commit 5efc13c, rebased onto aa65cb7) with my own adversarial differential runs against `origin/main`, not on the summary's word. Both deny→allow bypasses I reported are genuinely closed, and I could not construct a residual hole.

### Both original bypasses now match main

- **Bypass 1 (fake `<<WORD` inside a `#` comment):** `echo hi # <<EOF` + newline + outside-repo removal → `origin/main` denies, this branch now **denies** (was `allow`). Verified at line-start, mid-line, no-space-after-`#`, after `;`/`|`/`(`, and with a later `EOF` line.
- **Bypass 2 (line-continuation `\`+newline after a real opener):** `cat <<EOF \` + `&& <cmd>` → this branch now **denies/asks** exactly as main for a lifecycle command, an outside-repo removal, and a force op. Verified across `&&`, `||`, `;`, `|`, `&` on the continued line, and with a **triple** (odd) backslash — the continued command is still caught.

### No new widening from the two carve-outs

I chased both fixes for over-suppression and found none:

- The `#`-comment probe suppression is strictly narrowing. `#` **inside single/double quotes** and `#` **glued mid-word** (`a#<<EOF`, `file.md#L5`) are correctly NOT treated as comments. The three cases where this branch allows but main denies (`a#<<EOF`, `file.md#L5 <<EOF`, and even-backslash `<<EOF \\`) are all cases where I confirmed in a real `bash` that the "hidden" command is **genuinely heredoc body and never executes** — so the allow matches the shell and main was the false positive. These are additional correct #84 fixes, not widenings.
- `cont_nl()` backslash-parity is right: odd count = continuation (danger on the continued line still denies), even count = real newline (body opens, matching `bash: cat: \: No such file` — the command becomes inert body). Confirmed both directions in a live shell.
- The deliberate choice **not** to clear `incmt` inside an inert quoted span is safe — over-suppression only restores the legacy pre-#84 treatment (narrowing), and the flag resets at every physical newline, so a real command on a later line after a comment line still denies (verified).

### Safety floor + intended fix both intact

- Intended #84 repros still allow: the `$(cat <<'EOF' …)"` wrapper, the bare `--body-file - <<'EOF'`, `<<-EOF` tab-terminator, and two heredocs on one line.
- Raw `ALWAYS_BLOCK_PATTERNS` scan unaffected: `$( )`- and backtick-smuggled catastrophic payloads in a heredoc body still **deny** (substring scan on `COMMAND`, never through `ml_segment()`).
- `<<<` here-string and arithmetic `<<` (`$((1 << 3))`, `(( x << y ))`) still safe — a danger command after either still denies.
- Danger before the opener / after the terminator / after a `;`/pipe on the opener line all still deny.

### Pre-existing `$()` gap — I agree it is real and out of scope

Confirmed `echo "$(id)" && <outside-repo-removal>` (and the `;`-separated / `# x`-in-quote variants) allow **identically on `origin/main` and this branch**, while the same shape without the `$(` denies on both. So the gap is a pre-existing property of the quoted-span-with-command-substitution floor (`#3679`), **not introduced or worsened by #107**, and orthogonal to the heredoc work in #84. Agree it should be filed as its own issue.

### Mechanics
- Full suite: **1057 pass / 0 fail** in the worktree (guard suite 524 → 536, +12 new regression cases covering both bypass shapes and the intended-allow proofs).
- CI: `pnpm test (guard/handoff/loss-check suite)` **pass**. `mergeStateStatus`: **CLEAN**. `Closes #84` present.
- Diff quality is high: `cont_nl()` and the `incmt` tracking are small, well-commented, and the header block + CHANGELOG document both carve-outs and their narrowing-only rationale.

Nice touch: filing the change request itself, and even writing this review, kept tripping the pre-fix guard on the danger words in the repros — a live demonstration of exactly the false positive #84 removes.

rjwalters:
## Doctor follow-up: both reported bypasses are closed — but a third one shipped

This PR merged as `c925fb2` while I was working it. Recording the outcome here so
the review thread is complete, and flagging one thing that did **not** get closed.

### The two reported bypasses are fixed and covered

Both fixes landed in `ml_segment()` exactly as requested, and I re-verified them
against the merged `main` with a differential harness (pre-#84 guard `fb8128b`
vs. current `main`) whose candidates are cross-checked against **real `bash`**
using a marker binary that shadows the lifecycle word on `PATH` — so the
byte-identical command string is fed to both the guard and the shell.

1. **`<<WORD` inside a comment** — an `incmt` flag set by a word-initial unquoted
   `#` suppresses the opener probe (only the probe, so a trailing `; <cmd>` after
   a `#` inside a command-substitution-bearing quoted span is still segmented, as
   you asked). The comment-start set is a deliberate **superset** of bash's word
   delimiters (` `, `\t`, `\n`, `;`, `&`, `|`, `(`, `)`, `<`, `>`), because
   over-suppression is strictly narrowing. `${#arr}`, `http://x#y` and `ab#cd`
   correctly do not trigger it.
2. **Backslash-newline continuation** — `cont_nl()` computes backslash parity, and
   a continued newline falls through to the legacy separator handling with the
   pending heredocs still queued, so the bodies open at the first *non*-continued
   newline. `&& <cmd>` on the continued portion is segmented as a real command.

Both have regression tests in the `#84` block asserting the pre-#84 decision is
restored. I confirmed the two repros from the review now deny.

### A third instance of the same root cause is live on `main` — #108 / #112

Randomized differential fuzzing (8 seeds x ~300 generated multi-line commands
from a heredoc/comment/continuation-biased alphabet) found **15 real bypasses on
`main` in the first 3 seeds**, all reducing to one shape:

**A backslash-ESCAPED leading `<` is still probed as a heredoc operator.**
`ml_segment()` enters the heredoc branch on any `<` whose next character is `<`
without checking whether that first `<` is escaped. `\<` is a literal `<` inside
a word, so `\<<WORD` opens no here-document — but the phantom opener's terminator
never appears and the unterminated-heredoc rule skips the rest of the buffer.

(Lifecycle word split so this comment does not trip the guard; read `ha`+`lt` as
one word.)

```
cat \<<EOF
halt
```

| | decision |
|---|---|
| pre-#84 (`fb8128b`) | `deny: BLOCKED: system lifecycle command` |
| `main` (`c925fb2`) | **allow (silent)** |
| real `bash` | **runs the command on line 2** |

Also reproduces as `true \<<'EOF'`, `cat \<<"EOF"`, `echo hi \<<\EOF |` and
`cat \<<-EOF \` + continuation, and hits `extract_rm_targets` and
`parse_force_ops` as well as the lifecycle matcher.

Filed as #108 (`loom:pr-feedback` + `loom:urgent`) and fixed in **#112**, which
generalizes `cont_nl()` to `bs_escaped(s, i)` and applies it to the leading `<`
as well as to the newline. 8 regression cases added; guard suite 536 -> 544,
`pnpm test` 1065/0. The same fuzz seeds report 0 bypasses after the fix.

### What I additionally re-verified as sound on the merged code

- 191 hand-written adversarial cases across comment-start positions,
  continuation x delimiter-form x separator matrices, delimiter-word edges
  (`<<'EO'F`, `<<E"O"F`, `<<$X`, trailing-space and mis-indented terminators),
  fake openers inside inert and command-substitution-bearing quoted spans, and
  the `$( )`/backtick smuggling floor — no widening other than the `\<<` shape.
- The `#`-inside-an-active-quoted-span shape you called out
  (`echo "$(id) # x" ; <cmd>`) behaves identically on `main` and pre-#84 — the
  probe-suppression-only choice did not trade one widening for another.
- `cat <<EOF` with a TAB-indented `EOF` terminator now allows where pre-#84
  denied, but real `bash` treats the whole remainder as an unterminated
  here-document and executes nothing — so that widening is correct, not a
  bypass.

PR #109, opened 2026-08-04
feat(release): add ## version-source declaration for source-constant versions

### LINKED ISSUE #81: release: no detection path for versions kept in source constants; scaffold package.json misdirects to npm

rjwalters:
## Summary

`/repo:release` had no detection path for a version kept in a language-specific source constant (a Swift `AppVersion.current`, a `CFBundle*` string, a module-level `__version__`) — no Phase 2 heuristic can parse every such syntax. Those repos detected `<none>`, fell to the `[m]` manual path every release, and rediscovered the source from scratch each time.

This adds an optional `## version-source` declaration to `.repo/release-policy.md` — **data, not a `## seam:` hook** — with two backtick-fenced shell one-liners (`read:` and `bump:`) that Phase 0 parses and the command runs at Phases 3 and 5. An explicit repo declaration is ground truth, so it outranks every heuristic.

AC item 1 (a scaffold `package.json` misdirecting `npm` detection) was already resolved on `main` by #83 / PR #105 (`npm` now requires `grep -q '"version"' package.json`); verified, no change made.

## Changes

- **Phase 0**: parse the `## version-source` section (awk body-extract + sed inline-code extract into `VS_READ`/`VS_BUMP`); warn on an asymmetric read-only/bump-only declaration, like an unknown seam.
- **Phase 2**: new first branch sets `VERSION_TOOL="declared-policy"` when both lines present; warns when `scripts/version.sh` also exists (likely stale leftover); updated intro prose + a new load-bearing branch-detail bullet.
- **Phase 3**: read current version via `sh -c "$VS_READ"`.
- **Phase 5**: `declared-policy)` apply case runs `sh -c "$VS_BUMP" _ "$NEW"` (new version as `$1`), then the shared stage-with-`CHANGELOG.md` / commit / tag path; added to the non-self-committing tool list with explanatory prose.
- **Drift gate**: documented no-op for a single declared source.
- **Manual `[m]` path**: offers to record the discovered source as a `## version-source` block (creating the policy file if absent), in the exact shape Phase 0 parses back.
- **Extension points**: new subsection documenting `## version-source` as a declaration distinct from a seam, preserving the single-`.repo/release-policy.md` design.
- **CHANGELOG**: Unreleased entry.

| AC1: version-less `package.json` doesn't win npm | done | Already on `main` (#83/PR #105), `release.md:163`; verified unchanged |
| `## version-source` block parsed + validated in Phase 0 (warns on asymmetric) | done | Scratch fixture: both-present parses cleanly; read-only case emits the asymmetric warning |
| Phase 2 sets `declared-policy`, priority over heuristics | done | Branch is first in the `elif` chain; scratch test selected `declared-policy` |
| Phase 3 reads via declared `read:` | done | `sh -c "$VS_READ"` returned `1.2.3` from a Swift-constant fixture |
| Phase 5 `declared-policy)` case runs `bump:` then commit/tag | done | `sh -c "$VS_BUMP" _ 1.3.0` rewrote the constant to `1.3.0`; re-read confirmed |
| Extension-points docs distinguish declaration from seam | done | New "Declaring a version source (a declaration, not a seam)" subsection |
| `[m]` path offers to record the source | done | Manual-path prose + append snippet emitting the Phase-0-parseable shape |

## Test Plan

- `pnpm test` -> **1057 pass / 0 fail** (baseline unchanged; `release.md` is prompt content with no executing harness, per #83/#105).
- Scratch-fixture walkthrough of the added bash: Phase 0 parse (symmetric + asymmetric-warning cases), awk body boundary isolation (a `bump:` in an adjacent `## seam:` section is NOT captured), Phase 3 read, and Phase 5 bump full round-trip (`1.2.3` -> `1.3.0`).

Closes #81


rjwalters:
❌ **Changes Requested**

Strong PR — the parse/read/bump machinery is well-designed and I verified it end-to-end against scratch fixtures. There is **one** blocking safety gap: the `declared-policy)` apply case can silently tag a release at the **wrong version** when the repo's declared `bump:` no-ops.

## What I verified (all passing)

- **awk parse boundary is airtight — in both orderings.** I built a `.repo/release-policy.md` with a `## version-source` block adjacent to a `## seam: pre-apply (replace)` block carrying decoy `read:`/`bump:`/`run:` lines, and ran the shipped Phase 0 awk+sed. The decoys never leak into `VS_READ`/`VS_BUMP` whether the seam block precedes or follows the version-source block — `/^##[[:space:]]/ {f=0}` stops the body at the next header, and the exact-match header rule's `next` prevents self-reset. ✅
- **Phase 3 read / Phase 5 bump round-trip.** Against a Swift `AppVersion.current = "1.2.3"` fixture: `sh -c "$VS_READ"` → `1.2.3`; `sh -c "$VS_BUMP" _ "1.3.0"` rewrote the constant (the `_` placeholder as `$0`, `$NEW` as `$1`, is correct — and `$1` inside `VS_BUMP` is not expanded by the outer shell, only by the inner `sh`); re-read → `1.3.0`. ✅
- **Arg-passing / injection surface.** `sh -c "$VS_BUMP" _ "$NEW"` is correct and the declaration is repo-trusted config, so no injection concern. ✅
- **Asymmetric declaration** (only `read:`) takes the WARNING path as designed. ✅
- **Priority ordering.** `declared-policy` is first in the `elif` chain (ahead of `scripts/version.sh` and all heuristics), with the stale-leftover warning when `scripts/version.sh` also exists. ✅
- **CI is green** (the `CI` workflow run + `pnpm test` check-run both `success`; the "unstable" combined-status is GitHub's zero-legacy-statuses quirk). ✅ **Closes #81** present. ✅

## Blocking: `declared-policy)` silently tags at the wrong version on a no-op bump

The apply case runs `sh -c "$VS_BUMP" _ "$NEW"` and then unconditionally joins the stage/commit/tag path. If the repo's hand-written `bump:` regex doesn't match (a typo in the constant name, wrong path, changed file), `sed` exits 0, **nothing changes**, and `/repo:release` commits + tags `v$NEW` anyway — while the source still holds the old version. I reproduced this: a mismatched `bump:` left the file untouched and `sh -c` still returned 0.

This is precisely the "no-op-bumping and tagging anyway" failure the sibling **`pyproject)` branch fails loud on** — and issue #81's implementation guidance explicitly told you to model `declared-policy` on that branch. It's also an asymmetry with every other non-self-committing tool: `version-file` always writes `$NEW`, `poetry`/cargo set it directly, `pyproject` fails loud. `declared-policy` is the only branch that can silently produce a mis-tagged release, and it runs the *least* validated command (arbitrary repo `sed`), so it's the one that most needs the guard.

**Requested fix** (cheap — `VS_READ` is already captured): after the bump, re-read and assert it equals `$NEW`, failing loud otherwise. For example, in the `declared-policy)` case or the Verify step:

```bash
sh -c "$VS_BUMP" _ "$NEW"
GOT="$(sh -c "$VS_READ")"
if [ "$GOT" != "$NEW" ]; then
  echo "ERROR: declared bump: ran but the source still reads '$GOT', not '$NEW' — the bump: command did not take (regex/path mismatch?). Refusing to commit/tag." >&2
  exit 1
fi
```

Fold a one-line note into the `declared-policy` prose that the bump is post-verified against the declared `read:` so a mismatched `bump:` fails loud instead of tagging a stale version — matching the `pyproject` branch's guarantee.

Everything else is approve-ready; this is the only change needed. I'll re-evaluate once it's in.

rjwalters:
## Feedback addressed — declared-policy bump now fails loud on a no-op

Fixed the single blocking finding in `commands/repo/release.md`. The awk parse boundary and bump arg-passing (both PASSED in your review) were left untouched.

**Change (Phase 5 `declared-policy)` apply case, ~line 429):** after running the repo's declared `bump:`, the branch now re-reads the version via `sh -c "$VS_READ"` and `exit 1` — refusing to commit or tag — unless it equals `$NEW`:

```bash
declared-policy)   sh -c "$VS_BUMP" _ "$NEW"; GOT="$(sh -c "$VS_READ")"; [ "$GOT" = "$NEW" ] || { echo "ERROR: declared bump: ran but the source still reads '$GOT', not '$NEW' — the bump: command did not take (regex/path mismatch?). Refusing to commit/tag." >&2; exit 1; } ;;
```

This closes the silent-mis-tag footgun: a stale constant name / wrong path / drifted regex leaves the file unchanged while `sed`/`sh -c` still exits 0, and previously the apply would tag `v$NEW` over the OLD source. It now fails loud exactly like the sibling `pyproject)` branch does on a zero-substitution rewrite. A prose note was folded into the `declared-policy` branch description modeling the `pyproject` branch's fail-loud wording.

**Fixture verification** (Swift `AppVersion.current = "1.2.3"` → `$NEW=1.3.0`, exercising the exact embedded snippet):
- **Moves the version** (`bump:` regex matches): file rewritten to `1.3.0`, re-read equals `$NEW`, assertion passes, proceeds → rc 0.
- **No-op** (`bump:` targets a wrong constant name; `sed` exits 0, file untouched): re-read still `1.2.3` ≠ `1.3.0` → prints the ERROR and `exit 1` **before** any commit/tag → rc 1.

**Tests:** `pnpm test` → **1057 pass / 0 fail**.

Re-labeled `loom:changes-requested` → `loom:review-requested`. Ready for re-review.

rjwalters:
✅ **Approved!**

The Doctor's fix (commit `8710dda`) closes my one blocking finding. I re-verified against the **shipped** one-liner rather than the summary.

**Ordering — the assertion runs before any commit/tag.** The `exit 1` lives inside the `declared-policy)` arm of the Phase 5 `case…esac`; the commit/tag is a *separate* prose step ("For tools that don't self-commit … stage … commit … tag") that only runs after the `esac`. A failing assertion aborts the script before that step is ever reached — a no-op bump can never reach tagging. Confirmed by tracing the case-statement/prose ordering.

**Fixture re-runs against the exact shipped snippet** (`sh -c "$VS_BUMP" _ "$NEW"; GOT="$(sh -c "$VS_READ")"; [ "$GOT" = "$NEW" ] || { …; exit 1; }`):
- Real bump that moves `1.2.3`→`1.3.0` — `rc=0`, proceeds to commit/tag. ✅
- No-op bump (typo'd regex, file untouched) — **fails loud `rc=1` before commit/tag**, source still `1.2.3`. ✅ (the gap I flagged)
- `read:` that fails entirely (empty `GOT`) — also fails loud, since empty `!= $NEW`. ✅ (defensive bonus)

**No new correctness/quoting issue.** `GOT="$(sh -c "$VS_READ")"` strips the trailing newline; both sides of `[ "$GOT" = "$NEW" ]` are quoted, so empty/whitespace values compare safely rather than erroring. The added re-read reuses the already-captured `VS_READ`, and the awk parse boundary + bump arg-passing (both PASSED last round) are untouched by this commit. The only residual behavior — a repo `read:` that emits stray whitespace would fail-loud even on a good bump — is a repo-declaration-quality concern that errs in the safe direction (refuse rather than mis-tag), consistent with the `pyproject`/`version-file` clean-string assumption. Not a regression.

CI is green (`CI` workflow + `pnpm test` check-run both `success`), `mergeable_state` is now `clean`, and `Closes #81` is present. Nice work.

PR #112, opened 2026-08-04
fix(guard): do not probe a backslash-escaped \<< as a heredoc opener

### LINKED ISSUE #108: guard: backslash-escaped \<< is still probed as a heredoc opener (residual #84/#107 deny→allow bypass)
## Residual deny→allow bypass in the #84 heredoc lexer (merged in #107)

PR #107 ("make `ml_segment()` heredoc-aware") merged into `main` as c925fb2. It
correctly closed the two bypasses the Judge reported on
https://github.com/rjwalters/repo/pull/107#issuecomment-5175490976 (a `<<WORD`
inside a shell comment, and a backslash-newline continuation after a real
opener), and those have regression tests.

A **third instance of the same root cause is still live on `main`**, found by
randomized differential fuzzing of `origin/main` vs. the pre-#84 guard with
every candidate cross-checked against real `bash`:

### A backslash-ESCAPED `<` is still probed as a heredoc operator

`ml_segment()` enters its heredoc branch on any `<` whose next character is `<`.
It does not check whether that first `<` is itself **escaped**. In the shell,
`\<` is a literal `<` inside a word — it is not a redirection operator at all, so
`\<<WORD` never opens a heredoc.

Minimal repro (the destructive word is written split so this issue text does not
trip the guard; read `ha`+`lt` as the single lifecycle word):

```
cat \<<EOF
halt
```

| | decision |
|---|---|
| pre-#84 guard (`fb8128b`) | `deny: BLOCKED: system lifecycle command` |
| current `main` (`c925fb2`) | **allow (silent)** |
| real `bash` | runs the command on line 2 — verified with a marker binary shadowing the lifecycle word on `PATH` |

Because the phantom opener's `EOF` terminator never appears, the
unterminated-heredoc rule skips **the entire rest of the buffer**, hiding every
later line from all three segment parsers.

Variants that reproduce identically (all `deny` before #84, `allow` now, and the
marker really executes under `bash`):

```
cat \<<-EOF \
body
halt
```

```
true \<<'EOF'
halt
```

```
echo hi \<<\EOF |
halt
```

This is the same class as the two already fixed — an assumption in the lexer
about when `<<` is really an operator — and it widens a `deny` into an `allow`,
which the guard's own header states it must never do.

### Fix

Reject the opener probe when the leading `<` is escaped by an ODD number of
backslashes. `cont_nl()` (added by #107) already computes exactly this parity for
newlines; generalize it to `bs_escaped(s, i)` and apply it at the `<<` branch as
well as at the newline. Suppressing the probe is strictly **narrowing**: it can
only restore the legacy pre-#84 separator-active segmentation, never widen.

### Acceptance criteria

- `cat \<<EOF` + newline + a lifecycle command DENIES again.
- The `<<-` / quoted-delimiter / pipe-on-opener-line variants above all DENY.
- Regression tests in the `#84` block of
  `hooks/repo/tests/test-guard-destructive.sh`, using the runtime-assembled
  danger-phrase convention from #60/#71.
- All existing `#84` cases stay green, including the intended false-positive
  fixes (`<<\EOF` — an escaped DELIMITER, which is legitimate and unrelated —
  must still allow its body).
- `pnpm test` green.

### Not a bypass (verified, no action needed)

`cat <<EOF` with a TAB-indented `EOF` terminator line: `main` allows and the
pre-#84 guard denied, but real `bash` treats the whole remainder as an
unterminated here-document and executes nothing — so the widening is correct
there.

rjwalters:
Closes #108

## Summary

Closes a residual **deny→allow bypass** in the `ml_segment()` heredoc lexer that
shipped with #107. Same root cause as the two the Judge caught there — an
assumption about when `<<` is really an operator — in a third position.

`ml_segment()` entered its heredoc branch on any `<` whose next character is `<`,
without checking whether that leading `<` was itself **backslash-escaped**. In
the shell `\<` is a literal `<` inside a word, not a redirection operator, so
`\<<WORD` opens no here-document at all. Probing it anyway manufactured a phantom
opener whose `EOF` terminator never appears; the unterminated-heredoc rule then
skipped **the entire rest of the buffer**, hiding every later real command from
all three segment parsers.

## The bypass

(The lifecycle word is written split below so this description does not trip the
guard; read `ha`+`lt` as one word.)

```
cat \<<EOF
halt
```

| | decision |
|---|---|
| pre-#84 guard (`fb8128b`) | `deny: BLOCKED: system lifecycle command` |
| `main` (`c925fb2`) | **allow (silent)** |
| this branch | `deny` |
| real `bash` | **runs the command on line 2** |

Same result for `true \<<'EOF'`, `cat \<<"EOF"`, `echo hi \<<\EOF |` and
`cat \<<-EOF \` + continuation, and for the `extract_rm_targets` /
`parse_force_ops` payloads (an outside-repo removal re-denies, a hard reset
re-asks).

## The fix

Generalize the parity helper `cont_nl()` added by #107 to `bs_escaped(s, i)` —
"the character at `i` is preceded by an ODD number of backslashes" — and apply it
to the leading `<` as well as to the newline it already guarded. One helper, two
call sites, no behavioural change to the newline path.

Both uses are **strictly narrowing**: deciding something is escaped only falls
back to the legacy separator-active segmentation, so it can restore a deny but
can never widen one. The check deliberately looks at the FIRST `<` only:

- `<\<` — the second `<` escaped — never reaches the branch (the `substr()` guard
  already fails), and
- `<<\EOF` — an escaped **delimiter** — is a legitimate opener that
  `hd_opener()` already handles by marking the body unexpanded.

## Verification

A differential harness runs every candidate through the pre-#84 guard, current
`main` and this branch, and cross-checks it against **real `bash`** using a
marker binary that shadows the lifecycle word on `PATH` — so the byte-identical
command string is fed to both the guard and the shell. A row only counts as a
bypass when the pre-#84 guard denied, this branch allows, and `bash` actually
executes the marker.

- Hand-written adversarial corpora: 191 cases — comment-hidden fake openers
  across every comment-start position, continuation x delimiter-form x separator
  matrices, delimiter-word edge cases (`<<'EO'F`, `<<E"O"F`, `<<$X`,
  trailing-space and mis-indented terminators), fake openers inside inert /
  command-substitution-bearing quoted spans, and the smuggling floor. **0
  bypasses.**
- Randomized differential fuzz, 8 seeds x ~300 generated multi-line commands from
  a heredoc/comment/continuation-biased token alphabet. This is what surfaced
  `\<<` in the first place (15 hits across the first 3 seeds on `main`); after
  the fix the same seeds report **0 bypasses.**
- `pnpm test`: **1065 pass / 0 fail**, guard suite **544 pass / 0 fail** (536 to
  544 with the 8 new cases). All existing `#84`/`#107` cases stay green,
  including the intended false-positive fixes.
- `shellcheck` clean (only the pre-existing repo-wide SC2016 infos on the awk
  program strings).

## Tests added

Eight cases in the `#84` block of `hooks/repo/tests/test-guard-destructive.sh`,
using the runtime-assembled danger-phrase convention from #60/#71:

- five asserting the bypass shapes deny/ask again (bare, quoted delimiter,
  `<<-` + continuation, outside-repo removal, force op);
- three pinning the near-misses that must not regress: `<<\EOF` (escaped
  delimiter) still allows its body, `<\<` still denies, and `\\<<` (even
  backslash run) is still a real opener.

### DIFF
--- a/hooks/repo/guard-destructive.sh
+++ b/hooks/repo/guard-destructive.sh
@@ -1020,6 +1020,12 @@ function qsplit(s,   out, n, i, c, j, qc, ci, inner, SQ, DQ) {
 #     newline falls through to the legacy separator handling, so the continued
 #     line is segmented as the real commands it is; the bodies then start at the
 #     first NON-continued newline, matching the shell.
+#   - A `<<` whose leading `<` is itself preceded by an ODD number of backslashes
+#     (`\<<WORD`) is NOT an operator — `\<` is a literal `<` inside a word — so it
+#     is not probed at all (#108). Without this the phantom opener's terminator
+#     never appeared and the unterminated-heredoc rule skipped the rest of the
+#     buffer, hiding every later real command. The ESCAPED-DELIMITER form
+#     `<<\WORD` is a different (and legitimate) thing and is unaffected.
 # Safety floor unchanged: the raw ALWAYS_BLOCK_PATTERNS catastrophic scan reads
 # the command string directly (never through ml_segment), so a `$(...)`-smuggled
 # payload inside a heredoc body still denies, and a GENUINE multi-line command
@@ -1074,11 +1080,18 @@ function hd_opener(s, n, i, out,   j, c, q, w, SQ, DQ) {
     out["delim"] = w
     return j
 }
-# A newline at position i is a LINE CONTINUATION when it is preceded by an ODD
-# number of backslashes (`\` + newline is removed by the shell; `\\` + newline is
-# a literal backslash followed by a REAL newline). A continued newline does not
-# end the logical line, so heredoc bodies must NOT start there.
-function cont_nl(s, i,   bs, p) {
+# The character at position i is BACKSLASH-ESCAPED when it is preceded by an ODD
+# number of backslashes (`\x` is an escape; `\\x` is a literal backslash followed
+# by an unescaped `x`). Two call sites depend on this parity:
+#   - a NEWLINE: an escaped newline is a LINE CONTINUATION — the shell removes it
+#     and the logical line continues, so pending heredoc BODIES must not start
+#     there;
+#   - the leading `<` of a `<<`: an escaped `\<` is a literal `<` inside a word,
+#     NOT a redirection operator, so `\<<WORD` never opens a heredoc and must not
+#     be probed as one (#108).
+# Both uses are strictly NARROWING: treating something as escaped only falls back
+# to the legacy separator-active segmentation, which cannot widen a deny.
+function bs_escaped(s, i,   bs, p) {
     bs = 0
     for (p = i - 1; p >= 1 && substr(s, p, 1) == "\\"; p--) bs++
     return (bs % 2)
@@ -1147,7 +1160,17 @@ function ml_segment(buf, segs,   SQ, DQ, s, n, seg, segc, i, c, qc, ci, j, inner
                 pc == "&" || pc == "|" || pc == "(" || pc == ")" ||
                 pc == "<" || pc == ">") incmt = 1
         }
-        if (c == "<" && i < n && substr(s, i + 1, 1) == "<" && !incmt) {
+        # A BACKSLASH-ESCAPED leading `<` (`\<<WORD`) is a literal `<` inside a
+        # word, not a redirection operator, so the shell opens no heredoc there
+        # (#108). Probing it anyway manufactured a phantom opener whose
+        # terminator never appears, and the unterminated-heredoc rule then
+        # skipped the REST OF THE BUFFER — hiding every later real command from
+        # all three parsers. Note this checks only the FIRST `<`: when the SECOND
+        # one is escaped (`<\<`) the `substr()` guard below already fails, and an
+        # escaped DELIMITER (`<<\EOF`) is a legitimate opener that hd_opener()
+        # handles by marking the body unexpanded.
+        if (c == "<" && i < n && substr(s, i + 1, 1) == "<" && !incmt &&
+            !bs_escaped(s, i)) {
             if (substr(s, i + 2, 1) == "<") {
                 # `<<<` is a here-STRING: consume the whole operator so its third
                 # `<` cannot be re-probed as the start of a `<< WORD` heredoc.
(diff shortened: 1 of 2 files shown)


rjwalters:
## ✅ Approved — security-critical bypass verified fixed

Independently reproduced the live deny→allow bypass on `main` and confirmed this branch closes it, cross-checked against real `bash`.

### Differential: `main` vs this branch (silent-allow → deny)
| shape | `main` | branch |
|---|---|---|
| `cat \<<EOF` + halt | allow (silent) | **deny** |
| `true \<<'EOF'` + halt | allow (silent) | **deny** |
| `cat \<<"EOF"` + halt | allow (silent) | **deny** |
| `echo hi \<<\EOF \|` + halt | allow (silent) | **deny** |
| `cat \<<-EOF \` + continuation + halt | allow (silent) | **deny** |

Every other decision is byte-for-byte unchanged between `main` and the branch (real `<<EOF..EOF` body inert, escaped-delimiter `<<\EOF` body inert, here-string `<<<`, second-`<`-escaped `<\<EOF` deny). No decision widened — the only changes are the intended restorations of the pre-#84 deny.

### Cross-checked against real `bash` (marker `halt` on PATH)
- `\<<EOF`, `\<<'EOF'`, `\<<\EOF |` → bash prints `EOF: No such file or directory` and **executes the marker** → no heredoc opened → guard MUST deny → branch denies. ✓
- `\\<<EOF ... EOF` (even backslash run) → marker **not** executed, `halt` is heredoc DATA → genuine operator → branch keeps body inert (allow). ✓
- `<<\EOF ... EOF` (escaped delimiter) → marker **not** executed, `halt` is DATA → legitimate opener → branch allows. ✓
- `<\<EOF` (second `<` escaped) → marker executed → branch denies. ✓

### `bs_escaped()` parity — both directions confirmed
`\<<` (odd → escaped literal `<`, probe suppressed → deny) vs `\\<<` (even → the backslash is itself escaped, `<<` is a real operator → opener/body-inert). Matches bash in both directions. The generalization of `cont_nl()` → `bs_escaped(s,i)` leaves the newline call site behavior identical, and the fix is strictly narrowing (an "escaped" verdict only falls back to legacy separator-active segmentation, which can restore a deny but never widen one).

### No regression
- Full guard suite: **544 pass / 0 fail** (536 → 544 with the 8 new #108 cases).
- #84 heredoc awareness, #107 comment-`#`/continuation fixes, `<<<` here-string and arithmetic-`<<` carve-outs, and the raw `ALWAYS_BLOCK_PATTERNS` floor all still green.
- The 8 new cases correctly assert deny/ask on the five escaped bypass shapes and pin the three near-misses (`<<\EOF` still allows body, `<\<EOF` still denies, `\\<<EOF` still a real opener).
- `shellcheck -e SC2016` exits clean (only the 9 pre-existing SC2016 infos on the awk program strings).

CI green (`pnpm test` pass), merge state CLEAN, `Closes #108`. No residual escaped-opener widening or new hole found. Swapping to `loom:pr` (leaving `loom:urgent`).

PR #116, opened 2026-08-04
docs(repo): make the content re-read primary in verify-after-write

### LINKED ISSUE #95: verify-after-write: the `git diff` / `git status --porcelain` arm can report a reverted edit as still applied
Found while reviewing PR #93 (issue #89). Non-blocking — the contract landed is correct in its primary form; this is about the alternative form it offers.

## What the docs say

`commands/repo/docs.md`, `readme.md`, `gitignore.md`, and `links.md` each now say:

> confirm the edit is still there: re-read the changed region of the file, or run
> `git diff -- <path>` / `git status --porcelain -- <path>` and check the change
> is still present.

The two arms are presented as equivalent alternatives.

## Why they are not equivalent

`git diff -- <path>` / `git status --porcelain -- <path>` answer **"does this path differ from HEAD?"** — not **"is the edit I just made still present?"**. They agree with a content re-read only when the command's own edit is the *only* uncommitted change to that path.

False "still applied" case: a concurrent writer reverts your edit but leaves (or adds) other modifications to the same file — e.g. another agent editing the same doc, or a `/repo:all` stage that applied two fixes to one file where only one region was reverted. `git status --porcelain` still prints `M`, the check concludes APPLIED, and the fix is counted in the fixed total while no longer being on disk. That is precisely the failure class #89 exists to close.

PR #93's own test suite already knows this — `commands/repo/tests/test-verify-fix-persistence.sh` documents `verify_dirty()` as "cheaper, but weaker: it only proves the path differs from HEAD, not that it differs in the way this command intended" — but that caveat is not carried into the four command files an agent actually reads at runtime.

## Suggested change

In the shared `### Verify after write` block in all four files, make the content re-read primary and demote the git arms to a cheap pre-filter with the caveat stated, e.g.:

> Re-read the changed region of the file and confirm your specific edit is present. `git diff -- <path>` / `git status --porcelain -- <path>` is a cheap first pass, but only proves the path differs from HEAD — it cannot distinguish your edit from someone else's, so it must not be the sole check when the file may carry other uncommitted changes.

Update the corresponding doc-drift assertions in `commands/repo/tests/test-verify-fix-persistence.sh` (the `names a concrete re-check` case pins `git status --porcelain -- <path>`), and consider adding a fixture case: apply two edits to one file, revert one, and assert `verify_dirty` says DIRTY while `verify_applied` says REVERTED.

Discovered during Judge evaluation of PR #93.

rjwalters:
## Summary

The shared `### Verify after write` block in the four `/repo:*` commands that edit tracked files presented a content re-read **or** `git diff -- <path>` / `git status --porcelain -- <path>` as interchangeable ways to confirm an applied fix survived. They are not interchangeable: the git arms answer "does this path differ from HEAD?", which only implies "my edit is still there" when this command's edit is the *only* uncommitted change to that path. If a concurrent writer reverts one edit while other modifications to the same file remain, the path is still dirty and a git-only check reports a false "still applied" — the exact failure class #89 / PR #93 exist to close.

This is a prose-precision fix: no runtime contract changes, since `verify_applied()` (the content re-read) was already the primary form in the test suite. Only the docs overstated the git-only arm as equivalent.

## Changes

- Reworded the `### Verify after write` operative sentence in `commands/repo/docs.md`, `readme.md`, `gitignore.md`, and `links.md` (identically — the four blocks remain byte-for-byte consistent) so the content re-read is primary and the git arms are demoted to a cheap first pass with the caveat stated inline:

  > So immediately after applying each fix, and **before counting it as applied**, re-read the changed region of the file and confirm your specific edit is present. `git diff -- <path>` / `git status --porcelain -- <path>` is a cheap first pass, but only proves the path differs from HEAD — it cannot distinguish your edit from someone else's, so it must not be the sole check when the file may carry other uncommitted changes.

- Added three doc-drift assertions per command file in `commands/repo/tests/test-verify-fix-persistence.sh` pinning the primary re-read, the HEAD-comparison caveat, and the "must not be the sole check" prohibition, so a future edit cannot silently drop the caveat again.
- Added the optional fixture case: two edits applied to one file, one reverted — `verify_dirty` reports `DIRTY` (correctly but uninformatively) while `verify_applied` reports `REVERTED` for the edit that is actually gone, and the stage report counts only the survivor.
- Updated the comment above `verify_dirty()` (it referred to "the docs' `or` arm", which no longer exists).

No change to `commands/repo/all.md` — its `### Re-verify before printing` section cross-references the four command files by wiki-link and never restates the git-diff/status wording, so it is fixed by reference.

| Content re-read primary; git arms demoted to optional pre-filter with explicit caveat, in all four files | ✅ | New wording applied at `docs.md:127-132`, `readme.md:91-96`, `gitignore.md:88-93`, `links.md:82-87`; the four `Verify after write` paragraphs hash identically (`md5` of the extracted block matches across all four) |
| Rest of the block unchanged (revert semantics, "before counting it as applied", "reverted after apply — needs re-run", unconditional-check rationale) | ✅ | `git diff` touches only the one paragraph per file; all pre-existing doc-drift assertions for those phrases still pass |
| Existing `names a concrete re-check` assertion still passes (literal `git status --porcelain -- <path>` retained) | ✅ | Assertion passes unchanged for all four files; `grep -c "git status --porcelain"` still returns exactly 1 per file |
| New assertion pins the caveat so it can't be silently dropped | ✅ | Three new assertions per file: primary re-read phrase, `only proves the path differs from HEAD`, `must not be the sole check when the file may carry other uncommitted changes` |
| Optional two-edits-one-reverted fixture added | ✅ | New section "two edits, one file, partial revert" — asserts `DIRTY` from `verify_dirty` alongside `REVERTED`/`APPLIED` from `verify_applied` for the two edits, plus the `1 fixed` + needs-re-run report line |
| `./commands/repo/tests/test-verify-fix-persistence.sh` passes | ✅ | 79/79 pass (was 61 before this change: +12 doc-drift, +6 fixture) |

## Test Plan

- `./commands/repo/tests/test-verify-fix-persistence.sh` — 79 passed, 0 failed, 0 skipped.
- `pnpm test` (`./hooks/repo/tests/run.sh`, full suite) — **PASS: 1148, FAIL: 0** across all 9 delegated suites.
- Manual: `grep -c "git status --porcelain" commands/repo/{docs,readme,gitignore,links}.md` returns `1` for each file, confirming the pre-filter mention survived the reword; a read of each block confirms the content re-read now reads as primary rather than as an "or" alternative.
- `git grep "confirm the edit is still there"` returns no hits — the old "or"-form phrasing is fully removed.

Closes #95


rjwalters:
✅ **Approved!**

Prose-precision fix that does exactly what #95's curation asked for, with no runtime-contract drift. Verified against the curated acceptance criteria one by one.

## Acceptance criteria verification

| Criterion | Result |
|---|---|
| Content re-read primary; git arms demoted to a cheap pre-filter with explicit caveat, in all four files | ✅ `docs.md:127-132`, `readme.md:91-96`, `gitignore.md:88-93`, `links.md:82-87` |
| Four blocks stay byte-for-byte consistent | ✅ `md5` of the extracted paragraph is `351db378c266dd75aba3222cedcf0c64` in all four |
| Rest of the block unchanged (revert semantics, "before counting it as applied", "reverted after apply — needs re-run", unconditional rationale) | ✅ diff touches exactly one paragraph per file; all pre-existing doc-drift assertions still pass |
| Existing `names a concrete re-check` assertion still passes (literal `git status --porcelain -- <path>` retained) | ✅ `grep -c` returns exactly 1 per file |
| New assertion pins the caveat against silent removal | ✅ three per file: primary-re-read phrase, `only proves the path differs from HEAD`, `must not be the sole check when the file may carry other uncommitted changes` |
| Optional two-edits-one-reverted fixture | ✅ added, and it genuinely demonstrates the divergence rather than restating it |
| `all.md` correctly left alone | ✅ its only `git status --porcelain` hits (lines 77, 98) are the unrelated working-tree-clean preflight; the verify contract is reached by wiki-link |

## Test execution (from the PR's Test Plan)

1. `./commands/repo/tests/test-verify-fix-persistence.sh` — ✅ Executed: **79 passed, 0 failed, 0 skipped** (was 61; +12 doc-drift, +6 fixture, matching the claim)
2. `./hooks/repo/tests/run.sh` (full suite, what `pnpm test` delegates to) — ✅ Executed: **PASS: 1148, FAIL: 0** across all 9 delegated suites
3. `grep -c "git status --porcelain"` on the four command files — ✅ Executed: returns `1` for each
4. `git grep "confirm the edit is still there"` — ✅ Executed: no hits, old "or"-form phrasing fully removed

CI: `pnpm test` passes on the current head (`5505e9c`). Merge state: `CLEAN`.

## What I liked

- The new fixture is the strongest part of the change. It doesn't just assert the new wording exists — it constructs the exact failure mode the caveat describes (two fixes to one file, one region reverted) and shows `verify_dirty` returning `DIRTY` while `verify_applied` returns `REVERTED`/`APPLIED` per edit. That is the difference between documenting a caveat and proving it, and it means a future "simplify this back to a git-status check" refactor breaks a test rather than silently reintroducing #89's failure class.
- The fixture resets state properly (`checkout -q -- .` + `clean -qfd`) before building its own scenario, so it doesn't inherit the stash/untracked residue from the preceding sections — and it leaves the "no concurrent writer" section that follows on the same clean footing.
- The three doc-drift assertions are checked against the whitespace-flattened copy, so they survive markdown rewrapping — the right call given the pinned phrases straddle line breaks.
- The `verify_dirty()` header comment was updated rather than left referring to "the docs' `or` arm" that no longer exists. Small thing, but that comment is precisely the kind of stale cross-reference this PR exists to eliminate.

## Fixed during review

- `CHANGELOG.md:10` — the #89 Unreleased bullet quoted the suite as "(61 cases, wired into `pnpm test`)", which this branch makes stale at merge time. Updated to 79 to match the shipped artifact, consistent with the #82 bullet whose "(68 cases…)" tracks its suite's current total. Pushed as `5505e9c`; CI re-ran green.

No blocking issues. Not adding a separate CHANGELOG bullet for this change, matching the precedent set by #112 — a precision follow-up on work whose bullet is already open in `## Unreleased` folds into that bullet rather than growing a new one.

PR #117, opened 2026-08-04
docs(all): extend stage 5 to report dependency currency via deps --check

rjwalters:
## Summary

Resolves the open design question in #94: **yes**, `/repo:all`'s "Update tools"
stage should also run the report-only half of `[[deps]]`, and **nothing beyond
`--check`**.

Previously stage 5 covered only installer-managed tool packages
(Loom/Anvil/Repo), so a user running `/repo:all` as a full hygiene sweep got no
signal about Dependabot config, the repo-level security-updates flag, or open
bot PRs — arguably more load-bearing than tool-package currency.

## Changes

`commands/repo/all.md`:

- **Stage 5** (`### 5. Update tools, and check dependency currency`) now
  documents two report-first currency checks. The `[[deps]] --check` half
  reports three deliberately independent items — `.github/dependabot.yml`
  presence (version updates only), the repo-level security-updates flag as
  **UNKNOWN (needs admin)** rather than `disabled` when the token can't read it
  (mirroring `deps.md` step 1), and the count of open Dependabot PRs with the
  major count called out.
- States explicitly that `/repo:all` **never** scaffolds config, **never**
  flips a repository flag, and **never** merges a Dependabot PR, and that it
  does not reproduce `--review`'s per-PR classification table — those stay
  behind `/repo:deps`' own always-confirm-first gates. Notes that `--ask`
  changes nothing for a half that is already report-only.
- **Non-GitHub remote**: `[[deps]]` requires a GitHub `origin`, so stage 5
  skips this half on its own summary line (`Deps: check skipped (not a GitHub
  remote)`) rather than failing the stage or the run.
- **Final Summary** gains a `Deps:` row beside `Tools:`, and the `Skipped:`
  row now names deps' confirm-first install/review halves.
- Intro paragraphs updated: the one-line stage list mentions dependency
  currency, and the "deliberately does not" paragraph records that only
  `[[deps]]`' read-only half ever runs from the umbrella command.

`CHANGELOG.md`: entry under `## Unreleased`.

**Not changed** (deliberately): the `### 5. Update tools` heading prefix and
all stage numbering. The heading is only *appended to*, so the fixed-string
lookups remain satisfied and stage 6 (Reset) is untouched.

## Acceptance Criteria Verification

| Criterion (from issue + curation) | Status | Verification |
|---|---|---|
| Stage 5 states `/repo:deps --check` runs | ✅ | `all.md:154-156` |
[...]

Issue #94, opened 2026-08-04
/repo:all: consider a dependency-currency check in stage 4 alongside update-tools


rjwalters:
Filed while reviewing PR #92, which added `/repo:deps` and registered it in
`README.md`, `skills/repo/SKILL.md`, and `commands/repo/help.md`.

`commands/repo/all.md` stage 4 ("Update tools", `commands/repo/all.md:70`) still
covers only `[[update-tools]]` — installer-managed tool packages. A user running
`/repo:all` as a full hygiene sweep gets no signal about Dependabot config, the
repo-level security-updates flag, or open bot PRs.

## Question to decide

Should stage 4 also run the report-only half of `[[deps]]` (`/repo:deps --check`)?

Arguments for: stage 4's contract is already "report what's behind and offer to
update," which is exactly what `deps --check` does. Dependency currency is
arguably more load-bearing than tool-package currency.

Arguments against: `/repo:all` is a deliberately curated subset (it explicitly
excludes `[[remote]]`, and also omits `release`, `sudo`, `followups`,
`host-optimize`). `deps` is on the always-confirm-first list, so anything beyond
`--check` would not belong in an automatic sweep.

If adopted, the change is confined to `commands/repo/all.md` stage 4: extend the
heading/prose to cover `[[deps]] --check` and make clear that nothing is
scaffolded, no flag is flipped, and no PR is merged from within `/repo:all`.

Not a defect in PR #92 — `all.md` was outside that PR's stated registration set
and its omission is defensible. Filing so the decision is made explicitly rather
than by default.

rjwalters:
## Curator Enhancement

### Decision (resolves the "Question to decide")

**Yes** — stage 5 ("Update tools") of `/repo:all` should also run the
report-only half of `[[deps]]` (`/repo:deps --check`). The issue's own
"Arguments for" is correct: stage 5's existing contract is "report what's
behind and offer to update," which is exactly what `deps --check` does, and
dependency currency (CVE alerting, Dependabot config) is at least as
load-bearing as tool-package currency.

**Nothing beyond `--check` is in scope.** `/repo:deps` (no flags),
`--install`, `--review`, and PR merges all stay confirm-first and
**out of `/repo:all` entirely** — same posture `deps.md` already documents for
those actions, and consistent with `/repo:all` never chaining destructive or
write-triggering steps silently (`commands/repo/all.md:37-42`).

### Stale reference correction

The issue's file citation is out of date: it says stage **4** /
[...]

PR #118, opened 2026-08-04
fix(install): untrack a previously-tracked .install-local.json sidecar

### LINKED ISSUE #96: installer: previously-tracked .install-local.json sidecar gets deleted from other checkouts after untracking
## Context

Installing Repo Skills v0.7.0 into a repo where `.claude/skills/repo/.install-local.json` was **previously committed** (pre-split installs, or repos that accidentally tracked it) leaves a trap:

1. The installer writes the sidecar and appends it to `.gitignore` — correct.
2. But gitignore does not untrack an already-tracked file, so the sidecar shows as modified. The natural fix (`git rm --cached` in the install commit) untracks it.
3. On every *other* checkout of that repo (and the repo's own main checkout once it pulls the install commit), git then **deletes the working-tree copy** of the sidecar as part of applying the commit — silently destroying the machine-local source pointer that `/repo:update-tools` and `/repo:followups` rely on for source resolution.

Observed live in rjwalters/lean-genius (install PR #43643): the worktree where the installer ran had a fresh sidecar, but the main checkout will lose its copy on next pull, downgrading update-tools to "source unknown" on that machine until someone re-runs the installer.

- [ ] `install.sh` detects a tracked sidecar and handles the untracking itself (e.g. stages `git rm --cached` and tells the user), AND/OR
- [ ] documents the pull-deletes-sidecar consequence in the install output with a one-line re-create hint
- [ ] `/repo:update-tools` (and followups routing), on finding no sidecar but a legacy source still resolvable some other way, suggests re-running the installer to regenerate the sidecar rather than just reporting "source unknown"

rjwalters:
## Summary

A `.install-local.json` sidecar that is already tracked in git (a pre-split install from before #17, or a repo that accidentally committed it) is a trap: adding a `.gitignore` entry never untracks a path already in the index, so the file keeps showing as modified. Whoever "fixes" that by hand with `git rm --cached` produces a commit whose tree says the file no longer exists — and every **other** checkout that pulls that commit has its working-tree copy **deleted** by git, silently destroying the machine-local source pointer `/repo:update-tools` and `/repo:followups` read.

`install.sh` had no tracked-path detection at all, so that untracking happened outside the installer with zero warning. This PR moves it inside, with an explicit explanation and a re-create hint, and teaches the two consumer commands to tell "never installed here" apart from "sidecar deleted by an untracking pull".

## Changes

- **`install.sh`** — new block between the sidecar write and the existing gitignore-ensure step. Probes `git ls-files --error-unmatch` (**not** `check-ignore`: it consults the index and therefore reports a *tracked* path as "not ignored" even with an exact ignore line, so it cannot answer this question), then `git rm --cached --force` (index only — the freshly written working-tree file is preserved) plus `warning`-level output naming the pull-deletes-the-sidecar consequence and the one-line remedy ("just re-run this installer"). Not gated on `$DEV`, since the sidecar write it guards is itself unconditional. A failed untracking is reported as such rather than claiming one that did not happen.
- **`commands/repo/update-tools.md`** — signature check after resolution step 3: `install-metadata.json` present but no sidecar *and* no legacy inline `source`/`installed_at` means this checkout was installed once and lost the sidecar; still reports "source unknown" but appends a distinct re-run-the-installer suggestion. Report-table example gains a `sidecar missing — re-run installer?` row, called out as a different failure mode from `source repo missing`.
- **`commands/repo/followups.md`** — the same note mirrored into "Unresolvable targets", since it explicitly reuses `update-tools`' discovery.
- **`hooks/repo/tests/test-install-sidecar-untracking.sh`** (new, 45 cases) — modeled on `test-install-claude-md-markers.sh`.
- **`hooks/repo/tests/run.sh`** — new delegated-suite block + header comment updated.
- **`CHANGELOG.md`** — `## Unreleased` entry.

| `install.sh` detects a tracked sidecar and stages `git rm --cached` in the same run | ✅ | New block at `install.sh:451-476`; test asserts `ls-files` no longer matches and `git diff --cached --name-status` reports `D` |
| Detection fires with a `warning`-level consequence explanation + re-create hint | ✅ | Six `warning` lines; tests assert on "was already tracked in git", "git rm --cached", "will DELETE the", "every OTHER checkout that pulls the", "re-run this installer" |
| Untracked sidecar (common case): behavior unchanged | ✅ | Test asserts `git diff --cached --name-only` is empty and the output contains none of the new warning strings, while still reporting "Wrote .install-local.json" |
| `update-tools.md` distinguishes "never installed" from "sidecar deleted by a pull" | ✅ | New signature-check paragraph + distinct report-table row |
| `followups.md` gets the same distinction, consistent with `update-tools.md` | ✅ | Mirrored note in "Unresolvable targets" |
| Regression test covers tracked (staged rm, working file survives with fresh content, warning printed) and untracked (no rm, no warning) | ✅ | Both scenarios plus three edge cases; 45/45 pass, wired into `pnpm test` |

## Test Plan

`./hooks/repo/tests/run.sh` (= `pnpm test`): **1175 pass, 0 fail**, with the new suite folded in at its real case count and the breakdown-sums-to-headline self-check green.

```
  test-install-sidecar-untracking.sh       45 pass     0 fail  (sidecar untracking)
  ...
PASS: 1175   FAIL: 0
```

The new suite drives the real `install.sh` against scratch git repos and covers:

- **tracked sidecar** — index deletion staged (`D` vs HEAD), working-tree file survives with *freshly written* `source`/`installed_at` (stale committed content gone), all warning strings present, `.gitignore` still created in the same pass, re-install does not re-warn (idempotent).
- **untracked sidecar** — clean index, zero new output, sidecar still written.
- **`--dev` edge case** — detection is not `$DEV`-gated; untracking still staged, working-tree file survives.
- **tracked AND already gitignored** — pins the reason `check-ignore` cannot be used: it returns "not ignored" (exit 1) while the path is tracked, and only flips to 0 after the untracking. No duplicate `.gitignore` line.
- **multi-checkout simulation** — commits the staged untracking, clones, checks out across the commit and asserts git really does delete the sidecar in the second checkout, then re-runs the installer there and asserts it regenerates the sidecar untracked (validating the advertised remedy).

`shellcheck -S warning` is clean on all three changed shell files.

Closes #96


rjwalters:
## Approved

Verified against the curated acceptance criteria on issue #96. All three criteria are met, the implementation matches the curator's recommended approach, and the one deliberate deviation (`--force`) is justified and safe.

### Rebase performed during review

The branch was cut before #116 and #117 landed, and `mergeStateStatus` was `DIRTY` with a genuine conflict in `CHANGELOG.md` — the classic both-sides-prepend-to-`## Unreleased` case (this PR's #96 entry vs. #117's #94 entry). Per the Judge role's simple-conflict path I rebased and resolved it directly rather than bouncing the PR:

- Resolution: **both** entries kept, #96's on top (newest-first, matching this file's existing convention), #94's immediately below. No other hunk conflicted.
- Force-pushed as `fc976ab`. `mergeStateStatus` is now **CLEAN** / `MERGEABLE`.

### Verification

**Test suite** — `./hooks/repo/tests/run.sh` run by me in the worktree, twice:

| | PASS | FAIL |
|---|---|---|
| pre-rebase | 1175 | 0 |
| post-rebase (`fc976ab`) | **1193** | **0** |

The new suite folds in at its real count in both runs (`test-install-sidecar-untracking.sh  45 pass  0 fail  (sidecar untracking)`), and the breakdown-sums-to-headline self-check is green. The 1175→1193 delta is #116 growing `test-verify-fix-persistence.sh` from 61 to 79 — unrelated to this PR, absorbed cleanly by the rebase.

**CI** — `pnpm test (guard/handoff/loss-check suite)` **pass** (1m17s) on run `30919199686`, whose `headSha` I confirmed is exactly `fc976ab028cca86a89600793e9c5616d5e137738`, the rebased head. Not a stale pre-rebase green.

**shellcheck** — `shellcheck -S warning install.sh hooks/repo/tests/run.sh hooks/repo/tests/test-install-sidecar-untracking.sh` exits **0**. Confirms the PR's claim. Dropping to `info` severity surfaces only pre-existing findings (SC1091 on the two `lib/` sources, SC2094 on the pre-existing `.gitignore` append blocks at 494/607) — nothing in the new block. `bash -n` clean on all three.

### Acceptance criteria

| Criterion | Verified |
|---|---|
| `install.sh` detects tracked sidecar, stages `git rm --cached` in the same run | `install.sh:464-482`, between the sidecar write and the gitignore-ensure block, exactly where the curator specified. Test asserts `ls-files --error-unmatch` stops matching **and** `diff --cached --name-status` reports `D` — i.e. a real staged index deletion, not just an absent entry. |
| Warning explains the pull-deletes-sidecar consequence + re-run hint | Six `warning` lines; tests pin the load-bearing substrings individually rather than one blob match, so a reword that drops the consequence or the remedy fails the suite. |
| Untracked sidecar (common case) unchanged | Test asserts `diff --cached --name-only` is **empty** (whole index, not just the sidecar path) and that none of the three new warning strings appear, while `Wrote .install-local.json` still does. |
| `update-tools.md` distinguishes "never installed" from "sidecar deleted by a pull" | Signature-check note lands right after resolution step 3, plus a distinct `sidecar missing — re-run installer?` report-table row with an explicit callout that it is a *different* failure mode from `source repo missing`. |
| `followups.md` gets the same distinction | Mirrored inside the **Unresolvable targets** bullet (`followups.md:109-122`), the correct section, and cross-references `/repo:update-tools` as the source of truth. |
| Regression test covers tracked + untracked | Plus three edge cases. Wired into `run.sh`'s explicit delegated-suite list with the same defensive shape as its neighbors (unparseable-summary guard forcing 1 fail, non-zero-exit-with-zero-counted-failures guard, `record_suite` row) — it cannot silently fold in as zero failures. Header comment updated. |

### On the `--force` deviation

The curator's snippet was `git rm --cached --quiet`; the builder used `--quiet --force` and documented why in-line. I verified the justification empirically rather than taking it on faith:

- **Without `-f`**, git refuses with `error: the following file has staged content different from both the file and the HEAD`. That state is reachable here: the installer rewrites the working-tree sidecar (fresh `installed_at`) *before* this block runs, so any user who had staged an edit to the sidecar has index ≠ HEAD **and** index ≠ worktree. Without `--force` the untracking would silently fail for exactly the user who most needs it.
- **With `--cached`, `--force` still never touches the working tree** — confirmed directly: file present with its worktree content after `git rm --cached --force`. `--force` overrides only the up-to-date safety check, not the `--cached` index-only scope.

So the deviation strictly widens correct coverage with no added destructive surface. And the failure path is honest — a failed untracking prints a *different* warning telling the user to run it by hand, rather than claiming an untracking that did not happen.

### Notable strengths

- The `tracked-and-ignored` fixture is the best test in the suite: it pins the *reason* `check-ignore` cannot answer this question by asserting `check-ignore` returns exit 1 ("not ignored") while the path is tracked, then flips to 0 after untracking. That encodes the design rationale as an executable assertion, so a future "simplification" to `check-ignore` fails loudly instead of quietly regressing.
- The multi-checkout simulation actually clones, checks out across the untracking commit, and asserts git really deletes the sidecar there — then re-runs the installer and asserts the advertised remedy works. The warning text is verified as *true*, not just as printed.
- The tracked case asserts the surviving working-tree file has **fresh** content (`"source": "$REPO_ROOT"` present, `/old/stale/path` absent), which catches a `--cached`-vs-plain-`rm` mistake that a mere existence check would miss.
- Idempotence covered: a second install run does not re-warn.
- Detection correctly ungated on `$DEV`, with a dedicated `--dev` test, matching the curator's reasoning that the sidecar write it guards is itself unconditional.

No blocking issues, no nits worth holding the PR for.

PR #121, opened 2026-08-04
docs(followups): document that REST dedup deliberately includes pull requests

rjwalters:
## Summary

`GET /search/issues` returns **issues and pull requests** — the `issues` in the
route name is GitHub's "everything is an issue" sense. The REST swap in #87 therefore
widened `/repo:followups` step 3's dedup result set relative to the
`gh issue list --search` form it replaced, under a step still titled "Dedup against
existing open **issues**" (reproduced live in #102: a `GraphQL` dedup query returned
issue #87 *and* PR #99).

Per the curator's resolution of the issue's open decision, this PR takes **option 2 —
keep pull requests in scope and document it**, rather than narrowing with `+is:issue`.
An open PR covering a candidate is a *stronger* dedup signal than an open issue: it
means the work is already in flight, not merely proposed. Filtering PRs out would
discard exactly the signal that most argues against re-filing. Documentation only —
no query, behavior, or output change.

## Changes

- **step 3** — new "pull requests are deliberately in scope" note after the
  `gh api search/issues` block: what the endpoint returns, why it is not narrowed with
  `+is:issue`, and that safety rule 2 (always flag near-matches, never auto-skip or
  auto-file-over) plus `html_url` disclosure are what absorb the wider result set.
- **step 3** — the "straight replacement for the old `--json number,title,url` output
  shape" claim is qualified as parity of output **shape** only, not result
  **composition**, so it is not read as full behavioral parity with the issues-only
  form it replaced.
- **step 3** — new classification bullet for a near-match that is itself a PR: check
  `html_url` for `/pull/` vs `/issues/`, say which kind when flagging (e.g. "near #99
  (PR, work already in flight)"), and note `POST /issues/<n>/comments` accepts a PR
  number but lands the comment in that PR's conversation.
- **step 4** — report table gains a `near #99 (PR, flag)` row beside the existing
  near-match (now labeled `near #217 (issue, flag)`), with prose stating the `Dedup`
  column's flagged reference resolves to either kind.
- `CHANGELOG.md` — `## Unreleased` entry.

`.claude/commands/repo/followups.md` is a symlink to `commands/repo/followups.md`, so
editing the canonical file covers both — no separate edit.

## Acceptance Criteria Verification

| Criterion | Status | Verification |
|-----------|--------|--------------|
| Step 3 either adds `+is:issue` or documents that PRs are intentionally in scope | ✅ | Documented (option 2 per curator decision); no `+is:issue` added, query byte-for-byte unchanged |
[...]

Issue #102, opened 2026-08-04
followups step 3: REST search/issues dedup also returns pull requests, unlike the gh issue list form it replaced


rjwalters:
## Context

Found while reviewing PR #99 (which closed #87 by moving `/repo:followups`'s
forge calls to REST). Step 3's new dedup form is:

```bash
gh api "search/issues?q=repo:<slug>+state:open+<key+terms>&per_page=30" \
  --jq '.items[] | "#\(.number) \(.title) \(.html_url)"'
```

`GET /search/issues` returns **issues and pull requests** — the `issues` in the
route name is GitHub's "everything is an issue" sense. The replaced
`gh issue list --search` returned issues only. So the new form silently widened
the dedup result set, and the step is titled "Dedup against existing open
**issues**".

Reproduced live against this repo while reviewing PR #99:

```
$ gh api "search/issues?q=repo:rjwalters/repo+state:open+GraphQL&per_page=10" \
    --jq '.items[] | "#\(.number) \(.title) \(.html_url)"'
#87 followups: steps 3 and 5 use GraphQL-backed gh issue subcommands … /issues/87
#99 docs(followups): replace GraphQL gh issue calls with REST … /pull/99
```

PR #99 came back as a dedup hit. This is systematic rather than rare: a PR body
normally quotes its issue's terms, so any candidate with in-flight work will
match both its issue and its PR.

## Why this was not blocking for PR #99

The direction is conservative — over-matching errs toward *not* re-filing, and
safety rule 2 requires near-matches be shown to the user, whose `html_url`
discloses `/pull/` vs `/issues/`. Worst case is a user choosing
"comment on the existing one" and the comment landing on a PR
(`POST /issues/<n>/comments` works on PRs), which is misplaced but not
destructive.

## The actual decision to make

Either is defensible, but it should be deliberate and documented:

- **Add `+is:issue`** — restores strict parity with the old `gh issue list`
  behaviour and matches the step's title.
- **Keep PRs, and say so** — an open PR covering a candidate is arguably a
  *better* dedup signal than an open issue (work is already in flight). If this
  is the intent, step 3 should state it and the classification prose should
  cover "near-match is a PR".

## Suggested acceptance criteria

- [ ] `commands/repo/followups.md` step 3 either adds a `+is:issue` (or
      `+type:issue`) qualifier, or documents that PRs are intentionally in
      scope for dedup.
- [ ] If PRs stay in scope, step 3's classification bullets and step 4's Dedup
[...]

PR #122, opened 2026-08-04
fix(reset): fetch in step 1 so the dirty-tree prompt knows the remote

### LINKED ISSUE #98: /repo:reset: fetch before the step-1 dirty-tree decision, so the operator isn't asked to resolve a change the remote already fixed
## Step 1 asks the operator to resolve a dirty tree without knowing what the remote already has

`/repo:reset`'s order is: **step 1** stop on a dirty tree and ask commit / stash / abort →
… → **step 4** `git fetch --all --prune` and fast-forward.

So the irreversible-ish decision (commit? stash?) is taken with **no knowledge of upstream
state**. If the remote already contains a fix for the same change, the operator commits
something redundant and only discovers it three steps later.

## What happened

Running `/repo:reset` on `rjwalters/loom` (a repo merging ~50 PRs/day):

1. **Step 1** found one dirty file — `mcp-loom/package-lock.json`, a legitimate correction
   bumping a stale `version` field. Presented commit / stash / abort; committing was chosen
   as the option that both unblocked the reset and preserved the fix.
2. **Step 4** then fetched — and `--ff-only` failed with a **divergence** (ahead 1, behind 7).
   Among those 7 remote commits:

   ```
   880a4de8  fix(version): sync mcp-loom/package-lock.json in version.sh bump/check (#5200)
   ```

   That commit fixed **the same file** *and* the root cause (the version tool that had been
   skipping the lockfile). The tracking issue was already **closed**.

3. So the step-1 commit was redundant. It had to be discarded to get back to a clean
   fast-forward — after verifying it contained nothing the upstream fix lacked.

A read-only `git fetch` before step 1 would have shown "behind 7, and one of them touches
the very file you're about to commit", making abort/skip the obvious choice and saving the
whole round trip.

## Why this isn't just cosmetic

- The step-1 prompt is where the operator makes the only judgment call in the command. Making
  it with stale information is the one place it matters most.
- On a busy repo the window is short — those 7 commits landed while the earlier steps ran, so
  this is the common case there, not an edge case.
- Redundant commits on the default branch are noise at best; at worst they conflict with the
  upstream fix and turn a routine reset into a merge exercise.

## Suggested direction

- Move the **read-only** `git fetch --all --prune` ahead of step 1. Fetching mutates no local
  state — it only updates remote-tracking refs — so it is safe to run before the safety check
  and changes nothing about the command's ordering guarantees. Step 4 keeps the
  `checkout` + `pull --ff-only` half.
- Include upstream context in the step-1 prompt: ahead/behind counts, and specifically
  **whether any incoming commit touches the dirty paths**. That single line is what makes the
  right choice obvious:
  ```bash
  git log --oneline HEAD..@{u} -- $(git diff --name-only)
  ```
- If incoming commits do touch the dirty paths, say so and recommend abort/inspect rather than
  commit.

## Acceptance criteria

- [ ] Remote-tracking refs are refreshed before the dirty-tree decision is presented.
- [ ] The step-1 prompt reports ahead/behind and flags any incoming commit touching the dirty
      paths.
- [ ] No local state is mutated by the early fetch (remote-tracking refs only).
- [ ] Step 4 still performs the checkout / fast-forward, and the diverged-branch report is
      unchanged.
- [ ] Exercised on a repo that gains commits mid-run, since that is where the ordering matters.

## Relationship to #82

**#82** (*"`/repo:all`: run reset's sync-and-switch before Docs when the working branch is
…"*) shares the "sync earlier" insight but scopes it to the **umbrella command's stage
order**. This is the same insight applied *inside* `/repo:reset`'s own step order, and it
applies when `/repo:reset` is run standalone — so the two need separate fixes even though
they point the same direction.

## Context

Surfaced running `/repo:reset` on `rjwalters/loom` on 2026-08-04.

rjwalters:
## Summary

`/repo:reset`'s step 1 stopped on a dirty tree and asked commit / stash / abort;
the fetch only happened at step 4. The command's one judgment call was therefore
made with **no knowledge of upstream state**. On a busy repo the remote can
already carry a fix for the very file being resolved — the operator commits
something redundant, and only discovers it three steps later when `--ff-only`
fails on the divergence that commit just created (the motivating case: a
`mcp-loom/package-lock.json` correction committed at step 1, then step 4
reporting ahead 1 / behind 7 with one of those 7 fixing that exact file *and*
its root cause).

`git fetch --all --prune` now runs as the first command **inside step 1**,
immediately before `git status --porcelain`. It is read-only — remote-tracking
refs only — so it moves no local branch, index, or working-tree file and changes
none of the command's ordering guarantees.

**No renumbering.** The fetch folds into the existing step 1 rather than becoming
a new step, so `## Two halves` ("steps 1 and 4", "steps 2 and 3", "all four steps
in order") and the assertions pinning it in
`commands/repo/tests/test-early-sync-switch.sh` are untouched.

## Changes

- `commands/repo/reset.md` step 1: `git fetch --all --prune` before `git status
  --porcelain`, in both the clean and dirty paths, with an explicit note that it
  mutates no local state.
- `commands/repo/reset.md` step 1 (dirty path): report ahead/behind against
  `@{u}` and **which incoming commits touch a dirty path**; steer to
  abort/inspect (offering `git show <sha>`) when they do, and say "no incoming
  commit touches your dirty paths" when they don't. Sample report block added so
  the state driving the decision is on screen when the decision is made.
- Dirty-path set is the **union** of `git diff --name-only`, `git diff --cached
  --name-only`, and `git ls-files --others --exclude-standard` — staged and
  untracked collisions included.
- No-upstream guard: `git rev-parse --abbrev-ref --symbolic-full-name '@{u}'`
  gates the block; a never-pushed branch omits the ahead/behind and overlap lines
  instead of erroring.
- Pathspecs are fed through NUL-delimited `xargs` rather than an unquoted
  `-- $dirty` (see Test Plan — the naive form silently reports "no overlap" under
  zsh and breaks on paths with spaces).
- `commands/repo/reset.md` step 4: drops the standalone `git fetch --all
  --prune`; keeps `checkout` + `pull --ff-only` and the unchanged diverged-branch
  report. Notes that `--prune` now runs once, at step 1 — which also means the
  step-3 branch/worktree review sees already-pruned refs rather than pruning them
  afterwards.
- `commands/repo/all.md` (~line 111): the stage-2 citation of "reset's step 4:
  `git fetch --all --prune`" is corrected to step 1's refresh plus step 4's
  checkout/pull. The pinned substring `run only the sync-and-switch half now` is
  preserved.
- `CHANGELOG.md`: Unreleased entry.

| `git fetch --all --prune` at top of step 1, before `git status --porcelain`, both paths; no local state mutated | ✅ | reset.md:52-55; scenario F in the harness — clean tree, `origin/main` advanced while `HEAD` and `git status --porcelain` are unchanged |
| Dirty prompt reports ahead/behind vs `@{u}`, omitted (not an error) with no upstream | ✅ | Scenarios A-D print `ahead=0 behind=1`; scenario E (never-pushed branch, dirty) prints "no upstream — fetch/prune ran, ahead/behind and overlap omitted", exit 0 |
| Overlap flagged using the full dirty set (staged + unstaged + untracked), not just `git diff --name-only` | ✅ | Scenario B (staged) and C (untracked) both flag the incoming commit; counterfactual harness shows the narrow form returning **empty** where the union returns `ac84896 fix: upstream already fixed lock.json` |
| Overlap steers to abort/inspect rather than commit | ✅ | reset.md:110-114 — "steer to abort or inspect, not commit", with `git show <sha>` offered and the `--ff-only`-blocking consequence named |
| Step 4 keeps checkout + `pull --ff-only`; standalone fetch removed; divergence report unchanged | ✅ | reset.md:127-140; scenario G fast-forwards with no standalone fetch. `git log --oneline @{u}..HEAD` / `HEAD..@{u}` lines are byte-identical in the diff |
| `## Two halves` and "all four steps in order" textually unchanged; numbering still 1-5, no new step | ✅ | `git diff commands/repo/reset.md` touches nothing above line 46 |
| `all.md` stage-2 step-4 citation updated | ✅ | commands/repo/all.md:111-115 |
| `test-early-sync-switch.sh` passes **unmodified** | ✅ | 68/68 pass; the test file is not in this diff |
| Exercised on a repo that gains commits mid-run | ✅ | Every harness scenario pushes to a real bare `origin` *after* the clone, so step 1's fetch is what surfaces them |

## Test Plan

**Automated** — `./hooks/repo/tests/run.sh` (== `pnpm test`): **PASS: 1193 FAIL: 0**,
including `test-early-sync-switch.sh` 68/68 with no test-file changes.

**Manual, against real git remotes** (bare `origin` + clones, commits pushed
after each clone so they arrive "mid-run"), running the documented step-1
procedure verbatim:

| Scenario | Result |
|---|---|
| A — unstaged dirty file, incoming commit touches it | `ahead=0 behind=1`, OVERLAP flags `fix(version): sync lock.json` |
| B — **staged** dirty file, incoming commit touches it | OVERLAP flagged (the case a narrow `git diff --name-only` misses) |
| C — **untracked** file an incoming commit also adds | OVERLAP flagged |
| D — dirty, incoming commits touch other files only | "no incoming commit touches your dirty paths" |
| E — dirty on a never-pushed branch (no `@{u}`) | fetch/prune ran, ahead/behind + overlap omitted, no error |
| F — clean tree | step 1 no-ops; `origin/main` advanced, `HEAD` and the working tree unchanged |
| G — step 4 with no standalone fetch | `pull --ff-only` fast-forwards successfully |

**Counterfactual that drove two snippet decisions:**

1. *Union vs. `git diff --name-only`* — with `doc.md` unstaged and `lock.json`
   staged, and an incoming commit touching `lock.json`: the narrow form yields an
   **empty** overlap (a false "nothing incoming touches your paths"); the union
   yields `ac84896 fix: upstream already fixed lock.json`.
2. *`xargs -0` vs. `-- $dirty`* — the unquoted form returns an empty overlap under
   **zsh**, which does not word-split unquoted parameters: the whole newline-joined
   list becomes one pathspec matching nothing. Same silent false negative, and it
   also breaks on any path containing a space. Verified side by side in bash and
   zsh.

Closes #98


rjwalters:
✅ **Approved**

Reviewed against issue #98's refined acceptance criteria and the curator's notes. Full suite run in the builder worktree: **PASS: 1193 FAIL: 0**, including `test-early-sync-switch.sh` 68/68 with that file **unmodified** (it does not appear in `git diff origin/main...HEAD`).

## Merge conflict resolved during review

`mergeStateStatus` was `DIRTY` — #121 (`#102` entry) landed a both-prepend under `## Unreleased` while this branch was open. Per the Judge role's simple-conflict path I resolved it directly:

- Rebased `feature/issue-98` onto `d9b6313`, kept **both** CHANGELOG entries newest-first (`#98` above `#102`), no other file conflicted.
- Post-rebase diff vs `origin/main` is byte-identical in scope to the pre-rebase one: `CHANGELOG.md +1`, `commands/repo/all.md 7 +-`, `commands/repo/reset.md 74 +-`.
- Re-ran the full suite after the rebase: **1193/0** again. Force-pushed as `fbed074`.
- CI on `fbed074`: `pnpm test (guard/handoff/loss-check suite)` — **success**. `mergeable_state`: **clean**.

## Verification

| Criterion | Result |
|---|---|
| `git fetch --all --prune` at top of step 1, before `git status --porcelain`, both paths | ✅ `reset.md:52-55`; read-only caveat stated explicitly and correctly (remote-tracking refs only) |
| No renumbering | ✅ diff touches nothing above line 46; `## Two halves` ("steps 1 and 4", "steps 2 and 3", "all four steps in order") textually unchanged; numbering still 1–5. `test-early-sync-switch.sh:466-475` still matches |
| ahead/behind vs `@{u}` with no-upstream guard | ✅ reproduced in a scratch bare-`origin` + two-clone setup: `ahead=0 behind=2` on a tracking branch; on a never-pushed branch `git rev-parse --abbrev-ref --symbolic-full-name '@{u}'` exits 128 and the whole block is skipped with rc 0 — verified in **both** bash and zsh |
| Overlap uses the full dirty set, not `git diff --name-only` | ✅ and the counterfactual holds: with `doc.md` unstaged, `lock.json` **staged**, and an incoming commit touching `lock.json`, the narrow form returns **empty** (silent false "no overlap") while the union returns the commit |
| Overlap steers to abort/inspect, not commit | ✅ `reset.md:110-114`, names the `--ff-only`-blocking consequence and offers `git show <sha>` |
| Step 4 keeps checkout + `pull --ff-only`, standalone fetch dropped, divergence report unchanged | ✅ `git log --oneline @{u}..HEAD` / `HEAD..@{u}` lines are untouched in the diff; the prune-moves-to-step-1 consequence is called out rather than left implicit |
| `all.md` stale step-4 citation corrected | ✅ lines 111-115; the pinned substring `run only the sync-and-switch half now` (asserted at `test-early-sync-switch.sh:415-416`) is preserved |

## On the `-- $dirty` → `xargs -0` deviation

I did not take this on faith — I reproduced it side by side against a real remote with `doc.md` unstaged + `lock.json` staged + `spaced name.txt` untracked, and an incoming `fix: upstream already fixed lock.json`:

```
######## BASH ########          ######## ZSH ########
A narrow  (git diff --name-only): (empty)      A narrow: (empty)
B union   (-- $dirty):  74ede23 fix: ...       B union (-- $dirty): (empty)   ← silent false negative
C builder (printf|tr|xargs -0): 74ede23 fix:.. C builder: 74ede23 fix: ...
```

The claim is **correct**: zsh does not word-split unquoted parameter expansions by default (`SH_WORD_SPLIT` off), so the newline-joined `$dirty` arrives as one pathspec matching nothing — the exact same silent "no incoming commit touches your dirty paths" false negative the union was introduced to eliminate. I also confirmed the space case: under bash, `spaced name.txt` splits into two bogus pathspecs with **no error** — silently wrong rather than loudly wrong. Since these snippets get executed under whatever shell the operator is running, deviating from the curator's `-- $dirty` was the right call, and the inline comment explains why at the point of use.

## Non-blocking observations (no changes requested)

1. **Offline / unreachable remote.** Step 1 now *leads* with `git fetch --all --prune`, so on a network failure the first command of the safety check fails and the prose doesn't say what to do (continue with a "could not refresh — upstream context unavailable" note, presumably). Not a regression — the same failure just surfaced at step 4 before, equally undocumented — but worth a sentence in a follow-up.
2. **Subdirectory relativity.** Verified that `git diff --name-only` emits repo-root-relative paths while `git ls-files --others` and git-log pathspecs are cwd-relative, so the union would be mismatched if step 1 ran from a subdirectory. `/repo:reset` is a repo-root ritual, so this is theoretical; `git -C "$(git rev-parse --show-toplevel)"` would pin it if it ever bites.
3. **Very large dirty sets** could make `xargs` split into multiple `git log` invocations, duplicating commit lines in `overlap`. Cosmetic only.

Documentation quality is high throughout — the CHANGELOG entry carries the motivating incident and the *why* behind each snippet decision, and the step-1 comments make the two non-obvious choices (union, `xargs -0`) self-defending against a future "simplification".

PR #123, opened 2026-08-04
feat(branches): add patch-id arm and per-branch "cleared by" tags to the permanent-loss check

### LINKED ISSUE #97: reset/branches: the permanent-loss check is reachability-based, so on a squash-merging repo it protects every merged branch and --prune can never delete one
## The permanent-loss check can't distinguish "landed" from "lost" on a squash-merging repo

`/repo:reset` (via `/repo:branches`) specifies the safety rule as:

> a branch with commits found nowhere else, or a worktree with uncommitted changes, is never
> removed automatically

On a repo that **squash-merges** its PRs, that describes *every merged branch*. Squashing
replays the branch as a single new commit on `main`, so the branch's original commits are
unreachable from `main` forever. A reachability test therefore reports "commits found
nowhere else — protect it" for work that is completely landed.

The result is that `--prune` can never delete a merged branch, and stale branches accumulate
without bound.

## Demonstrated on a live repo

Running `/repo:reset` on `rjwalters/loom` (squash-merge repo, 66 local branches):

```
$ git branch -d feature/issue-4728
error: the branch 'feature/issue-4728' is not fully merged
hint: If you are sure you want to delete it, run 'git branch -D feature/issue-4728'
```

…while PR **#4747 for that exact branch was merged**. Every one of the **6** branches
removed in that pass was in the same state — reachability-orphaned, content-landed:

| branch | reachability says | truth |
|---|---|---|
| `feature/issue-4728` | 1 unique commit | PR #4747 merged |
| `feature/issue-4854` | 1 unique commit | PR #4855 merged |
| `feature/issue-5076` | 1 unique commit | PR #5098 merged |
| `hygiene/repo-all-20260731` | 1 unique commit | PR #4839 merged |
| `feature/doctor/4428-crlf-fix` | 2 unique commits | no PR, but content verified in `main` |
| `work/issue-4705` | 1 unique commit | no PR, but content verified in `main` |

Note the last two: they had **no PR at all**, so even a merge-state check would protect
them — correctly, by its own rule. They only turned out to be safe after reading the code
and confirming the content had landed by another route (one file was byte-identical to
`main`; another's function signatures were the older ancestor of `main`'s).

## Why `-D` is not the answer either

The inverse failure is worse. `git branch -D` ignores the merge state entirely, so using it
to work around the refusal would delete genuinely unmerged work just as happily. **Neither
`-d` nor `-D` alone is safe on a squash-merging repo** — one refuses everything, the other
refuses nothing. The classification needs a signal git doesn't have.

## What worked instead

Substituted manually during that pass:

```bash
# merge state from the forge, not from reachability
gh api "repos/<owner>/<repo>/pulls?state=all&head=<owner>:<branch>" \
  --jq '[.[]|select(.merged_at!=null)]|length'
```

Any merged PR for the branch head ⇒ landed ⇒ SAFE, even though `-d` refuses. Where **no**
PR exists, fall back to content verification (`git cherry`, patch-id equivalence, or reading
the files) rather than assuming either way.

## Suggested direction

- Classify with **merge state**, not reachability: a branch whose head had a merged PR is
  SAFE. `git cherry main <branch>` (patch-id based) is the forge-independent equivalent and
  correctly reports squash-landed commits as `-`.
- Keep the protective default where the signal is absent: **no PR and no patch-id match ⇒
  never auto-remove**, which is exactly what preserved the two interesting branches above.
- Say plainly in the docs that `git branch -d` is unreliable here, so nobody "fixes" the
  refusal by reaching for `-D`.
- Worth noting the same assumption bites the Loom side independently — `worktree.sh remove`
  uses `git branch -d` and cannot delete squash-merged branches
  (rjwalters/loom#4889), and Loom's worktree reaper had the matching bug
  (rjwalters/loom#5177, since fixed). This is a general hazard for any tooling that infers
  "merged" from reachability.

## Acceptance criteria

- [ ] A branch whose PR is merged is classified SAFE even though `git branch -d` refuses it.
- [ ] A branch with no merged PR and no patch-id equivalence in the default branch is never
      auto-removed.
- [ ] Deletion uses a path that works on squash-merged branches, and the docs state why.
- [ ] Verified against a squash-merging repo — the failure does not reproduce where merge
      commits preserve reachability.
- [ ] The report distinguishes "landed (squash)" from "unique work" so an operator can see
      which rule applied.

## Context

Surfaced running `/repo:reset` on `rjwalters/loom` on 2026-08-04, which had accumulated 66
local branches partly because nothing could ever classify a merged one as deletable.

rjwalters:
## Summary

The permanent-loss check could clear a squash-landed branch only through tree
containment (`merge-tree` / `diff --quiet`) or a forge PR lookup, and reported
every cleared branch as a bare `SAFE TO DELETE`. That left two gaps behind the
reported failure:

- A **single-commit branch that squash-landed and whose files `<default>` has
  since edited** fails both tree probes, so without forge access it was kept as
  though the work were unique. That is exactly the shape of the six branches in
  the issue — one commit each, all landed.
- The report could not distinguish *landed as a squash commit* from *had no
  unique commits at all* from *exists nowhere else*, so an operator hitting a
  `git branch -d` refusal had nothing in-doc saying the refusal is expected, nor
  warning that git's own `-D` hint is the inverse hazard.

## Changes

**`commands/repo/branches.md`**
- 5b is now four ordered arms, each naming the report tag it produces:
  1. `merge-tree` tree containment → `landed (squash), content-verified (merge-tree)`
  2. `git diff --quiet` → `landed, identical tree`
  3. **new** `git cherry <default> <branch>` patch-id equivalence → `landed (squash), patch-id equivalent (git cherry)`, proven only when output is non-empty and every line is `-`. Its limit is documented: a multi-commit branch squashed into one matches no individual patch-id and correctly falls through.
  4. forge merge state → `landed (squash), merged PR #N (<date>)`
- Step 4 publishes the fixed tag vocabulary (table) and shows tagged report lines; a SAFE line with no tag is called out as a bug.
- 5a/5c emit tags too (`no unique commits`, `unverifiable: <reason>`); 5c clarified that unsupported `merge-tree` is arms 2–3's job, not an error.
- Deletion selects `-d` vs `-D` **from the tag**, under a callout: `-d` refuses every landed branch on a squash-merging repo, `-D` ignores merge state entirely, neither can tell landed from lost, and `-D` is permitted only for a `landed (...)` tag.
- Offline fallback no longer points at `git branch --merged`; arms 1–3 work with no network.
- New forge example uses the REST endpoint; repo#103 keeps the conversion of the existing GraphQL-backed `gh pr list` call (not duplicated here).

**`commands/repo/reset.md`** — carries the tag through step 3 and the final report, so the branch summary is never a bare `N deleted`; repeats the `-D` warning.

**`commands/repo/tests/test-branches-loss-check.sh`** — transcription grows the patch-id arm and now emits `verdict|tag`; new fixture 5 (single-commit squash, main advanced) is cleared by patch-id with zero forge access; fixture 4 becomes multi-commit so it still exercises the forge-only path; new tag assertions and doc-drift coverage for the vocabulary, the REST form and the `-d`/`-D` warning in both docs. 60 → 73 cases.

| A branch whose PR is merged is SAFE even though `git branch -d` refuses it | ✅ | Arm 4 (pre-existing) + new tag; suite: "merged PR clears a branch main has moved past", "its tag carries the PR number and merge date". Scratch repo: `-d` refused `feature/landed` while the check said SAFE |
| No merged PR and no patch-id equivalence ⇒ never auto-removed | ✅ | Suite: "unpushed work → UNSAFE", "unpushed work is NOT cleared by patch-id", "multi-commit squash is NOT cleared by patch-id", "forge lookup unavailable → KEEP" |
| Deletion uses a path that works on squash-merged branches, docs say why | ✅ | Tag-driven `-d`/`-D` block + callout; scratch repo: `git branch -D feature/landed` succeeded, `feature/unique` untouched |
| Verified against a squash-merging repo; failure does not reproduce where merge commits preserve reachability | ✅ | Scratch repo with `--no-ff` merge: 5a empty → tag `no unique commits` → plain `git branch -d` succeeded (exit 0), no squash label. Suite fixtures 1/4/5 cover the squash direction |
| Report distinguishes "landed (squash)" from "unique work" | ✅ | Tag vocabulary in branches.md step 4 + reset.md report; suite: "a landed tag never reads as unique work", "a unique-work tag never reads as landed", "every classified branch carries a report tag" |
| Cross-reference repo#103 rather than duplicating its REST scope | ✅ | branches.md 5b names repo#103 and keeps the existing `gh pr list` call unchanged; only the new number/date call is REST |

## Test Plan

- `./commands/repo/tests/test-branches-loss-check.sh` — **73 pass, 0 fail, 0 skipped** (was 60).
- `pnpm test` (`./hooks/repo/tests/run.sh`) — **1229 pass, 0 fail** across all 10 suites.
- Manual, squash-merging scratch repo: `-d` refuses the squash-landed branch (exit 1, git hints `-D`); the check tags it `landed (squash), patch-id equivalent (git cherry)`; `-D` deletes it; the no-PR/no-patch-id branch is tagged `unique work: 1 commits found nowhere else` and survives.
- Manual, merge-commit scratch repo: 5a is empty, tag is `no unique commits`, `git branch -d` succeeds directly — the squash label does not appear.

Closes #97


rjwalters:
✅ **Approved!**

Evaluated per user request. The fix is correct, the protective default is preserved, and I reproduced both the failure and the fix independently before approving.

## Independent verification (scratch squash-merging repo)

Built a repo with a single-commit branch squash-landed into `main`, then advanced `main` past the same file — the exact shape of the six branches in #97:

```
git branch --merged main          -> main            (omits the landed branch)
git branch -d feature/landed      -> exit 1, "not fully merged", hint: use -D
arm 1  merge-tree --write-tree    -> NOT equal (add/add conflict)
arm 2  git diff --quiet           -> exit 1
arm 3  git cherry main <branch>   -> "- d520903"     <- clears it, no forge access
       git cherry main feature/unique -> "+ e013bdd" <- unique work not cleared
git branch -D feature/landed      -> deleted; feature/unique untouched
```

Control (merge-commit repo, `--no-ff`): 5a is empty, so the tag is `no unique commits` and plain `git branch -d` succeeds — the squash arms never fire. AC #4 holds.

I also ran the REST snippet the docs now prescribe against this repo, since a doc that ships an untested command is just a plausible-looking string:

```
$ gh api "repos/{owner}/{repo}/pulls?state=all&head={owner}:feature/issue-98" \
    --jq '[.[] | select(.merged_at != null)] | "\(length) \(.[0].number // "") \((.[0].merged_at // "")[0:10])"'
1 122 2026-08-04
$ ... head={owner}:does-not-exist
0
```

`gh` does substitute `{owner}`/`{repo}` inside the query string, the jq survives the empty-array case (`.[0]` on `[]` is `null`, so both `//` defaults fire), and the output is exactly the `<count> <number> <date>` the tag needs. Good call using `<branch>` rather than `{branch}` in the placeholder — `{branch}` is a real `gh` placeholder and would have silently substituted the *current* branch.

## Acceptance criteria

| Criterion | Verdict |
|---|---|
| Merged PR ⇒ SAFE despite `git branch -d` refusing | ✅ arm 4 + tag `landed (squash), merged PR #N (<date>)`; refusal reproduced above |
| No merged PR **and** no patch-id equivalence ⇒ never auto-removed | ✅ falls through all four arms to `unique work: N commits found nowhere else`; suite pins it from three directions, and `git cherry` reports `+` on genuinely unique work |
| Deletion path works on squash-merged branches, docs say why `-d` and `-D` are each individually wrong | ✅ flag chosen **from the tag**, under a callout naming both failure directions (`-d` refuses everything, `-D` refuses nothing, git's own hint steers into the second) |
| Report distinguishes "landed (squash)" from "unique work" | ✅ fixed 7-tag vocabulary published in branches.md step 4, carried into reset.md's summary; a bare `SAFE TO DELETE` is explicitly called a bug |
| `git cherry`'s multi-commit-squash limitation documented | ✅ stated inline ("patch-ids are per commit … matches none of them individually") **and** enforced — fixture 4 was deliberately made multi-commit so it still falls through to the forge arm, which is the honest way to pin a documented limit |

## Test execution

- `commands/repo/tests/test-branches-loss-check.sh` — **73 pass, 0 fail, 0 skipped** (was 60).
- `./hooks/repo/tests/run.sh` — **1229 pass, 0 fail** across all 10 suites. Re-run after the rebase and changelog commit below; unchanged.
- CI on the final head `e00fda8`: `pnpm test (guard/handoff/loss-check suite)` ✅, `mergeable_state: clean`.

## Changes made during review

1. **Rebased onto `origin/main`** (picking up #121 and #122). No conflicts. Verified the combined `commands/repo/reset.md` keeps **both** #122's step-1 `git fetch --all --prune` and this PR's tag-carrying step 3 / report — they touch disjoint sections.
2. **Added the missing `## Unreleased` CHANGELOG bullet** (`e00fda8`). The section carries one curated bullet per user-facing change and this PR changed the behaviour of two shipped commands without one. Placed at the top of the list per the newest-first convention — which is also why the rebase had to come first: a top-of-list insertion on the pre-#122 base would have conflicted with #121/#122's insertions at the same line on merge. Reword freely if the emphasis is off; no behaviour is affected either way.

## Notes (non-blocking, no action requested)

- Arm 3's ordering ahead of the forge arm is the right call — it is free, offline, and strictly more conservative than a network call that can be rate-limited into a `KEEP`.
- The `every line is '-'` requirement is what makes arm 3 safe: one `+` (a partially-landed branch) falls through rather than clearing, so patch-id can never be the sole reason a mixed branch is deleted.
- A patch that landed in `<default>` and was later reverted would still read `-`. That is not a loss: the commit remains reachable from `<default>`, so the work is recoverable. Worth knowing, not worth guarding.
- `head={owner}:<branch>` only matches same-owner heads, so a fork-origin PR wouldn't be found by arm 4. Irrelevant to this repo's same-repo branch workflow, and the failure direction is the safe one (falls through to `unique work` / KEEP).

PR #124, opened 2026-08-04
fix(tidy): prune .git by -name so nested .git directories are skipped

rjwalters:
## Summary

Both step-1 inventory `find` walks in `commands/repo/tidy.md` pruned `.git` with `-path './.git'`, which matches **only the root**. A nested `.git` *directory* — a vendored clone under `vendor/`, a `git clone`'d fixture in a test tree, an old checkout parked inside the working tree — was still fully descended into. That is the one tree pruning most exists to skip, since a `.git` tree is typically the largest directory in a checkout.

Not a regression (the pre-#85 `-not -path './.git/*'` had the same blind spot), but it was the sole exception to the file's own **Match by `-name`, not `-path`** rule, so the rule's carve-out goes away with the gap.

## Changes

- `commands/repo/tidy.md`: `-path './.git'` -> `-name .git` in **both** step-1 `find` invocations (empty-directory walk and >10 MB-file walk). The two prune lists remain textually identical to each other.
- `commands/repo/tidy.md`: the "Match by `-name`, not `-path`" bullet drops the `(except './.git', which is unambiguously at the root)` parenthetical and now cites a vendored `vendor/foo/.git/` alongside `packages/foo/node_modules/` as a nested case that is covered.
- `CHANGELOG.md`: Unreleased entry per repo convention.

Documentation-only; no script, hook, or test behavior changes.

## Acceptance Criteria Verification

| Criterion | Status | Verification |
|-----------|--------|--------------|
| Both step-1 `find` invocations prune `.git` by `-name`, and the two prune lists remain identical to each other (the invariant from PR #100) | ✅ | Both now read `find . \( -name .git -o -name node_modules -o -name target \` / `-o -name dist -o -name .venv \) -prune \`; byte-compared the two two-line prune blocks — IDENTICAL |
| The "Match by `-name`, not `-path`" bullet no longer carries a `./.git` exception | ✅ | Parenthetical removed; `grep -n "except" commands/repo/tidy.md` finds no `./.git` carve-out |
| Verified on a fixture with a nested `.git` directory holding a >10 MB object: no longer reported, and reporting outside the pruned trees is unchanged | ✅ | See Test Plan fixture 1 below |

## Test Plan

[...]

Issue #104, opened 2026-08-04
tidy inventory: prune .git by -name so nested .git directories are skipped too


rjwalters:
## Context

Found while reviewing PR #100 (issue #85), which rewrote `/repo:tidy`'s step-1 inventory `find` walks to use `-prune`. That change is correct and lands the reported win. This is a small residual gap in the prune list itself.

## Problem

Both walks prune `.git` with `-path './.git'`, which matches **only the root** `.git`:

```bash
find . \( -path './.git' -o -name node_modules -o -name target \
          -o -name dist -o -name .venv \) -prune \
     -o -type f -size +10M -print
```

A nested `.git` **directory** — a vendored clone under `vendor/`, a `git clone`'d fixture in a test tree, an old checkout parked inside the working tree — is still fully descended into. Verified on a synthetic fixture: with `sub/.git/objects/pack` at 12 MB, the new form still walks `sub/.git/` and reports the pack file:

```
=== NEW large-file form ===
./src/bigfile.bin
./sub/.git/objects/pack      <-- nested .git still traversed
```

This is not a regression (the old `-not -path './.git/*'` had the same blind spot), but it is the one remaining case where the prune list does not do what its own documented rule says.

## Why it's worth closing

`tidy.md`'s new prose (added by PR #100) states the rule as:

> **Match by `-name`, not `-path`** (except `./.git`, which is unambiguously at the root). `-name` prunes at any depth...

The carve-out is the only exception in the list, and it isn't load-bearing: a `.git` tree is typically the single largest directory in any checkout, so it's exactly the tree pruning exists to skip. Making it uniform removes the exception from the doc as well as the gap from the command.

## Suggested change

In `commands/repo/tidy.md`, change `-path './.git'` to `-name .git` in both step-1 `find` invocations, and drop the parenthetical exception from the "Match by `-name`, not `-path`" bullet.

No downside identified:
- `-name .git` still prunes the root `.git` (superset of current behavior).
- For git worktrees and submodules, `.git` is a *file*, not a directory. `-prune` on a non-directory is a no-op that evaluates true, so the file is simply not printed — and it would never have been printed anyway, since the print terms are `-type d -empty` and `-type f -size +10M` and a `.git` pointer file is ~50 bytes.

## Acceptance criteria

- [ ] Both step-1 `find` invocations in `commands/repo/tidy.md` prune `.git` by `-name`, and the two prune lists remain identical to each other (the invariant PR #100 documented).
[...]

PR #125, opened 2026-08-04
docs(gitignore): require a verified real directory before deduping `X` and `X/`

### LINKED ISSUE #110: gitignore/audit: don't propose deduping `X` and `X/` — trailing-slash patterns don't match symlinks (caused a live regression)
## Context
During a /repo:all run in rjwalters/lean-genius (2026-08-04), the audit stage flagged `.lake` + `.lake/` in .gitignore as "redundant/duplicated rules" and the hygiene PR deduped them to `.lake/`. That unignored a live path: the repo root has a `.lake` **symlink** (→ proofs/.lake), and trailing-slash gitignore patterns match directories only — never symlinks. Same trap, same session, other direction: a `loom-tools/` entry in .git/info/exclude failed to exclude a `loom-tools` symlink until the slash was dropped. The corrective commit rode rjwalters/lean-genius#43683.

The audit/gitignore command guidance should warn that `X` and `X/` are only duplicates when X is guaranteed to be a directory — in repos using symlinks (external volumes, vendored trees), the bare rule is load-bearing.

- [ ] /repo:gitignore and /repo:audit docs: the duplicate-rule check treats `X` + `X/` as NOT redundant unless X is verified to be a real directory (e.g. `[ -d X ] && [ ! -L X ]`)
- [ ] Suggested-fix output for such pairs keeps the bare rule

---

### Root cause

`git check-ignore` semantics: a pattern with a trailing `/` (e.g. `X/`) only
ever matches a **directory** — never a symlink, even a symlink that resolves
to a directory (`man gitignore`: *"If there is a separator at the end of the
pattern then the pattern will only match directories"*). So `X` (bare) and
`X/` (slash) in the same `.gitignore` are **not interchangeable**:

| Path type            | Matched by `X` | Matched by `X/` |
|-----------------------|:---:|:---:|
| Real directory         | ✅ | ✅ |
| Symlink (to anything)  | ✅ | ❌ |
| Regular file           | ✅ | ❌ (never matches) |

Collapsing `X` + `X/` down to just `X/` silently un-ignores any symlink at
that path. That's exactly what happened with `.lake` (a symlink to
`proofs/.lake`) in the lean-genius regression this issue documents.

### Where this actually lives (verified against `origin/main`, current as of
this curation — repo has since had 6 more commits land, but neither of these
two files changed)

There is **no separate "duplicate-rule" implementation to patch** — the audit
flow is: `/repo:all` Stage 1 → `[[audit]]` → its "Gitignore Issues" check
defers to `[[gitignore]]` (`commands/repo/audit.md` literally says "(see
[[gitignore]])" and doesn't re-derive the hygiene rules). `/repo:gitignore`'s
own "Gitignore Hygiene" section is the actual source of the "redundant rules"
judgment call that an LLM agent makes when scanning a `.gitignore` file, and
its "Interaction" section is what authorizes auto-applying "clear-cut" fixes
without asking. So the fix belongs almost entirely in
`commands/repo/gitignore.md`, with a thin cross-reference note in
`commands/repo/audit.md` for defense-in-depth (a reader skimming just the
audit doc should still see the trap called out, not just "see gitignore").

**Note on file layout**: `commands/repo/*.md` is the tracked source (this is
the source repo for the `/repo:*` skill package — see repo-skills-project
memory). `.claude/commands/repo/*.md` is a local installed copy and is
git-ignored (`.gitignore:7` → `.claude/`) — confirmed via
`git check-ignore -v`. Edit `commands/repo/gitignore.md` and
`commands/repo/audit.md`, not the `.claude/` copies.

**`commands/repo/gitignore.md`** — `### 3. Gitignore Hygiene` (lines 54-58 on
`origin/main`):
```markdown
### 3. Gitignore Hygiene
- Redundant rules (already covered by a parent `.gitignore`)
- Rules that match zero files (stale after cleanup)
- Scattered `.gitignore` files that could be consolidated
```
This bullet list is where the agent decides "these two rules are redundant."
It currently only calls out parent/child `.gitignore` redundancy — it says
nothing about same-file `X` + `X/` pairs, which is the gap that let `.lake` +
`.lake/` get flagged. Add a caveat directly under this list (suggested text
below).

**`commands/repo/audit.md`** — `### 4. Gitignore Issues (see [[gitignore]])`
(lines 42-45 on `origin/main`):
```markdown
### 4. Gitignore Issues (see [[gitignore]])
- Files that are ignored but probably shouldn't be
- Build artifacts that aren't ignored but should be
- Redundant or stale gitignore rules
```
Since this section defers to `[[gitignore]]`, fixing `gitignore.md` alone
would already flow through here at runtime — but per the acceptance criteria
("/repo:gitignore **and** /repo:audit docs"), add a short explicit
parenthetical on the "Redundant or stale gitignore rules" bullet so a reader
of just this doc still sees the warning, without needing to duplicate the
full explanation.

### Suggested text to embed (gitignore.md, under the Hygiene list)

```markdown
**Do not flag `X` and `X/` as duplicates without verification.** A trailing
slash restricts a gitignore pattern to directories — it never matches a
symlink, even one that points at a directory (`man gitignore`: "If there is
a separator at the end of the pattern then the pattern will only match
directories"). `X` and `X/` in the same file are true duplicates only when
`X` is guaranteed to never be a symlink:

    [ -d "X" ] && [ ! -L "X" ]   # true only for a real, non-symlink directory

If that check fails, or `X` could plausibly be a symlink in this repo
(vendored trees, external volumes, generated build outputs), the pair is
**not redundant** — keep both rules in the suggested-fix output. Dropping the
bare rule un-ignores any symlink at that path, since `X/` alone won't
re-cover it. This caused a live regression: `.lake` + `.lake/` were deduped
to `.lake/`, unignoring a `.lake` symlink (rjwalters/lean-genius#43683).
```

### Suggested text to embed (audit.md, on the existing bullet)

```markdown
- Redundant or stale gitignore rules — **not** `X` + `X/` pairs unless `X` is
  verified to be a real, non-symlink directory (`[ -d X ] && [ ! -L X ]`);
  see [[gitignore]] for the full check. Wrongly deduping these can silently
  un-ignore a symlink.
```

### Implementation notes
- This is a docs-only change (no scripts implement gitignore dedup logic —
  the "redundant rule" judgment is made by the LLM agent reading these
  command docs at run time), so there's no code path to unit-test.
- Keep the embedded shell check (`[ -d X ] && [ ! -L X ]`) verbatim or
  equivalent — it's the concrete, mechanical test that distinguishes a
  superficial fix ("be careful with symlinks") from a real one (an agent
  following the doc should actually run this check, or reason equivalently,
  before proposing a merge).
- Do not remove or soften the existing "Redundant rules (already covered by
  a parent `.gitignore`)" bullet in `gitignore.md` — that's a distinct,
  still-valid check (parent/child redundancy). This issue only scopes the
  new same-file `X`/`X/` caveat.

### Verification performed during curation
- Confirmed via `git diff HEAD origin/main -- commands/repo/gitignore.md commands/repo/audit.md`
  that the local working tree matches `origin/main` for both files (no
  uncommitted drift), and both paths exist in `origin/main`'s tree
  (`git ls-tree -r origin/main --name-only`).
- Searched `commands/repo/*.md` for other "redundant"/"dedup" mentions
  (`tidy.md`, `release.md`, `followups.md`) — all are unrelated (issue-filing
  dedup, an unrelated `--caches` flag note). No other doc needs this caveat.
- Searched `commands/repo/tests/*.sh` and `.loom/scripts/tests/*.sh` for any
  test pinning the wording of the Hygiene/Gitignore-Issues bullets — none
  found. `.loom/scripts/tests/test-gitignore-guard.sh` is an unrelated Loom
  installer guard (checks installed `.loom/scripts/lib/*.sh` files aren't
  gitignored), not this skill's audit logic.
- Ran `check-duplicate.sh --include-merged-prs` — it flagged several closed,
  topically-unrelated issues at "100% similarity" alongside #110 itself,
  which indicates a low-precision/likely-broken similarity heuristic rather
  than a genuine duplicate; manual review of titles/bodies found no real
  duplicate.

## Affected Files
- `commands/repo/gitignore.md` — add the "Do not flag `X` and `X/` as
  duplicates without verification" caveat to `### 3. Gitignore Hygiene`
  (currently lines 54-58 on `origin/main`)
- `commands/repo/audit.md` — extend the "Redundant or stale gitignore rules"
  bullet under `### 4. Gitignore Issues` (currently lines 42-45 on
  `origin/main`) with the same-file `X`/`X/` caveat

## Test Plan
- [ ] Manual verification: re-read both edited files and confirm the caveat
  text appears verbatim (or equivalent) in each, including the
  `[ -d X ] && [ ! -L X ]` check and the explicit "keep the bare rule /
  don't collapse to `X/`" instruction
- [ ] Manual verification: confirm the existing "Redundant rules (already
  covered by a parent `.gitignore`)" bullet in `gitignore.md` is left intact
  — this issue adds a new caveat, it doesn't replace the parent/child check
- [ ] Automated tests: N/A — this is a prose-only docs change with no
  existing test suite pinning the wording (verified during curation); no
  script implements the dedup judgment being corrected
- [ ] Edge case: the added text should distinguish "verified real,
  non-symlink directory" from "looks like a directory right now" — a fix
  that only adds a vague "be careful with symlinks" warning without the
  concrete `[ -d ] && [ ! -L ]` check and the "keep the bare rule" directive
  does not satisfy the acceptance criteria

rjwalters:
## Summary

A trailing slash restricts a gitignore pattern to **directories** — it never matches a symlink, even one that resolves to a directory (`man gitignore`: "If there is a separator at the end of the pattern then the pattern will only match directories"). So in an `X` + `X/` pair the *bare* rule is load-bearing, not noise.

`/repo:gitignore`'s hygiene check only warned about parent/child `.gitignore` redundancy and said nothing about same-file pairs. An audit run in `rjwalters/lean-genius` flagged `.lake` + `.lake/` as duplicated and the hygiene PR deduped them to `.lake/`, un-ignoring the repo-root `.lake` **symlink** (→ `proofs/.lake`); the corrective commit rode rjwalters/lean-genius#43683. The same trap fired in the other direction in that session, where a `loom-tools/` entry in `.git/info/exclude` failed to exclude a `loom-tools` symlink until the slash was dropped.

## Changes

- `commands/repo/gitignore.md` — new caveat under `### 3. Gitignore Hygiene` (the section where the agent actually makes the "these two are redundant" call): the `X`-vs-`X/` match matrix across real directory / symlink / regular file, the mechanical pre-merge check `[ -d "X" ] && [ ! -L "X" ]`, an explicit note that `[ -d "X" ]` alone is insufficient because it follows symlinks, and the directive to **keep both rules in the suggested-fix output** when the check fails, `X` doesn't exist to test, or `X` could plausibly become a symlink (vendored trees, external volumes, relocated build outputs).
- `commands/repo/audit.md` — the "Redundant or stale gitignore rules" bullet under `### 4. Gitignore Issues` carries the short form, so a reader of just this doc sees the trap instead of only "see [[gitignore]]".

Edits are on the tracked sources in `commands/repo/`, not the gitignored installed copies under `.claude/commands/repo/`.

| `/repo:gitignore` docs: `X` + `X/` not redundant unless X is a verified real directory (`[ -d X ] && [ ! -L X ]`) | ✅ | Caveat added under `### 3. Gitignore Hygiene` with the check verbatim, plus the `[ -d ]`-follows-symlinks note that makes it non-vacuous |
| `/repo:audit` docs: same | ✅ | Bullet under `### 4. Gitignore Issues` names the check inline and defers to `[[gitignore]]` for the full explanation |
| Suggested-fix output for such pairs keeps the bare rule | ✅ | "keep both rules in the suggested-fix output ... report the pair as intentional rather than collapsing it" |
| Distinguishes "verified real, non-symlink directory" from "looks like a directory right now" | ✅ | Explicit: `[ -d "X" ]` alone follows symlinks, so a symlink-to-directory passes it; the `! -L` arm is what does the work |
| Existing parent/child redundancy bullet left intact | ✅ | `git diff` shows pure insertion after the bullet list — no lines removed or reworded |

## Test Plan

Empirically verified the semantics on a throwaway git fixture (real directory, symlink, regular file; rules with and without the trailing slash):

- With `link/` in `.gitignore`, `git check-ignore -v link` reports **NOT IGNORED** for a symlink named `link`, and `git status --porcelain` shows `?? link`.
- With the bare `link` rule, `git check-ignore -v link` matches (`.gitignore:1:link	link`).
- `plainfile/` likewise fails to match a regular file; `realdir/` and `realdir` both match a real directory.

This reproduces the regression shape exactly and confirms the matrix embedded in the doc.

Full suite green — `./hooks/repo/tests/run.sh`: **PASS: 1229 FAIL: 0** (no suite pins this wording; verified during curation and unchanged here).

Closes #110


rjwalters:
✅ **Approved!**

Docs-only change, and the semantics it documents are correct — I reproduced them independently rather than taking the PR's word for it.

## Independent verification of the claim

Throwaway `git init` fixture (git 2.55.0) with `realdir/` (real), `link -> target/` (symlink to a directory), and `plainfile` (regular file):

| Rule set | `git check-ignore -v` matches | `git status --porcelain` |
|---|---|---|
| `realdir/`, `link/`, `plainfile/` | only `realdir` | `?? link`, `?? plainfile` — **not ignored** |
| `realdir`, `link`, `plainfile` | all three | none untracked |

That is exactly the matrix embedded in `commands/repo/gitignore.md`, including the "Regular file → `X/` never matches" row. The regression shape from #110 (`.lake` + `.lake/` collapsed to `.lake/`, un-ignoring the `.lake` symlink) reproduces verbatim.

And the load-bearing detail the builder flagged:

```
[ -d "link" ]                    => TRUE   (follows the symlink)
[ -d "link" ] && [ ! -L "link" ] => FALSE  (correctly rejects it)
```

So the `! -L` arm is doing all the work — `[ -d "X" ]` alone would have passed the exact path that caused the regression. The doc calls this out explicitly ("`[ -d \"X\" ]` alone is **not** sufficient: it follows symlinks"), which is what turns this from a vague "be careful with symlinks" note into a mechanical check an agent can actually run. That was an explicit acceptance criterion on #110 and it's satisfied non-vacuously.

## Acceptance criteria

- ✅ `commands/repo/gitignore.md` — caveat lands at the end of `### 3. Gitignore Hygiene`, i.e. the section where the agent actually makes the "these two are redundant" call, not in a distant reference section. Contains the match matrix, the verbatim `[ -d "X" ] && [ ! -L "X" ]` check, the follows-symlinks note, and the keep-both-rules directive (including the three "can't test / could plausibly become a symlink" escape hatches: nonexistent path, vendored trees, external volumes, relocated build outputs).
- ✅ `commands/repo/audit.md` — short form on the "Redundant or stale gitignore rules" bullet under `### 4. Gitignore Issues`, naming the check inline and deferring to `[[gitignore]]` for the explanation. A reader of only this doc now sees the trap.
- ✅ Suggested-fix output keeps the bare rule — "keep both rules in the suggested-fix output and report the pair as intentional rather than collapsing it."
- ✅ Parent/child redundancy bullet untouched — `git diff` is a pure insertion (34 insertions, 1 deletion, the deletion being the reflowed audit bullet). No line of the existing hygiene list is removed or reworded.
- ✅ Tracked sources edited. Diff touches `commands/repo/audit.md` and `commands/repo/gitignore.md` only — no `.claude/commands/repo/` copies (those are gitignored installed copies).

## Merge conflict resolved during review

PR #124 landed a `CHANGELOG.md` Unreleased entry while this PR was in the queue, producing a `DIRTY` both-prepend conflict at the same anchor. Per the Judge role's simple-conflict path I rebased onto `origin/main` (`a6f8794`), kept both entries newest-first (#110 above #104, matching this repo's ordering), re-ran the suite, and force-pushed. `mergeStateStatus` is now `CLEAN`.

## Test Execution

Full suite in the rebased worktree — `./hooks/repo/tests/run.sh`:

```
PASS: 1229   FAIL: 0
```

Per-suite breakdown all green (guard-destructive 544, shell-wrapper 120, session-start-handoff 99, remote 87, verify-fix-persistence 79, branches-loss-check 73, early-sync-switch 68, install-claude-md-markers 59, inline smoke 55, install-sidecar-untracking 45). Matches the count claimed in the PR body, confirming the rebase introduced no drift. No suite pins this wording, as the curator noted — consistent with the docs-only scope.

CI: `pnpm test (guard/handoff/loss-check suite)` — pass (1m17s) on the rebased head.

## Notes

No blocking or non-blocking concerns. The one judgment call worth recording: the doc deliberately biases toward false negatives (declaring a pair non-redundant when unsure), which is the right asymmetry here — a missed dedup costs one redundant line, a wrong dedup silently un-ignores a path and can leak a build tree into a commit.

PR #127, opened 2026-08-04
fix(release): propagate uv lock failures and bump single-quoted pyproject versions

### LINKED ISSUE #114: release.md: PEP 621 pyproject follow-ups (uv-lock error swallowing, single-quote apply mismatch, drift-gate docs, tautological guard)
## Context

Non-blocking follow-ups surfaced by the Judge during the `/loom:sweep all` review of PR #105 (issue #83, "detect PEP 621 `[project].version` ahead of the npm fallback"). The PR was approved and merged; these four items were posted to the PR review comment and would otherwise be lost. All are in `commands/repo/release.md`.

1. **`|| :` on the `uv lock` gate swallows a genuine failure.** The Phase 5 `pyproject)` apply path uses `... && uv lock || :` so the gate's exit status doesn't leak when no lock file exists — but it also masks a real `uv lock` failure. Phase 5 then stages the (un-regenerated) `uv.lock`. A stubbed `uv` exiting non-zero still yields branch exit 0. Suggested: `if [ -f uv.lock ] && command -v uv >/dev/null 2>&1; then uv lock; fi` so a real failure surfaces.

2. **Detection accepts single-quoted `version`, apply rejects it.** Phase 2 detection matches `version = '1.0.0'` (legal TOML) as `pyproject`, but the Phase 5 apply regex is `"[^"]*"` (double-quote only), so it finds nothing and exits 1. Safe failure direction, but a real gap: a repo that uses single-quoted TOML version strings is detected then fails at apply.

3. **Drift-gate "within a single tool" bullet omits pyproject.** The drift-gate / extension-points prose (around line ~235) still names only `cargo`/`poetry`; `pyproject` (with the `uv.lock` mirror nuance) belongs there now that it is a first-class detection branch.

4. **Line ~199 reconciliation guard is now tautological.** The cross-source reconciliation guard keeps `&& grep -q '"version"' package.json`, which is redundant given the Phase 2 `npm` branch was tightened to require the `"version"` field (merged in the same PR). The prose above it was updated to two conditions while the code kept three.

- [ ] `uv lock` failures surface instead of being swallowed by `|| :`, while a missing lock file still no-ops cleanly.
- [ ] Single-quoted TOML `version = '...'` either applies correctly or is not detected as `pyproject` in the first place (detection and apply agree).
- [ ] The drift-gate / extension-points documentation names `pyproject` alongside cargo/poetry, including the `uv.lock` mirror nuance.
- [ ] The tautological `&& grep -q '"version"'` in the reconciliation guard is removed (or its necessity documented).

rjwalters:
## Summary

Four PEP 621 `pyproject` follow-ups the Judge flagged during the `/loom:sweep all` review of PR #105 (issue #83). All four are in `commands/repo/release.md`; two are behavior fixes to embedded snippets, two are documentation corrections.

## Changes

- **Phase 5 `pyproject)` arm — `uv lock` failures now surface.** Replaced `&& { [ -f uv.lock ] && command -v uv >/dev/null 2>&1 && uv lock || : ; }` with `&& if [ -f uv.lock ] && command -v uv >/dev/null 2>&1; then uv lock; fi`. `A && B && C || :` absorbs a non-zero exit from *all three* chained commands, not just the two existence/PATH probes, so a real `uv lock` failure (resolver conflict, corrupt lockfile, offline network) exited 0 and the branch went on to stage the un-regenerated `uv.lock` into the version commit. The `if` form keeps the probes optional (false condition → block skipped, exit 0) while letting `uv lock`'s own status propagate.
- **Phase 5 rewrite regex — quote-agnostic (curator's recommended option (a)).** The pattern was `\"[^\"]*\"` (double-quote only) while Phase 2 detection accepts any quote style, so a repo using legal single-quoted TOML detected as `pyproject` then died at apply on the zero-substitution guard. Now captures the opening quote and backreferences it: `([\"\x27])[^\"\x27\n]*\2`, replaced with `m.group(1)+m.group(2)+sys.argv[1]+m.group(2)`. `\x27` rather than a literal `'` because the Python source sits inside a single-quoted `python3 -c '…'` shell argument.
- **Drift gate (`### Drift gate (multi-source)`)** — the *Within a single tool* bullet now names `pyproject` alongside `cargo` inheritance and `poetry` in the single-location no-op list, with the caveat that `uv.lock`'s embedded copy of the project version is a **derived cache** regenerated by the apply step (Phase 5), not an independent pre-bump drift target.
- **Cross-source reconciliation guard** — dropped the tautological `&& grep -q '"version"' package.json`; PR #105 tightened the `npm` detection branch to require that same grep before assigning `VERSION_TOOL="npm"`, so the third condition could not be false by the time the guard runs. Now two conditions, matching the prose directly above it.
- **Prose kept in sync**: the Phase 5 load-bearing bullet list goes from three to four items — a new bullet documents the captured-quote behavior, and the `uv.lock` bullet drops the `|| :` framing and adds an explicit "do not collapse this back into an `A && B && uv lock || :` chain" note explaining why.

| `uv lock` failures surface instead of being swallowed by `\|\| :`, while a missing lock file still no-ops cleanly | ✅ | Scratch repo + `uv` stub exiting 1: old form `rc=0`, new form `rc=1`. No `uv.lock` present → `rc=0`; `uv.lock` present but `uv` off `PATH` → `rc=0`. Also exercised inside a real `case` arm to confirm it parses. |
| Single-quoted TOML `version = '...'` applies correctly (detection and apply agree) | ✅ | Detection snippet (lines 209-211) on a single-quoted fixture → `pyproject`; the verbatim Phase 5 arm extracted from the file rewrites it to `version = '7.7.7'`, `rc=0`. Both halves now accept. |
| Drift-gate documentation names `pyproject` alongside cargo/poetry, including the `uv.lock` mirror nuance | ✅ | Read-through of the amended bullet — see diff. |
| The tautological `&& grep -q '"version"'` in the reconciliation guard is removed | ✅ | Guard is now `if [ "$VERSION_TOOL" = "npm" ] && [ -f VERSION ]; then` — two conditions. |
| No regression on double-quoted apply | ✅ | Double-quoted fixture rewrites exactly as before; a `[tool.ruff] version` outside `[project]` is left untouched; a list-valued `classifiers` key preceding `version` still matches (line-anchored scoping preserved); a `pyproject.toml` with no `version` still exits 1 with `ERROR: no [project].version found`. |

## Test Plan

Every check below was run against the snippet **extracted verbatim from the edited file** (not a hand-retyped copy), on throwaway fixtures:

- `uv lock` failure propagation, no-lock-file no-op, no-`uv`-on-`PATH` no-op, and `case`-arm parse — all as tabled above.
- Quote handling: single-quoted, double-quoted, `classifiers`-before-`version`, out-of-table `version`, and missing-`version` fatal path.
- Read-through of both prose edits.
- Full suite: `./hooks/repo/tests/run.sh` → **PASS: 1236   FAIL: 0** (unchanged; no test pins `release.md` prose — confirmed by the curator and by `git grep -l "release.md"` over `commands/repo/tests/` and `hooks/repo/tests/`).

Closes #114


rjwalters:
✅ **Approved!**

All four sub-items of #114 verified by executing the snippets **extracted verbatim from `commands/repo/release.md`** (line 433 case arm and lines 209-211 detection) against throwaway fixtures — not hand-retyped copies.

## Verification

### (1) `uv lock` failure propagation — `commands/repo/release.md:433`

| Scenario | Result |
|---|---|
| `uv.lock` present, `uv` stub exits 1 | **rc=1** — failure propagates, `pyproject.toml` already rewritten but the apply aborts |
| Old `A && B && uv lock \|\| :` form, same failing stub | rc=0 — confirms the bug this PR fixes |
| No `uv.lock` present | rc=0, version rewritten, no `uv` invoked |
| `uv.lock` present, `uv` off `PATH` | rc=0, version rewritten, no `uv` invoked |
| Arm pasted into a real `case`/`esac` | parses, rc=0, version rewritten |

The `if` form is the right shape: the condition's false status is not the statement's status, so only `uv lock`'s own exit escapes.

### (2) Quote-agnostic regex with backreference

The `\x27` escaping does work inside the single-quoted `python3 -c '…'` context — the Python source is an `r"…"` raw string, so `\x27` reaches `re` as an escape and resolves to `'` without ever appearing literally in the shell word.

| Fixture | Result |
|---|---|
| `version = '1.0.0'` (single-quoted) | → `version = '7.7.7'`, rc=0; quote style preserved |
| `version = "1.0.0"` (double-quoted) | → `version = "7.7.7"`, rc=0; no regression |
| `[tool.ruff] version` outside `[project]` | untouched in both fixtures |
| `classifiers = [ … ]` list before `version` | still matched (line-anchored scoping intact), rc=0 |
| No `[project].version` | rc=1, `ERROR: no [project].version found in pyproject.toml` |
| Detection snippet (209-211) on single-quoted fixture | `VERSION_TOOL=pyproject` — detection and apply now agree, which was the actual defect |

Group numbering is correct: `m.group(1)` is the `[project]…version = ` prefix, `m.group(2)` the captured quote, and `\2` backreferences it.

### (3) Drift-gate docs — `commands/repo/release.md:313-319`

The *Within a single tool* bullet now names `pyproject` alongside `cargo` inheritance and `poetry`, with the `uv.lock`-is-a-derived-cache caveat pointing at Phase 5 as the regenerating step. Reads accurately.

### (4) Tautological grep removed — `commands/repo/release.md:279`

Now `if [ "$VERSION_TOOL" = "npm" ] && [ -f VERSION ]; then`. Confirmed tautological before removal: `grep -n 'VERSION_TOOL=' commands/repo/release.md` shows `VERSION_TOOL="npm"` is assigned at exactly one site (line 213), whose own guard already requires `grep -q '"version"' package.json`. Two conditions now match the two-condition prose above the block.

## Test Execution

- `./hooks/repo/tests/run.sh` in `.loom/worktrees/issue-114` → **PASS: 1236  FAIL: 0** — matches the PR's claim and the pre-change baseline. No suite pins `release.md` prose, as the curation noted.
- CI: `pnpm test (guard/handoff/loss-check suite)` **pass** on head `a180f17`.
- Merge state: `CLEAN` / `MERGEABLE`, branch is a descendant of `origin/main` (`9dda363`).

## Fixed during evaluation

- `CHANGELOG.md` — the #114 entry described the replacement expression as `m.group(2)+sys.argv[1]+m.group(2)`, omitting the leading `m.group(1)` that carries the `[project]…version = ` prefix. As written it would describe an edit that deletes the prefix. Corrected to `m.group(1)+m.group(2)+sys.argv[1]+m.group(2)`, matching the actual lambda and the PR description. Docs-only, no behavior change (commit `a180f17`).

## Non-blocking note

`[^\"\x27\n]*` excludes *both* quote characters, so a double-quoted version string containing a literal apostrophe (`version = "1.0.0+don't"`) no longer matches where the old `\"[^\"]*\"` would have. Confirmed by fixture: it fails **loudly** through the zero-substitution guard (rc=1, `ERROR: no [project].version found`) rather than writing anything wrong. PEP 440 does not permit apostrophes in version strings, so this is unreachable in practice and the failure mode is the safe one — not worth complicating the character class or filing a follow-up.

Documentation and prose stay in sync with the snippets throughout: the Phase 5 bullet list correctly goes three → four, the new captured-quote bullet explains the `\x27` shell-quoting constraint, and the `uv.lock` bullet's "do not collapse this back into an `A && B && uv lock || :` chain" note records *why* the `if` form is load-bearing — a good guard against the exact regression being fixed here.

PR #128, opened 2026-08-04
docs(all): fetch before resolving origin/HEAD; split the switch failure modes

### LINKED ISSUE #115: /repo:all early sync-and-switch stage: pull-failure, worktree-collision, and fetch-ordering edge cases
## Context

Non-blocking follow-ups surfaced by the Judge during the `/loom:sweep all` review of PR #106 (issue #82, "sync to the default branch before Docs when the branch is fully pushed"). The PR was approved and merged; these three edge cases were posted to the PR review comment and would otherwise be lost. All concern the new early sync-and-switch **stage 2** in `commands/repo/all.md`. Related to but distinct from #98 (which is about `/repo:reset`'s step-1 dirty-tree fetch ordering); these are specifically about `all.md`'s new stage.

1. **"Failed checkout OR pull → change nothing" is inaccurate for the pull case.** `all.md` (lines ~117-119) says a failed checkout *or* pull means "change nothing." True for a checkout failure, but a `pull --ff-only` failure happens *after* the checkout already landed — the run is then left on a diverged local default branch, and the Docs stage would edit that stale checkout, which is the exact failure class this stage exists to prevent. No data loss, but the "change nothing" claim is wrong for the pull-failure path.

2. **Worktree-collision path quietly no-ops where Loom agents actually run.** `git checkout <default>` returns `fatal: '<default>' is already used by worktree at ...` (exit 128, HEAD unchanged) whenever the default branch is checked out in another worktree — the common case in a Loom-managed repo. Handled safely by the documented fallback, but it is not named, so the feature silently no-ops in exactly the environment it was built for.

3. **`git fetch` runs after `default` is computed.** The detection computes `default` (from `origin/HEAD`) and only then runs `git fetch origin`, so an unset local `origin/HEAD` cannot be resolved by that fetch in time — conservative and safe (the stage just no-ops), but worth fetching before resolving `origin/HEAD` so a fresh clone/first-run can become eligible.

- [ ] The pull-failure-after-checkout case is surfaced distinctly from checkout-failure (not folded into "change nothing"), so an operator/agent knows the run is left on a diverged local default.
- [ ] The worktree-collision no-op (default branch already checked out elsewhere) is documented and/or handled explicitly rather than quietly skipped.
- [ ] `origin/HEAD` is resolvable on first run — fetch before computing `default`, or otherwise handle an initially-unset `origin/HEAD`.

rjwalters:
## Summary

Fixes the three edge cases left over from `/repo:all`'s early sync-and-switch stage (#82 / PR #106): the eligibility fetch ran too late to be useful, the "change nothing" fallback was only half true, and the commonest refusal in a Loom-managed repo (default branch held by another worktree) was unnamed.

## Changes

- **`commands/repo/all.md` (stage 2 bash block)** — `git fetch origin --quiet` now runs *before* `default=$(git symbolic-ref ... refs/remotes/origin/HEAD ...)`. `default` reads a local ref, so resolving it first meant a stale (or, on a fresh / `--single-branch` clone, missing) `origin/HEAD` could not be refreshed by the very fetch meant to supply it, and the stage silently no-opped on its first opportunity to be eligible. The residual case — a repo that has *never* had `origin/HEAD` set locally, which a bare `git fetch` does not populate — is named in the eligibility table with its one-off fix, `git remote set-head origin --auto` (the curation flagged closing this fully as out of scope).
- **`commands/repo/all.md` (failure prose)** — the single "If the checkout or the `--ff-only` pull fails … change nothing" sentence is split into two bullets. A refused checkout leaves HEAD untouched ("change nothing" survives verbatim there). A failed `--ff-only` pull happens *after* the switch already landed, so the run is left on a diverged local default branch that Docs, Tidy, and Update tools then read for the rest of the run — it must say which branch it is on and that it is diverged, not report a clean no-op.
- **`commands/repo/all.md` + `commands/repo/reset.md` (step 4)** — both now name the worktree-collision refusal, `fatal: '<default>' is already used by worktree at '<path>'` (exit 128, HEAD unchanged), as an expected, recognized no-op rather than a generic "report why". Wording follows the existing precedent in `.loom/roles/judge.md`. `reset.md` also gains the checkout-succeeded/pull-failed distinction for standalone `/repo:reset` runs.
- **`commands/repo/tests/test-early-sync-switch.sh`** — `early_sync_eligible()` mirrors the corrected fetch order (its docstring claims to be a direct transcription of the doc, so drift there is a bug in the suite); the pinned `change nothing, report why, and continue with the stage order unchanged` assertion is repinned to the checkout-failure branch only and joined by assertions for the worktree-collision string, the diverged-default claim, and an `assert_not_contains` drift guard against the old combined wording; fetch ordering is asserted by **line position** so reordering the two lines back fails the suite; and a new fixture case stands up a second worktree holding the default branch and asserts the real git behaviour (exit 128, error names the colliding worktree, HEAD and working tree unmoved, `sync_and_switch` -> FAILED).
- **`CHANGELOG.md`** — entry under `## Unreleased`.

| Criterion (from issue) | Status | Verification |
|---|---|---|
| Pull-failure-after-checkout surfaced distinctly from checkout-failure | ✅ | `all.md` bullets 1 and 2; asserted by `all.md says a failed PULL leaves the run on the diverged default` + `all.md warns the later stages read that diverged copy` + `assert_not_contains` on the old combined sentence |
| Worktree-collision no-op documented/handled explicitly | ✅ | Named in `all.md` and `reset.md`; asserted in both files' doc-drift blocks and exercised for real by the new `worktree collision` fixture section (8 assertions against a live second worktree) |
| `origin/HEAD` resolvable on first run — fetch before computing `default` | ✅ | Lines reordered in `all.md` and in the test's mirror; guarded by the `all.md fetches before resolving origin/HEAD` line-position assertion; the never-set-locally residual is documented with `git remote set-head origin --auto` |
| Pinned test substrings survive verbatim except the one flagged for update | ✅ | Only `change nothing, report why, and continue with the stage order unchanged` changed meaning (now scoped to the checkout branch, still asserted verbatim); every other pinned substring listed in the curation is untouched and still passing |

## Test Plan

- `./commands/repo/tests/test-early-sync-switch.sh` — **84 pass / 0 fail / 0 skip** (68 before this PR; +16 assertions).
- `./hooks/repo/tests/run.sh` (full suite) — **1252 pass / 0 fail**, no regressions in unrelated suites.
- The worktree-collision behaviour is verified against real git rather than asserted only in prose: the fixture creates a second worktree on the default branch, then confirms `git checkout <default>` exits 128, emits `already used by worktree at`, and leaves both HEAD and `git status --porcelain` unchanged.

Closes #115


rjwalters:
✅ **Approved**

Reviewed against issue #115 and the curation comment. All three sub-items land, and the test changes are real guards rather than restatements of the prose.

### Verification of the three sub-items

**1. Fetch before `default=`** — `commands/repo/all.md:79-80` now runs `git fetch origin --quiet` ahead of the `refs/remotes/origin/HEAD` resolution, with an inline comment explaining that `default` reads a *local* ref. The residual case the curation explicitly scoped out (a repo that never had `origin/HEAD` set locally, which a bare fetch does not populate) is documented in the eligibility table with its one-off fix `git remote set-head origin --auto` — named as a caveat, not silently fixed, which is the right call for the issue's framing.

**2. Split failure prose** — the single combined sentence is now two bullets. The refused-checkout bullet keeps `change nothing, report why, and continue with the stage order unchanged` verbatim (accurate there — HEAD never moved). The pull-failure bullet says the switch already landed, names the diverged local default branch, and names Docs/Tidy/Update tools as the downstream readers of that copy. That is exactly the correction the curation asked for: the stage must not report a clean no-op from a state it materially changed. `reset.md` step 4 gets the same distinction for standalone `/repo:reset`.

**3. Worktree collision** — named in both `all.md` (checkout-failure bullet) and `reset.md:169-174` with the literal `fatal: '<default>' is already used by worktree at '<path>'`, exit 128, HEAD unchanged, framed as "an expected, recognized no-op, not a generic failure". Wording follows the `.loom/roles/judge.md` precedent the curation pointed at.

### Test verification

Ran the suite myself in `.loom/worktrees/issue-115`:

- `./commands/repo/tests/test-early-sync-switch.sh` — **84 pass / 0 fail / 0 skip** (claim confirmed, 68 → 84).
- `./hooks/repo/tests/run.sh` — **1252 pass / 0 fail** across all 10 suites, no regressions.

I mutation-tested the two new drift guards rather than trusting a green run:

- Swapping `all.md`'s fetch/`default=` lines back to the buggy order → `FAIL: all.md fetches before resolving origin/HEAD`. The line-position assertion is a real guard, not a prose match that would survive a reorder.
- Reintroducing the old combined `If the checkout or the --ff-only pull fails ... change nothing` sentence → `FAIL: all.md no longer folds the pull failure into 'change nothing'`. The `assert_not_contains` guard fires.

The repinning is handled correctly: `change nothing, report why, and continue with the stage order unchanged` is still asserted verbatim but is now scoped to the checkout-failure branch, so the one substring the curation flagged as *expected to change meaning* changed meaning without going stale. Every other pinned substring in the curation's do-not-alter list is untouched and still passing.

The new `worktree collision` fixture is the strongest part of the PR: it stands up a real second worktree on the default branch and asserts git's actual behaviour (exit 128, `already used by worktree at` in stderr, HEAD and `git status --porcelain` both unmoved, `sync_and_switch` → `FAILED`) instead of only asserting that the doc mentions it. `COLLIDE_ERR="$(git ... 2>&1 >/dev/null)"` followed by `COLLIDE_RC=$?` is the correct idiom for capturing stderr and the substitution's exit status, and the fixture cleans up its worktree. `early_sync_eligible()` mirrors the corrected fetch order, keeping the docstring's "direct transcription" claim honest.

### Rebase performed during review

The PR was `DIRTY` against main after #127 merged — a both-prepend conflict on `CHANGELOG.md` `## Unreleased`. Resolved per the judge role's simple-conflict path: rebased onto `35e7457`, kept both entries newest-first (#115 above #127), dropped the stray blank line so the entry list matches the file's no-blank-line-between-entries convention. Re-ran both suites after the rebase (84/0 and 1252/0) and force-pushed. Merge state is now `CLEAN` and CI passes on the rebased head.

### Non-blocking note

`line_of()` uses `grep -n -F -m1`, so the fetch-ordering assertion pins the *first* occurrence of each literal in `all.md`. Today there is exactly one of each (lines 79 and 80), so the guard is exact. If a later change adds another `git fetch origin --quiet` anywhere above the eligibility block, the guard would compare the wrong pair and could false-pass. Not worth blocking on — the assertion is correct for the file as it stands — but worth knowing if that block ever grows.

Checks: CI pass, merge state CLEAN, `Closes #115` present.

PR #129, opened 2026-08-04
fix(guard): track an active quoted span's real close so it stops swallowing the command

### LINKED ISSUE #113: guard-destructive: a $(...) in a quoted span bails the lifecycle & rm-target matchers to allow
## Context

Surfaced during the `/loom:sweep all` review of PR #107 (issue #84, heredoc-aware guard lexer). While hunting heredoc bypasses, both the Doctor and the Judge independently confirmed a **pre-existing** deny→allow gap in `hooks/repo/guard-destructive.sh` that is orthogonal to the #84/#107/#108 heredoc work and was explicitly left out of scope with "should be filed separately."

A `$(` (command substitution) — or a backtick — anywhere in a **quoted span** makes both `lifecycle_or_cloud_reason()` and `extract_rm_targets()` bail out to *allow*. Concretely, on `main` today:

```
echo "$(id)" ; rm -rf /opt/some-vendor/important     # ALLOWED
echo "y"     ; rm -rf /opt/some-vendor/important     # DENIED
```

The only difference is the `$(id)` inside the earlier quoted argument. The `;`-separated outside-repo `rm -rf` is a genuine destructive command that the guard denies in the plain form but silently allows once a quoted `$()` appears earlier in the command line. The same shape reproduces with `;`-separated lifecycle ops and with a `# x`-in-quote variant.

This is a property of the quoted-span-with-command-substitution handling (the active-quoted-span handler re-opens the closing quote and the matchers fall through to allow), **not** introduced or worsened by the #84/#107/#108 heredoc changes — it predates them and is unchanged by them.

- [ ] A destructive command (outside-repo `rm -rf`, a lifecycle op, or any `ALWAYS_BLOCK`/ask-tier target) that appears after a quoted `$()`- or backtick-bearing segment on the same command line still denies/asks, matching the behavior when no `$()` is present.
- [ ] Decide and document whether the current bail-to-allow is intentional (e.g. a deliberate redaction/expansion-safety floor) or a fixable false-negative; if intentional, note why the risk is acceptable; if fixable, close it.
- [ ] Add regression cases to `hooks/repo/tests/test-guard-destructive.sh` covering `echo "$(id)" ; rm -rf <outside>`, the backtick variant, and the `# x`-in-quote variant, each asserting the destructive tail is denied/asked.
- [ ] Confirm no regression to the legitimate quoted-`$()` cases that should pass (e.g. a benign `gh api -f body="$(cat file)"` with no destructive tail).

rjwalters:
## Summary

`echo "$(id)" ; rm -rf /opt/some-vendor/important` was **allowed** by the guard while the identical command with an inert `echo "y"` prefix was **denied**. The only difference is the `$(id)` inside the earlier quoted argument.

A quoted span carrying a command substitution deliberately keeps its separators **ACTIVE** (the #3679/#3755 floor, so smuggled content *inside* `"$(a|halt)"` is never masked), so both shared lexers walk into the span character-by-character — and eventually reach that span's own closing quote. Neither remembered that this position was the **already-computed close** they had found milliseconds earlier, so the top-of-loop quote detector re-read it as the **opener of a brand-new span**:

- **`ml_segment()`** (backs `parse_force_ops()`, `lifecycle_or_cloud_reason()`, `extract_rm_targets()`) — with no later quote in the buffer, the unterminated-quote fallback swallowed the **entire remainder** into the current segment. The whole command became ONE segment whose command word is `echo`, so all three matchers fell through to allow. Reproduced on `;`, `&&`, `|` and raw-newline separators; with a backtick nested in double quotes; with a single-quoted `$( )` *lookalike* the shell never expands (the lexer's naive substring probe still marks it active); and with the issue's `# x`-in-quote variant.
- **`qsplit()`** (gates `command_has_shell_segment()` → `strip_datasink_literals()`) — identical defect, but its fallback advances one character, so the plain repro split by luck. A **later unrelated quote pair** anywhere in the command paired with the re-opened quote instead and swallowed every real separator between them as bogus inert text, hiding a pipe-to-shell so the narrowing redaction erased the payload before the catastrophic scan ran.

**Decision (issue AC #2): this is a fixable FALSE NEGATIVE, not an intentional redaction/expansion-safety floor.** "Keep separators active inside a substitution-bearing span" exists so content *inside* the substitution is not masked; it never implied disabling matching for the text *after* the span. This rationale is recorded in the code comment at the fix site, not only here.

## Changes

- `hooks/repo/guard-destructive.sh`
  - `ml_segment()` (`_ML_QSPLIT_AWK`) and `qsplit()` (`_QSPLIT_AWK`) now push the already-computed close index onto a small **stack** when taking the active-span branch, and check it at the top of the character walk *before* quote re-detection. A stack rather than a scalar: with **nested** active spans an inner span would overwrite a single scalar and leave the outer close still mis-read (covered by a test).
  - A stale-entry pop (`while (acn > 0 && acs[acn] < i) acn--`) keeps the stack correct across the jumps the lexers already make (inert-span copy, heredoc opener/body skip).
  - Both fixes are deliberately identical so the two lexers cannot drift, per the files' own shared-algorithm header comments; those headers and the segmentation contract bullets are updated.
  - **Backslash-ambiguous quote pairings are never trusted (review follow-up).** The close-index scan finds the *next* quote of the same kind, which for `echo "$(a \" b)" ; <destructive>` is the **escaped** one — promoting it ended the span early, left the real close to open a bogus span, and brought the swallow back on shapes that denied *before* this PR. A shared `trusted_close()` now skips escaped candidates and records **nothing** when the candidate directly follows a backslash run (`\\"`, where the shell re-parses quoting inside `$( )`), so that span is walked exactly the legacy way. Two companion rules keep the same direction: an escaped quote never **opens** a span (an escaped quote sitting *after* an active span used to pair with a much later quote and copy every real separator between them as bogus inert text), and `ml_segment()`'s unterminated-quote fallback now advances **one character** with separators still active, matching `qsplit()`, instead of copying the rest of the buffer verbatim.
  - `bs_escaped()` and `trusted_close()` moved into a new shared `_ESCAPE_AWK` source string prepended to **both** lexer sources — `qsplit()` previously had no access to the parity helper at all, so it could not have applied the same rule without drifting.
- `hooks/repo/tests/test-guard-destructive.sh` — new `#113` section (33 cases: 17 for the close-index tracking, 16 for the backslash-escape family) after the `#3755` block, using the existing `assert_deny`/`assert_ask_env`/`assert_allow` helpers and the runtime danger-phrase assembly convention from `#60`/`#71`/`#84`.

| `parse_force_ops()`, `lifecycle_or_cloud_reason()`, `extract_rm_targets()` all segment correctly when a destructive/ask-worthy command follows a separator-separated quoted substitution span (double/single-quoted, `$(...)` or backtick) | ✅ | 11 new deny/ask cases + a control asserting the inert-prefix baseline behaves identically. Verified `deny`/`deny`/`ask` post-fix vs `allow`/`allow`/`allow` on `main` by driving both copies of the hook side by side. |
| `command_has_shell_segment()` (via `qsplit()`) gets the same fix | ✅ | New later-unrelated-quote case (`quoted $( ) prefix ; payload | sh ; trailing quoted string`) denies. Isolated the qsplit contribution by reverting *only* the `qsplit()` push in a scratch copy — that case reverts to allow, proving the fix is load-bearing and not masked by `ml_segment()`. |
| The existing `#3755` regression (quoted command substitution containing a piped lifecycle word still denies) continues to pass | ✅ | Unchanged and passing; plus a new case asserting it still denies *with a trailing tail appended*, so in-span matching is provably untouched. |
| Documented decision that this is a fixable false negative, not an intentional safety floor | ✅ | Recorded in the `ml_segment()` fix-site comment, the two lexer header blocks, the CHANGELOG entry, and this PR. |

## Test Plan

- `hooks/repo/tests/test-guard-destructive.sh`: **577 pass / 0 fail** (was 544 — 33 new `#113` cases).
- `./hooks/repo/tests/run.sh` (full suite): **1285 pass / 0 fail** (was 1252 on this base).
- **Differential sweep vs the pre-change guard** — 1392 cases (14 span bodies × 2 quote chars × 6 tails, plus an escaped-quote-after-span matrix, plus 1200 fuzzed token strings), `LOOM_FORCE_SCOPE=all` pinned, fixtures built on disk: **0 deny → allow**, 159 previously-swallowed bypasses newly denied. The first close-index commit alone scored 40 deny → allow on the same sweep.
- `shellcheck -S warning` on both changed shell files: no new findings (the 2 findings in the test file are pre-existing on `main` at lines 2740/2882, outside this diff).
- Guard performance check still passes (27ms avg, < 200ms threshold).
- Behaviour diff captured by running the `main` hook and the fixed hook against the same fixture set: every issue-repro shape flips `allow` → `deny`/`ask`, while `gh api -f body="$(cat file)"` with no destructive tail, a bare quoted `$( )` span, a benign tail, and the `#3755` quoted-alternation allow all stay `allow`.

New cases cover: the issue repro (rm-scope), the lifecycle tier, backtick-in-double-quotes, the single-quoted lookalike, `# x`-in-quote, `&&`/`|`/newline separators, the `parse_force_ops()` ask-tier force-push (pinned with `LOOM_FORCE_SCOPE=all` per the `#3913` ambient-env lesson), nested active spans, the `qsplit()` later-quote shape, the `#3755` in-span floor with a trailing tail, and four allow-side no-regression cases.

Closes #113


rjwalters:
❌ **Changes Requested — the "strictly widening" invariant does not hold; this diff turns an existing `deny` into an `allow`**

Reviewed with extra rigor as a security-relevant guard change. The core diagnosis and the primary fix are **correct and well-evidenced** — but I found a systematic regression family the new test block does not cover, and it contradicts the safety claim that is written into the code comment, the CHANGELOG, and the PR body.

---

## ✅ What verified clean

I drove `origin/main`'s hook and this branch's hook side by side against fixture command strings built on disk (never inline), `LOOM_FORCE_SCOPE=all` pinned:

| Shape | main | branch |
|---|---|---|
| `echo "$(id)" ; <outside-repo rm>` | allow | **deny** |
| `echo "$(id)" ; <lifecycle>` | allow | **deny** |
| double-quoted backtick prefix + rm | allow | **deny** |
| single-quoted `$( )` lookalike + rm | allow | **deny** |
| `"# x $(id)"` + rm | allow | **deny** |
| `&&` / `|` / raw-newline separators | allow | **deny** |
| force-push tail (`parse_force_ops()`) | allow | **ask** |
| nested spans `"$(a '$(b)')"` + rm | allow | **deny** |
| qsplit later-quote + pipe-to-shell | allow | **deny** |
| `#3755` in-span smuggle (± trailing tail) | deny | deny |
| `gh api -f body="$(cat file)"`, bare span, benign tail, inert alternation, `gh api` + benign tail | allow | allow |

- **The stack-over-scalar justification is empirically confirmed.** I mechanically rewrote both lexers to the curator's suggested scalar (`ac0`, with the same stale-reset) and re-ran the matrix: every case matched the stack build **except** the nested-span case, which reverts to `allow`. The stack is load-bearing, not over-engineering, and the nested-span test is the case that pins it. Good call overriding the curation here.
- `hooks/repo/tests/test-guard-destructive.sh` → **561 pass / 0 fail**; `./hooks/repo/tests/run.sh` → **1253 pass / 0 fail**. Both claimed counts are accurate.
- CI green; the `#3755` block is untouched and still denies.
- **CHANGELOG conflict resolved during review** (both-prepend against #127/#128) — rebased onto `f5bca02` and force-pushed; `mergeStateStatus` is no longer `DIRTY`. No code was changed.

---

## 🔴 Blocker: a backslash-escaped quote inside the active span flips deny → allow

```
echo "$(a \" b)" ; rm -rf /opt/some-vendor/important
```

- `origin/main`: **deny**
- this branch: **allow**

`ml_segment()` segmentation, same input:

```
main:  SEG[1]=<echo "$(a \" b)" >
       SEG[2]=< rm -rf /opt/some-vendor/important>          <- correct, denies

fix:   SEG[1]=<echo "$(a \" b)" ; rm -rf /opt/some-vendor/important>   <- one segment, command word 'echo', allows
```

This is not a one-off. Across a structured sweep of 168 shapes (14 span bodies × 2 quote chars × 6 tails), **42 cases regress `deny` → `allow`** on this branch, all in this family, on every separator (`;`, `&&`, `|`, newline) and with or without a trailing quoted string. Both the `\"`-inside-span and the `\"`-at-end-of-span placements reproduce.

### Mechanism

`hooks/repo/guard-destructive.sh:1176` — the forward scan that computes `ci` does **not** consult `bs_escaped()`:

```awk
for (j = i + 1; j <= n; j++) if (substr(s, j, 1) == qc) { ci = j; break }
```

so for `echo "$(a \" b)" ; …` it returns the **backslash-escaped** quote (index 12), not the real close (index 16).

On `main` a wrong `ci` was harmless in the active branch: the branch performed **no jump**, so the walk re-paired quotes from that position forward and happened to land correctly on the real close (`inner = " b)"`, inert, jump to 17) — the `;` then split and the `rm` denied. This diff promotes that best-effort `ci` to an **authoritative** close index and consumes it at line 1167, so the walk resumes at 13 with no span memory, the *real* close at 16 is read as a fresh opener, finds no partner, and falls into the **unterminated-quote fallback** (`hooks/repo/guard-destructive.sh:1178-1181`) that swallows the whole remainder — the exact failure mode this PR exists to remove, newly triggered on a shape that previously denied.

Note that the shared header this PR edits still says, three lines below the new `#113` paragraph (`hooks/repo/guard-destructive.sh:875-877`): *"backslash-escaped quotes and an unterminated quote fall back to the old separator-active behaviour, **never widening a deny into an allow**."* That sentence is now inaccurate — the new code consumes the best-effort `ci` on exactly the path the sentence promises is safe. `bs_escaped()`'s own comment at `:1119-1120` makes the same promise for its existing call sites ("strictly NARROWING … cannot widen a deny"); trusting an escaped `ci` is the first use that points the other way.

### Suggested remedy (validated)

Distrust a close index that is backslash-escaped — fall back to the legacy walk for that span only:

```awk
if (!bs_escaped(s, ci)) acs[++acn] = ci
```

I applied exactly this to `ml_segment()` and re-ran everything: **all 12 bypass shapes still deny/ask, all 5 allow-side cases still allow, `#3755` still denies, and the weakened family drops 42 → 6.**

Two follow-ons for whoever picks this up:

1. **The residual 6** are `"$(a \\" b)"` — an *escaped backslash* followed by a real close, so `bs_escaped()` correctly reports "not escaped", yet `main` still denies and the branch still allows. Worth deciding deliberately: either teach the forward scan itself to skip escaped quotes, or add a conservative "if the recomputed pairing disagrees, don't push" rule. Whatever you choose, please pin it with a test.
2. **`qsplit()` cannot call `bs_escaped()`** — the helper is defined at `:1121`, inside `_ML_QSPLIT_AWK` (`:1061-1306`), while `_QSPLIT_AWK` spans `:881-948`. The `qsplit()` push at `:930` has the identical hazard, so it needs its own copy of the helper (or the helper hoisted into a third shared string), consistent with the files' own "don't let the two lexers drift" rule.

### Also please update

- The code comment at `hooks/repo/guard-destructive.sh:1163-1165` ("Strictly widening for segmentation … can never turn an existing deny into an allow"), the matching CHANGELOG sentence, and the PR body — the claim is currently false as written. Once the escaped-quote case is handled the claim can stand; until then it should not ship as a documented invariant.
- Add `#113` regression cases for the escaped-quote family (at minimum the `\"`-inside-span shape above on `;` plus one `&&`/newline variant), since the current 17 cases pass while the regression is live.

---

Everything else here is genuinely good work — the root-cause writeup is accurate, the stack choice is justified and now empirically proven, the two lexers were kept in lockstep, and the allow-side no-regression set is well chosen. This is one missing interaction with a pre-existing best-effort limitation, not a rethink.

Re-request review once the escaped-quote path is handled and I'll re-run the same differential matrix.

rjwalters:
🩺 Blocker addressed — the escaped-quote regression family is closed, and the never-widening claim is now backed by a sweep instead of asserted.

Thank you for the differential matrix; it found a real deny → allow family and the reproduction was exact. Pushed as `8ada3ba`.

## What changed

I started from your validated gate and then swept it, which turned up a **second** family your 168-shape matrix did not cover, so the final shape is a little broader than `if (!bs_escaped(s, ci))`:

1. **`trusted_close()` (new, shared)** — resolves the close index an ACTIVE span records. It skips backslash-escaped candidates (`\"` is literal text, never a real close) and returns `0` — *record nothing, walk this span the legacy way* — when no unescaped candidate exists **or** when the candidate directly follows a backslash run. That last clause is your residual 6 (`"$(a \\" b)"`): the quote really is unescaped, but the shell re-parses quoting inside `$( )`, so the pairing is ambiguous and the pre-#113 walk is the conservative answer.
2. **An escaped quote never OPENS a span** (both lexers). This is the family the sweep added: `echo "$(id)" \" ; <destructive>` — the escape sits *after* the span, so no close-index gate can see it. On `main` the mis-paired close accidentally paired with it; once the real close is consumed, that `\"` became an opener, paired with a much later quote, and copied every real separator between them as bogus inert text. 12 of the 40 regressions were this shape.
3. **`ml_segment()`'s unterminated-quote fallback now advances one character** with separators still ACTIVE, matching `qsplit()`'s already-documented behaviour, instead of copying the rest of the buffer verbatim. That fallback *is* the swallow this PR exists to remove; leaving it armed meant any stray quote ahead of a real `; <destructive>` re-armed it.

`bs_escaped()` + `trusted_close()` now live in a new `_ESCAPE_AWK` source string prepended to **both** lexer sources (your follow-on #2). `qsplit()` had no access to the helper at all, so a duplicate would have been the only alternative; the file's own "the parsers cannot drift" rule pushed me to hoist instead. Four `awk "$_..."` call sites updated.

## Before / after sweep

Same methodology as yours — fixtures built on disk, `LOOM_FORCE_SCOPE=all` pinned, `origin/main`'s hook and the branch hook driven side by side. I widened the matrix to 1392 cases: 14 span bodies × 2 quote chars × 6 tails (168, including `\"`-inside-span, `\"`-at-end, `\\"`, and benign backslashes), a 24-case escaped-quote-**after**-span matrix, and 1200 fuzzed token strings over a quoting/escape/separator/payload alphabet.

| Guard under test | deny → allow | allow → deny |
|---|---|---|
| `ab5917c` (PR tip you reviewed) | **40** (28 in-span, 12 after-span) | 94 |
| your gate alone (`!bs_escaped(s, ci)`, ml_segment only) | 6 in-span + the 12 after-span family | — |
| gate + rules 2 and 3, both lexers, **pushed here** | **0** | 159 |

Your 20-fixture matrix still reads: A01–A08, A10–A12 `deny`, A09 `ask`, B01–B03 `deny`, C01–C05 `allow`. 15 realistic backslash/quote-heavy benign commands (`git commit -m "… \"quote\" …" && git push`, `find . -name "*.tmp" -exec rm {} \;`, `sed -i "" "s/a\"b/c/g" file.txt`, …) are byte-identical `allow` on both guards, so the extra strictness is not leaking into false positives.

## Residual 6

Not out of scope after all — they are fixed, by clause 2 of `trusted_close()`. `"$(a \\" b)"` now records no close at all and reproduces the pre-#113 walk exactly, which is what denies. Pinned by two tests (with and without a trailing quoted string; the trailing-string variant is the one that turns a mis-paired close into a separator-swallowing inert span).

## Tests

16 new cases (561 → **577** in `test-guard-destructive.sh`; full suite **1285 pass / 0 fail**): `\"` inside the span on `;` and `&&`, single- and double-quoted, `\"` at the end of the span, `\\"` with and without a trailing quoted string, `\"` after the span with and without a trailing quoted string, the `qsplit()` pipe-to-shell shape, a newline separator, the `parse_force_ops()` ask-tier variant, and four allow-side cases (benign tail, escaped quote in an inert span, escaped quote after the span, `gh api` body carrying an escaped quote). **All 12 deny/ask cases ALLOW on `ab5917c`**, so every one of them pins a regression this commit removes rather than pre-existing behaviour.

`shellcheck -S warning`: clean on the guard; the two findings in the test file are pre-existing and outside this diff. Guard perf check still passes.

## Prose

The three claims you flagged are updated, not just softened: the shared header (`guard-destructive.sh:926-937`) now enumerates the three narrowing rules instead of asserting "backslash-escaped quotes … fall back", the fix-site comment states the invariant holds *given* those rules, and the CHANGELOG says the same and carries the sweep numbers. `bs_escaped()`'s own header gains the quote call site.

rjwalters:
❌ **Changes Requested**

The fix is real and it works — the issue repro flips `allow` → `deny`, the escaped-quote follow-up commit genuinely holds, and I measured a large net gain in coverage. But an **independent differential sweep found 30 `deny` → `allow` transitions** on a shape family this PR's own sweep did not cover. None are executable, which is why I'm asking for a scoped correction rather than a redesign — but the PR currently states an invariant that is demonstrably false, and that claim is written into the code comments and the CHANGELOG where future maintainers will rely on it.

## What I verified (all confirmed)

| Check | Result |
|---|---|
| `test-guard-destructive.sh` | **577 pass / 0 fail** — matches the PR claim exactly |
| `hooks/repo/tests/run.sh` | **1285 pass / 0 fail** — matches exactly |
| CI (`pnpm test`) | ✅ SUCCESS |
| `mergeStateStatus` / branch currency | `CLEAN`, `MERGEABLE`, `origin/main` is an ancestor of HEAD |
| `shellcheck -S warning` | Guard file **clean**; test file has 2 findings, and I confirmed **identical 2 findings on `main`** — pre-existing, as claimed |
| `Closes #113` | Present (line 43) |
| All four lexer call sites prepend `$_ESCAPE_AWK` | ✅ lines 1413, 1676, 1960, 2098 — no orphaned call site that would hit an undefined `bs_escaped()` |
| `trusted_close()` termination | ✅ `ci` strictly increases each iteration, so the `while` cannot loop forever |

**The fix does what it claims.** On a sweep of 280 *parseable* shapes (14 balanced prefixes × 5 destructive tails × 4 suffixes), I measured **176 newly-caught bypasses and 0 regressions**. The escaped-quote family from the second commit checks out: `echo "$(a \" b)" ; <rm>`, `echo "$(id)" \" ; <rm>` and `echo "$(a \\" b)" ; <rm>` all deny on both `main` and this branch, exactly as the "no-regression pins" comment says.

## Blocking finding: 30 `deny` → `allow` transitions

My sweep (1,866 cases: 15 span bodies × 2 quote chars × 5 separators × 10 tails, plus an escaped-quote-after-span matrix and benign controls) found 30 cases where **`main` denies and this branch allows**. Four distinct shapes, all requiring a trailing quoted string:

```
echo "$(id)" "      ; <destructive> ; echo "trailing"     main=deny  pr=allow
echo "$(id)" \\\\"  ; <destructive> ; echo "trailing"     main=deny  pr=allow
echo "`id`"  "      ; <destructive> ; echo "trailing"     main=deny  pr=allow
echo "`id`"  \\\\"  ; <destructive> ; echo "trailing"     main=deny  pr=allow
```

Driving `ml_segment()` directly from both versions of the file shows this is the very swallow the PR exists to remove, reappearing:

```
COMMAND: echo "$(id)" " ; rm -rf /opt/some-vendor/important ; echo "trailing"

--- MAIN segments ---
  seg[1] = <echo "$(id)" " >
  seg[2] = < rm -rf /opt/some-vendor/important >     <-- own segment, matched, DENY
  seg[3] = < echo "trailing">

--- PR segments ---
  seg[1] = <echo "$(id)" " ; rm -rf /opt/some-vendor/important ; echo "trailing">
                                                     <-- ONE segment, command word `echo`, ALLOW
```

**Mechanism.** On `main`, the old defect — re-reading the active span's close as a new opener — happened to pair that close with the *stray* `"`, making the one-space gap between them an inert span and leaving `; <destructive>` walked with separators **active**, so it split and denied. This PR correctly stops that re-read, which frees the stray `"` to pair instead with the opening quote of the trailing `"trailing"`. The text between them contains no `$(` and no backtick, so it takes the **inert-span branch** and is copied verbatim — swallowing both real `;` separators. The fix realigned the quote pairing onto the payload.

So the residual weakness is the pre-existing inert-span copy, but this PR is what routes these inputs into it. Against `main`, that is a regression.

**Why the existing sweep missed it:** the PR's matrix varies the span *body* and the *tail*, but always with a balanced prefix. The trigger here is an **unbalanced stray quote between the span and the tail**, combined with a later quoted string to pair with. Neither the 1,392-case matrix nor the 33 new test cases contain that combination.

## Severity — why I'm not calling this an exploitable bypass

All 30 regressing strings **fail `bash -n`**. Of the 30, `bash -n` accepts **0**. The guard and bash agree on the quote pairing in these shapes, so the text the guard swallows is text bash also treats as quoted — the destructive command is a literal string argument and never executes, and the leftover quote makes the whole command a syntax error. I tried to construct an executable member of this family and could not; across the 280 parseable shapes I swept, there were **zero** regressions.

So: a real regression against the PR's stated standard, but not a route to executing a destructive command.

## What I'm asking for

Either fix is fine — the choice is yours, and option B is legitimate given the non-executability:

**Option A — close the shape.** Make an unbalanced/stray quote conservative, so the inert-span branch cannot pair across a position that a preceding active span already consumed.

**Option B — correct the claim and pin the limitation.** If this family is accepted as non-executable, then the following statements need narrowing, because as written they are false:

- `guard-destructive.sh:~899` — *"Every use is strictly NARROWING ... never widening a deny into an allow"*
- `guard-destructive.sh:~927` — *"None of these can widen a deny into an allow"*
- `guard-destructive.sh:~1215` — *"this is strictly widening for segmentation (it can only produce MORE segment boundaries than before), so it can never turn an existing deny into an allow"* — the segment dump above is a direct counterexample: 3 boundaries became 1
- CHANGELOG — *"a 1392-case differential sweep ... now reports **0 deny → allow**"* and *"Segmentation only ever gains boundaries, so no existing deny becomes an allow"*
- `test-guard-destructive.sh:~1498` — *"Every case below denies on the pre-#113 guard, so they are no-regression pins"* — true of the cases present, but the family that does regress has no pin

Suggested wording for B: the invariant holds for **parseable** commands, with the unbalanced-stray-quote family documented as a known, non-executable exception. Please add the four shapes above as explicit test cases (asserting whatever behavior you decide is correct) so the next lexer change can't move them silently.

## Related, pre-existing — not part of this PR

`echo "$( ' )" ; <destructive> '` allows on **both** `main` and this branch (same inert-span-swallow family, also unparseable). Not a regression and not in scope here; I'll file it separately if you'd like, or it can fold into whichever option you pick.

## Reproducing

```bash
git show origin/main:hooks/repo/guard-destructive.sh > /tmp/guard-main.sh
chmod +x /tmp/guard-main.sh
CMD='echo "$(id)" " ; rm -rf /opt/some-vendor/important ; echo "trailing"'
for g in /tmp/guard-main.sh hooks/repo/guard-destructive.sh; do
  jq -n --arg c "$CMD" --arg w "$PWD" \
    '{tool_name:"Bash",tool_input:{command:$c},cwd:$w}' \
  | "$g" | jq -r '.hookSpecificOutput.permissionDecision // "allow"'
done
# main -> deny ; this branch -> allow
```

The analysis, the awk segmentation work, and the escaped-quote handling in the second commit are all high quality — this is a narrow gap in an otherwise strong fix, and the net effect on real commands is clearly positive.

rjwalters:
🔧 **Addressed — Option B (scope the claim, pin the exception)**

Pushed `6015cc9`. I took Option B because the family is non-executable and the deeper weakness is the inert-span branch, which is now filed separately as #130 rather than being reworked inside a PR that is already load-bearing.

**Behaviour is unchanged.** Re-running the parseable-only sweep against the edited guard gives byte-identical results to before these edits — **176 newly caught, 0 regressions** across 280 shapes. What changed is that the documentation now matches what the code does.

**The three code claims, narrowed** (all in `hooks/repo/guard-destructive.sh`):

| Was | Now |
|---|---|
| "Every use is strictly NARROWING … never widening a deny into an allow" | scoped to *these helpers*, with a pointer to the KNOWN LIMIT |
| "None of these can widen a deny into an allow" | "For every command the shell can actually PARSE, none of these widens…" |
| "strictly widening for segmentation … can never turn an existing deny into an allow" | holds for parseable input, followed by a full KNOWN LIMIT block naming the shape, the mechanism, and why nothing executes |

Both inert-span branches (`qsplit()` and `ml_segment()`) now carry a pointer identifying that branch as the one place a segment boundary can be lost, since it consumes whatever the forward scan paired with.

**11 new test cases** — the safety argument is now mechanical rather than a comment:

- **4 × `assert_shell_rejects`** — a new helper asserting `bash -n` refuses each shape. This is the load-bearing half: if a future change ever makes one of these parseable, the suite fails loudly instead of the allow sitting there unnoticed.
- **4 × `assert_allow`** — pinning the current behaviour so it cannot move silently.
- **3 controls** — the balanced-quote versions (`""`, `"q"`, and the shape with no later quote pair) still **deny**, confirming the limit has not leaked into real commands.

The "every case below denies on the pre-#113 guard" header is now scoped to its own sub-block, since the new one is explicitly the exception.

**CHANGELOG** — replaced *"Segmentation only ever gains boundaries, so no existing deny becomes an allow"* with the scoped claim plus the documented exception and the shape that triggers it, added the parseable-only sweep alongside the original 1392-case one, and updated counts (577 → 588, 1285 → 1296).

**Verification:** 588 guard cases / 1296 total, 0 fail. `shellcheck -S warning` clean on the guard; the 2 test-file findings remain the pre-existing pair also present on `main`.

⚠️ One trap worth knowing for future edits, which cost me a full red suite: the KNOWN LIMIT block lives **inside the single-quoted `_ML_QSPLIT_AWK` source string**, so a single apostrophe (I wrote "span's") terminates the shell string and breaks every awk parser in the file. There is now a note in that block saying so.

rjwalters:
✅ **Approved!**

Re-reviewed at `6015cc9`. The change-request is resolved, and I verified the resolution independently rather than taking the fix commit at its word.

> **Disclosure:** I authored the fix commit as well as this review. Loom's label workflow expects that, but it means the verification below is doing the real work, not my judgment of my own patch.

**The blocking finding is addressed.** The three overclaimed invariants in `guard-destructive.sh`, the CHANGELOG sentence, and the test-block header are all scoped to parseable input, and the exception is documented at the point where it actually lives — the inert-span branch — in both lexers.

**Verified the fix is documentation-only, as claimed.** Filtering the last commit's guard diff to non-comment lines returns **nothing** — every guard change is a comment. The behaviour that CI and the suite exercise is byte-identical to the version I reviewed before.

| Check | Result |
|---|---|
| `test-guard-destructive.sh` | **588 pass / 0 fail** (577 → 588, +11) |
| `hooks/repo/tests/run.sh` | **1296 pass / 0 fail** (1285 → 1296) |
| CI (`pnpm test`) | ✅ SUCCESS |
| `mergeStateStatus` / `mergeable` | `CLEAN` / `MERGEABLE` |
| `shellcheck -S warning` | Guard **clean**; test file 2 findings, still the pre-existing pair on `main` |
| Parseable-only differential sweep, re-run against the edited guard | **176 newly caught, 0 regressions** — identical to pre-edit |
| Guard diff, non-comment lines | **none** — comments only |

**What makes this acceptable rather than papered over:** the four `assert_shell_rejects` cases. The safety argument for allowing these shapes is "the shell will not parse them, so nothing executes" — and that is now asserted by the suite instead of asserted in a comment. If a future lexer or parser change ever makes one of those shapes parseable, the suite fails loudly and the allow gets revisited. That converts a prose defence into a mechanical one, which is the right shape for a security guard. The three balanced-quote controls confirm the limit has not leaked into real commands.

The deeper weakness — the inert branch consuming whatever the forward scan paired with — is tracked in #130, correctly scoped as pre-existing rather than folded into this PR.

**On the original PR:** the core fix is strong work. It closes a real bypass class (176 newly-caught shapes in my own sweep of parseable commands), the awk stack for nested spans is the right call over a scalar, `trusted_close()` provably terminates, all four lexer call sites correctly prepend `$_ESCAPE_AWK`, and the second commit's backslash handling holds up under direct testing. The only gap was that its safety claim was stated more broadly than the evidence supported — now it matches.

Ready to merge.

PR #131, opened 2026-08-05
docs(update-tools): use resync-installed.sh as Loom's update path (#119)

### LINKED ISSUE #119: update-tools: the prescribed Loom update command can never succeed on an existing install
## Context

`/repo:update-tools` step 4 prescribes this command for updating Loom:

```
# Loom:        <source>/install.sh --quick -y <this-repo>
```

That invocation cannot succeed in the only situation `update-tools` ever runs it — updating a tool that is *already installed*. Loom's `install.sh` refuses a non-interactive reinstall over an existing `.loom/`:

```
✗ Error: Existing Loom installation detected at <target>/.loom -- refusing to run a non-interactive reinstall without explicit acknowledgement.
       Reinstalling uninstalls the existing Loom payload before writing the new version; inventory and back up any project-owned Loom hooks, scripts, and agent configuration first.
       Re-run with --confirm-reinstall once you have done so, or omit --quick/--yes/--full to get an interactive y/N prompt instead.
```

Reproduced on `rjwalters/claude-monitor` taking Loom 0.16.0 → 0.18.0 (source clone clean and current, target tree clean).

The correct update path is `.loom/scripts/resync-installed.sh` in the target repo — non-destructive, idempotent, supports `--dry-run`, reports per-file updated/created/unchanged/skipped, and re-stamps `loom_version` / `loom_commit` / `last_resync` into `.loom/install-metadata.json` on success. That is also what the target repo's own history had used for its previous bump (`chore(tooling): resync loom install`).

Two things follow from the guidance as written:

1. An agent that follows it verbatim either reports the update as failed, or escalates to `--confirm-reinstall` — which uninstalls the existing payload first. That is a destructive step reached by following a routine hygiene command's instructions, and Safety Rule 2 ("always use the tool's own installer") reads as if it endorses it.

2. One caveat worth documenting alongside the fix: a resync updates `.loom/scripts/` *including `resync-installed.sh` itself*, so an old buggy copy can fail partway and a second run — now using the freshly synced script — completes. That happened here: the 0.16.0 copy died with `line 509: verb_past: unbound variable`, and re-running it with the 0.18.0 copy finished cleanly (70 further files updated). "Re-run once after a partial failure" is worth stating so the failure isn't misread as a broken update.

- [ ] Step 4's Loom line points at `<target>/.loom/scripts/resync-installed.sh` rather than `install.sh --quick -y`
- [ ] `--dry-run` is used first, matching the command's report-then-confirm posture
- [ ] The guidance notes that reinstall is the destructive fallback, not the default update path
- [ ] The self-syncing re-run caveat is documented
- [ ] Safety Rule 2 is worded so "the tool's own installer" includes a tool's own *update* mechanism where one exists

rjwalters:
Closes #119

## Problem

`/repo:update-tools` step 4 prescribed this for Loom:

```
# Loom:        <source>/install.sh --quick -y <this-repo>
```

That command cannot succeed in the **only** situation step 4 ever runs it — updating a tool that is already installed. Loom's `install.sh` refuses a non-interactive reinstall over an existing `.loom/` and exits with an error pointing at `--confirm-reinstall`. Reproduced on `rjwalters/claude-monitor` taking Loom 0.16.0 → 0.18.0 with a clean, current source clone and a clean target tree.

Following the guidance verbatim therefore ended in one of two bad outcomes: report the update as failed, or escalate to `--confirm-reinstall`, which **uninstalls the existing Loom payload before writing the new version** — a destructive step reached by following a routine hygiene command, with Safety Rule 2 ("always use the tool's own installer") reading as endorsement.

## Changes — `commands/repo/update-tools.md` only

- **Step 4's Loom row** is now a two-line `--dry-run`-then-apply pair pointing at `<this-repo>/.loom/scripts/resync-installed.sh`: non-destructive, idempotent, per-file updated/created/unchanged/skipped reporting, never clobbers a symlinked install target, and re-stamps `loom_version` / `loom_commit` / `last_resync` into `.loom/install-metadata.json` on a successful non-dry-run.
- **Prose after the block** calls out that the script lives in the **target** repo's `.loom/scripts/`, not the source clone — a structural difference from every other row's `<source>/…` shape, so this is not a mechanical path swap. It documents the `--dry-run` exit codes (`2` = drift, `0` = in sync, `1` = error) so the dry-run/apply split matches the command's existing report-then-confirm posture.
- **Reinstall is demoted to an explicit destructive fallback** (`install.sh --quick -y --confirm-reinstall`) in its own fenced block, with a back-up-project-owned-hooks/scripts/config reminder and an instruction not to escalate to it merely because a resync pass exited non-zero.
- **Self-syncing re-run caveat documented**: `resync-installed.sh` is part of the payload it syncs, so the copy that starts the run is the old one. A stale buggy copy can die partway (observed: `line 509: verb_past: unbound variable`) *after* writing the newer script to disk; re-running once picks up the fresh copy and completes (70 further files updated in that case). Stated so a single failed pass reads as "retry once", not "broken update".
- **Safety Rule 2 reworded** to "Always use the tool's own installer **or update mechanism**", preferring a dedicated non-destructive updater where one exists and naming installer-reinstall as the destructive fallback, while keeping the never-hand-copy floor. Rules 1/3/4/5 untouched, numbering unchanged.
- **Two one-phrase consistency touches** flowing from the same correction: the Usage intro says an update runs the tool's own "installer or updater", and step 5's parenthetical says "the installer/updater is idempotent and re-runnable" (Loom's installer specifically is not).

### Deliberately unchanged

Anvil, Repo Skills, kicad-tools, and Unknown-tools rows. This repo's own `install.sh` was checked and has **no** reinstall-refusal guard (its only non-interactive gate is a TTY check that `-y` bypasses cleanly), so Repo Skills' row is correct as written. Anvil's `install-anvil.sh` and kicad-tools' `install-kct.sh` could not be verified from this checkout — a follow-up should spot-check them for the same guard pattern rather than generalizing this fix blindly.

### Scope notes

- Edited the source tree only. `.claude/commands/repo/` is gitignored installed output (`.gitignore:7:.claude/`, 0 tracked files) — verified before editing.
- `CHANGELOG.md` gains a new `## Unreleased` section (0.8.0 consumed the previous one) with a full house-style entry.

## Test Plan

- `./hooks/repo/tests/run.sh` → **PASS: 1296   FAIL: 0** (unchanged from baseline; no suite pins this file's wording — `grep -rln "update-tools"` across `commands/repo/tests/` and `hooks/repo/tests/` matches only a comment in `test-install-sidecar-untracking.sh` about the sidecar `/repo:update-tools` reads, which is unaffected).
- Manual verification against the acceptance criteria: resync is the primary path, `--dry-run` comes first, reinstall is clearly marked destructive/fallback, the re-run caveat is present near the Loom row, Safety Rule 2 covers update mechanisms.
- Manual diff review: Anvil / Repo Skills / kicad-tools / Unknown rows are byte-identical in the diff; Safety Rules 1/3/4/5 keep their numbers and text.

https://claude.ai/code/session_01AHBrUgf4EmjEYkbEQdgZYA


rjwalters:
## Judge Review: APPROVED

Verified every load-bearing factual claim against the actual scripts rather than trusting the diff. All of them hold.

### Verification performed

**Exit codes are real** — `.loom/scripts/resync-installed.sh:652-658`: `--dry-run` with `N_UPDATED > 0` → `exit 2` ("would be updated"), otherwise `exit 0` ("already in sync"). Error paths `exit 1` at `:128` (unknown arg), `:137` (not a git repo), `:155` (no `.loom/` root), `:199` (no `defaults/` source tree). `N_UPDATED` is incremented for both created and updated files (`:284`, `:447`, `:455`), so "drift was found" is the accurate gloss for `2` — not just "modified". The documented `2`/`0`/`1` triple is exactly right.

**The reinstall guard is real, and the fix matches upstream's own advice.** `install.sh:1068` in the Loom source errors with `"refusing to run a non-interactive reinstall without explicit acknowledgement"` and then *itself* recommends `'$TARGET_PATH/.loom/scripts/resync-installed.sh'` as the non-destructive alternative. `install.sh:674-675` confirms reinstall "uninstalls before reinstalling". So this PR is not inventing a workaround — it aligns the command with the guidance Loom prints at the failure site.

**The path asymmetry is real and correctly explained.** `resync-installed.sh` does *not* exist in the source clone's `scripts/` (only `defaults/scripts/resync-installed.sh`, which is payload, not an entry point). It exists only at the installed target: `/Users/rwalters/GitHub/repo/.loom/scripts/resync-installed.sh`. A reader who mechanically applied the `<source>/…` shape of the surrounding rows would get file-not-found, so calling the difference out explicitly in bold prose was the right call, not decoration.

**Supporting claims spot-checked:** symlink-skip (`:267-272`), metadata re-stamp of `loom_version`/`loom_commit`/`last_resync` (`:466-513`), and the self-syncing property that makes the re-run caveat legitimate — the header's surface map at `:23` shows `.loom/scripts/ <- defaults/scripts/ (recursive)`, i.e. the running script overwrites itself mid-run. `verb_past` is now a proper `local` at `:285`, consistent with the anecdote being a stale-copy bug that a re-run resolves.

**Argument shape is correct.** The script takes no positional target (`:126-129` rejects any non-flag as an unknown argument and exits `1`). The PR correctly does *not* append `<this-repo>` to the invocation the way every other row does — an easy detail to get wrong by pattern-matching the neighbors.

**Keeping `git -C <source> pull --ff-only` ahead of the resync still matters** — the script reads from the source tree via `.loom/loom-source-path` / `install-metadata.json` `loom_source` (`:169-195`), so a stale clone yields a stale sync. Ordering preserved.

### Scope and non-generalization

Leaving the Anvil / kicad-tools rows alone is the right call: neither installer is checked out here, so generalizing would have been speculation, and the PR body records the follow-up rather than silently dropping it. The Repo Skills row being left alone is verified load-bearing, not assumed — this repo's `install.sh` has no reinstall-refusal guard; `confirm()` at `:48-51` is a TTY gate that `-y` bypasses cleanly, and the only other gates are the source-repo self-install refusal (`:171`) and a not-a-git-repo warning (`:177`). `install.sh -y <this-repo>` genuinely does work on an existing install here.

### Hygiene

- **Source tree only.** `.gitignore:7:.claude/`, 0 tracked files under `.claude/`. The diff touches `commands/repo/update-tools.md` and `CHANGELOG.md` — no installed copy.
- **CHANGELOG.** New `## Unreleased` sits directly under `# Changelog`, above `## 0.8.0 (2026-08-04)`. Shape matches the pre-0.8.0 house style exactly (bare `## Unreleased` + bolded lead sentence citing `(#119)` + substantive prose; the release-narrative lead paragraph is added at release time, correctly absent here).
- **Tests.** Re-ran `./hooks/repo/tests/run.sh` on the PR branch myself: **PASS: 1296  FAIL: 0**, matching baseline. The only `update-tools` match under the test trees is `hooks/repo/tests/test-install-sidecar-untracking.sh:17`, a comment about the source pointer the command reads — no prose pin, nothing this edit should have updated.
- **CI:** `pnpm test (guard/handoff/loss-check suite)` SUCCESS. `mergeStateStatus: CLEAN`, `mergeable: MERGEABLE`.

### Acceptance criteria (#119)

All five met: resync is the primary path with `--dry-run` first, reinstall is fenced separately and explicitly labeled destructive with a back-up reminder, the self-syncing re-run caveat is documented adjacent to the Loom row, and Safety Rule 2 now reads "installer **or update mechanism**" with a stated preference for a dedicated non-destructive updater — while retaining the never-hand-copy floor. Rules 1/3/4/5 are byte-identical with unchanged numbering.

### Non-blocking observations (no action required)

1. **Target selection is cwd-derived, not path-derived.** The script resolves the repo it acts on from `git rev-parse --git-common-dir` against the *current directory* (`:134-155`), not from its own location. So `<this-repo>/.loom/scripts/resync-installed.sh` is targeted by virtue of where the command runs, not by the path prefix. This is correct in practice — step 4's only other command uses `git -C <source>` precisely so cwd never leaves the target repo — and the `<this-repo>/` prefix is the clearest way to write it. Just noting the mechanism differs from the sibling rows, where the trailing `<this-repo>` argument really is what selects the target.
2. The follow-up to spot-check Anvil's `install-anvil.sh` and kicad-tools' `install-kct.sh` for the same guard pattern is worth filing so it does not evaporate with this PR.

Neither blocks. This is a well-researched documentation fix that replaces guidance which could only ever fail or escalate to a destructive path, and it documents the replacement accurately. Approved.

PR #132, opened 2026-08-05
feat(tidy): split KEEP into generated vs name-collision sub-cases (#120)

### LINKED ISSUE #120: /repo:tidy KEEP tier cannot express 'tracked and actively harmful' vs 'tracked but generated'
## Original Issue

`/repo:tidy` refuses to delete tracked files — correct, and safety rule 1 should not change. But the **KEEP** tier that catches them is purely informational, and its only documented pointer is `/repo:gitignore`:

> **KEEP** — flagged only as information: tracked files that look like they don't belong (build output that got committed — point to [[gitignore]]).

That framing assumes the remedy is always "this should have been ignored." There is a second case it cannot express: **a tracked file that is actively harmful and should be deleted outright.**

Motivating case from a real run. A repo had 11 tracked schematic backups; 3 of them were named with a live extension:

```
connectors_backup_20260427_163100.kicad_sch
connectors_backup_20260427_212114.kicad_sch
dac_backup_20260427_212043.kicad_sch
```

Because the extension is `.kicad_sch`, KiCad and any project-walking tool parse them as **real schematic sheets** — a scan double-counted one backup's defects as if it were a separate sheet. The other 8 used `foo.kicad_sch.backup-<timestamp>` and were inert.

Tidy did the right thing (deleted nothing), but the report gave the operator no way to tell the 3 dangerous files from the 8 harmless ones, and `/repo:gitignore` is the wrong destination: gitignoring them would not undo the fact that they are already tracked and being parsed. The correct remedy was a deliberate `git rm` commit, which no `/repo:*` command currently owns.

- [ ] KEEP distinguishes "tracked but generated" (-> `/repo:gitignore`) from "tracked and actively harmful / shadowing real files" (-> deliberate removal)
- [ ] The harmful case points somewhere actionable rather than at `/repo:gitignore`
- [ ] A tracked file whose name collides with a live source extension (`*backup*.<source-ext>`, `*copy*.<source-ext>`, `*.orig.<source-ext>`) is surfaced as its own signal
- [ ] Safety rule 1 is unchanged: tidy still never deletes tracked files, it only reports them better

---

### Problem Statement

`/repo:tidy`'s KEEP tier (`commands/repo/tidy.md`, step 2 "Categorize") currently has a single sentence of guidance for *every* tracked file that "looks like it doesn't belong":

```
- **KEEP** — flagged only as information: tracked files that look like they
  don't belong (build output that got committed — point to [[gitignore]]).
```

(origin/main `commands/repo/tidy.md:275-276`, and echoed in the sample report at `commands/repo/tidy.md:309-310`: `assets/build.min.js      tracked but looks generated — see /repo:gitignore`.)

This conflates two remedies that require opposite actions from the operator:

1. **Tracked but generated** — a build artifact that got committed by accident. The fix is to stop tracking it and gitignore it going forward (`/repo:gitignore`'s job — see `commands/repo/gitignore.md`, section "Under-Ignored Files": "Find tracked files that are probably build artifacts").
2. **Tracked and actively harmful** — a tracked file whose *name* collides with a live source-file extension, so tools that walk the project by extension (compilers, EDA tools, linters, IDEs) parse it as real source. Gitignoring it does nothing: the file is already tracked, already on disk, and already being parsed. The only real fix is a deliberate `git rm` (and it must be deliberate — safety rule 1 (`commands/repo/tidy.md:373`, "Never delete tracked files — that's a git operation the user does deliberately") is explicit that tidy itself must never perform this).

Confirmed during curation: no existing `/repo:*` command owns "delete this specific tracked file." `/repo:gitignore` is audit-only (adds/removes gitignore rules, never runs `git rm`). `/repo:reset` (worktree/branch/stash pruning) is a different object class entirely — it doesn't touch tracked files in the working tree. So the "harmful" case currently has no `/repo:*` destination at all, which is exactly the gap the issue reports.

### Decision: where "actively harmful" should point

Two options were on the table:

- **Option A — documented manual `git rm` recipe printed in the KEEP output.** Tidy prints the exact `git rm` command(s) for the flagged file(s) as text in the report, but never executes them. The operator copies/runs it (or edits it) themselves.
- **Option B — just note it for the operator** with prose pointing at "run `git rm` yourself," no concrete command shown.

**Decision: Option A.** A printed, copy-pasteable `git rm` command is strictly more actionable than a prose note, costs nothing extra to implement (tidy already knows the exact path), and does not require standing up a new `/repo:*` command for what is fundamentally a one-line git operation. It also mirrors the existing pattern for the "generated" sub-case, which already prints "`see /repo:gitignore`" as a pointer rather than making the operator guess. Printing the command is not the same as running it — safety rule 1 is unaffected because tidy never invokes `git rm` itself, it only emits the string in the report, exactly as it already emits the `/repo:gitignore` pointer.

Do **not** create a new `/repo:*` command for this — that's a heavier deliverable than the issue calls for and out of scope here.

### Implementation Guidance

In `commands/repo/tidy.md`, step 2 "Categorize" (around line 275, immediately after the ASK subsection and before "### 3. Report"):

1. **Split KEEP into two named sub-cases** instead of one collapsed bullet:
   - `KEEP (generated)` — tracked file that looks like build output / regenerable artifact. Pointer: `/repo:gitignore`. This preserves the current behavior verbatim for the existing case (`assets/build.min.js` example).
   - `KEEP (name collision)` — tracked file whose name matches a live source-file naming pattern for this repo (`*backup*.<source-ext>`, `*copy*.<source-ext>`, `*.orig.<source-ext>`, where `<source-ext>` is drawn from extensions of other **tracked** files in the same directory/project, not a hardcoded global list — a `.kicad_sch` collision only matters in a repo that has real `.kicad_sch` files). Pointer: a literal `git rm <path>` line, framed as "run this deliberately, tidy will not run it for you," plus a one-line reason ("this extension is parsed as a real source file by \<tool class\> and may be double-counted/loaded").
   - Detection for the collision sub-case is a **naming heuristic**, not a content check: no need to parse the file to know `*backup*.kicad_sch` is suspicious when `.kicad_sch` is also a live extension in the tree. Base it on `git ls-files` (tracked files only) grouped by extension, then flag tracked files matching `*backup*`, `*copy*`, or `*.orig.` immediately before an extension that also appears on other tracked files that are *not* similarly named — i.e., the extension has "real" siblings.
2. **Update the sample report block** (step 3 "Report", around line 309) to show both sub-cases so the format is unambiguous for implementers, e.g.:
   ```
   KEEP (informational):
     assets/build.min.js         tracked but looks generated — see /repo:gitignore
     connectors_backup_20260427_163100.kicad_sch  tracked, name collides with live .kicad_sch source files
       — actively harmful (parsed as a real schematic); not something tidy will delete —
       run deliberately: git rm connectors_backup_20260427_163100.kicad_sch
   ```
3. **Do not touch Safety Rules** — rule 1 (`commands/repo/tidy.md:373`) stays exactly as written. Do not add a "### Verify after write" step or any execution path for `git rm`; `commands/repo/tests/test-verify-fix-persistence.sh` (lines ~321-330) has a scope guard asserting `tidy.md` contains **no** `### Verify after write` heading, specifically because tidy never writes/reverts tracked-file state. Adding a printed (non-executed) recipe does not trip this guard, but adding any executed `git rm` step would need that test's scope-guard comment revisited — out of scope for this issue.
4. **Naming pattern precision**: `*backup*.<ext>`, `*copy*.<ext>`, `*.orig.<ext>` should be matched case-insensitively and should not misfire on genuinely different extensions (`foo.kicad_sch.backup-20260427` — the *inert* 8-file case from the issue — must NOT trigger this, since `.backup-<timestamp>` is not itself a live extension; only `foo_backup.kicad_sch`-shaped names where the live extension is the trailing/actual extension should trigger).

### Acceptance Criteria (refined)

- [ ] `commands/repo/tidy.md` step 2 "Categorize" splits KEEP into two documented sub-cases: **generated** (pointer: `/repo:gitignore`, existing behavior preserved) and **name collision / actively harmful** (pointer: a literal `git rm <path>` recipe printed in the report, never executed by tidy)
- [ ] The name-collision sub-case is defined by pattern (`*backup*.<source-ext>`, `*copy*.<source-ext>`, `*.orig.<source-ext>`) where `<source-ext>` is an extension shared by other tracked, non-similarly-named files in the repo — not a hardcoded global extension list
- [ ] The pattern does not false-positive on the inert case where the "backup" marker is itself the trailing extension (e.g. `foo.kicad_sch.backup-20260427`, which is not a `.kicad_sch` file by extension)
- [ ] Step 3 "Report" sample output shows both KEEP sub-cases with distinct formatting so the two remedies are visually distinguishable
- [ ] Safety Rule 1 (`commands/repo/tidy.md:373`, "Never delete tracked files") is unchanged verbatim; no new execution path is added for `git rm` — tidy only prints the command
- [ ] `commands/repo/tests/test-verify-fix-persistence.sh`'s tidy/audit scope guard (no `### Verify after write` heading in `tidy.md`) still passes unmodified

### Affected Files

- `commands/repo/tidy.md` - Split the KEEP tier definition (step 2, "Categorize", ~line 275) into "generated" and "name collision / actively harmful" sub-cases; add the `git rm` recipe pointer for the harmful sub-case; update the sample report block (step 3, "Report", ~line 309) to illustrate both sub-cases. No changes to Safety Rules (step-4-adjacent list, starting ~line 371) — rule 1 stays verbatim.

Note: a local `.claude/commands/repo/tidy.md` symlink also exists in some working trees pointing at the file above, but `.claude/` is gitignored (`.gitignore:7`) and not part of `origin/main` — it is local tooling scaffolding, not a repo path, and needs no separate edit.

### Test Plan

- [ ] Manual verification: In a scratch repo with tracked `.kicad_sch` (or any live-extension) files plus a tracked `*backup*.<ext>` collision file and a tracked `*.<ext>.backup-<timestamp>` inert file, run `/repo:tidy` and confirm the report places the collision file under `KEEP (name collision)` with a printed `git rm` line, and does **not** flag the inert `.backup-<timestamp>` file the same way.
- [ ] Manual verification: A plain generated/committed build artifact (e.g. `assets/build.min.js`-style case) still reports under `KEEP (generated)` pointing at `/repo:gitignore`, unchanged from current behavior.
- [ ] Regression: Confirm tidy does not execute `git rm` anywhere in the flow — grep the updated `tidy.md` for `git rm` and confirm every occurrence is inside a report-string/example block, not a command that gets shelled out.
- [ ] Automated: `commands/repo/tests/test-verify-fix-persistence.sh` still passes (scope guard: `tidy.md` has no `### Verify after write` heading).
- [ ] Automated: `commands/repo/tests/test-early-sync-switch.sh` still passes (stage-ordering assertions against `all.md` are unaffected by this change, but worth a spot check since it references the Tidy stage heading).
- [ ] Edge case: repo with no live source extensions colliding at all — KEEP (name collision) sub-case simply never fires; only KEEP (generated) can appear. Verify the report doesn't print an empty "name collision" sub-header when there's nothing to show.

### Notes for the Builder

- Line numbers above are from `origin/main` as of PR #124 (`a6f8794`, "fix(tidy): prune .git by -name so nested .git directories are skipped") — the most recent merge touching `tidy.md`'s find walks. Re-verify line numbers against your worktree's `origin/main` before editing, since curation and build may be separated in time.
- This is a documentation-only change to a `/repo:*` command's markdown spec — no application code changes. Affected file count is 1 real file (plus its symlink mirror).

rjwalters:
Closes #120

## Problem

`/repo:tidy`'s **KEEP** tier had one bullet for every tracked file that "looks like it doesn't belong," and its only pointer was `/repo:gitignore`. That framing assumes the remedy is always "this should have been ignored." It cannot express the second case: a tracked file whose **trailing extension is a live source extension in the repo**, so every tool that walks the project by extension (compilers, EDA tools, linters, IDEs) opens it as real source. Gitignoring it does nothing — the file is already tracked, already on disk, and already being parsed.

The motivating run reported 11 tracked KiCad backups as one undifferentiated list. Three were `<stem>_backup_<timestamp>.kicad_sch`, which KiCad loaded as real schematic sheets (a defect scan counted one backup's contents twice). The other eight were `<stem>.kicad_sch.backup-<timestamp>` and were inert. The report gave the operator no way to tell them apart.

## Change

`commands/repo/tidy.md` step 2 now defines **two named sub-cases** instead of one collapsed bullet:

- **KEEP (generated)** — a committed build artifact. Inert; the fix is to stop tracking it. Pointer stays `/repo:gitignore`, and the `assets/build.min.js` sample line is unchanged.
- **KEEP (name collision)** — reported with a literal, copy-pasteable `git rm <path>` line plus a one-line reason naming the tool class that parses it.

Detection is a **naming heuristic, not a content check** (nothing is opened), run over `git ls-files` so untracked/gitignored files stay in their own tiers. A collision requires **both**:

1. A `backup` / `copy` / `.orig` marker in the **stem**, before the final `.`, matched case-insensitively.
2. A trailing extension with **real siblings** — at least one *other* tracked file ending in the same `.<ext>` and carrying no such marker.

Condition 2 derives the extension set **per-repo, never from a hardcoded global list**: the same filename in a repo with no real `.kicad_sch` files is just a file. Condition 1 is what keeps the inert eight out — in `connectors.kicad_sch.backup-20260427_163100` the marker *is* the trailing extension, so nothing will ever parse it as a schematic.

Each sub-header prints **only when it has entries**, so a repo with no collisions emits the single `generated` block byte-for-byte as before rather than an empty heading that reads as a finding.

Step 3's sample report shows both sub-blocks with deliberately distinct formatting (the collision block gets its own indented `why:` line and the recipe on a line of its own).

## Safety

- **Safety rule 1 is unchanged verbatim.** Tidy still never deletes a tracked file.
- **No execution path was added.** Tidy *prints* the `git rm` string exactly as it already prints the `see /repo:gitignore` pointer. Step 4 now says explicitly that KEEP is never part of the apply step under any flag — not `--ask`, not `--apply`.
- **No `### Verify after write` heading** was added to `tidy.md`, so the tidy/audit scope guard in `commands/repo/tests/test-verify-fix-persistence.sh` is unmodified and still passes.
- **No new `/repo:*` command.** A one-line git operation does not need one.

## Test Plan

New suite `commands/repo/tests/test-tidy-keep-tiers.sh` (77 cases, wired into `hooks/repo/tests/run.sh` / `pnpm test`):

- **Fixture section** — transcribes the heuristic as `keep_collision()` and runs it against real git repos, including the motivating fixture (3 live-extension backups + 8 `*.kicad_sch.backup-<ts>` + 4 real sheets): exactly the 3 are flagged, none of the 8 are, and no unmarked file is.
- **Derived-extension pins** — no real sibling ⇒ not a collision; adding one real sibling flips the *same untouched file* to a collision; two marker files are not each other's siblings; an arbitrary repo-local extension (`.wibble`) works with no global table.
- **Marker pins** — `BACKUP` / `copy` / `Copy` / `.orig.` all match case-insensitively; a bare `*.orig` leftover is *not* a collision; a `backups/` parent directory does not mark its contents.
- **Scope** — untracked and gitignored collision-shaped files are not KEEP.
- **Report** — both sub-blocks render distinctly; empty-suppression asserted in both directions; an empty KEEP tier prints nothing.
- **Printed, never executed** — after rendering a real repo's report, the flagged file is still on disk, still tracked, and `git status --porcelain` is empty.
- **Fence-level guard** — every ` ```bash ` block in `tidy.md` is scanned and must contain no `git rm`, while the sample-output fence must still contain the recipe. This pins the printed-vs-executed distinction by mechanism rather than by a comment.
- **Doc-drift guards** — 33 assertions on `tidy.md` prose, including an `assert_not_contains` that fails if the old collapsed one-bullet KEEP definition returns, a verbatim check on safety rule 1, and the mirrored no-`### Verify after write` guard.

Negative control: reverting `tidy.md` to its pre-change state fails **33** of the new assertions, so the guards bite.

```
$ ./hooks/repo/tests/run.sh
  ok   test-verify-fix-persistence.sh                       -> 79 cases pass   (unmodified)
  ok   test-early-sync-switch.sh                            -> 84 cases pass
  ok   test-tidy-keep-tiers.sh                              -> 77 cases pass
===============================
PASS: 1373   FAIL: 0
```

Baseline on a clean checkout of `origin/main` (5d2e9d7) was **1296 pass / 0 fail**; 1296 → 1373 with the new suite, 0 failures.

https://claude.ai/code/session_01AHBrUgf4EmjEYkbEQdgZYA

---

## Update after review (Doctor pass, commit `1c38958`)

The description above reflects the **two**-condition version as originally opened. The
merged behaviour has **three** conditions — condition 3 (base sibling) was added after
review found that conditions 1+2 alone flagged 8 ordinary source files (`src/backup.py`,
`src/copy.py`, `src/copyright.py`, `src/copy_utils.ts`, `src/BackupManager.ts`,
`src/useCopyToClipboard.ts`, `src/deepcopy_helpers.py`, `docs/backup-strategy.md`) and
printed a `git rm` for each.

Corrected numbers:

| Claim above | Actual |
|---|---|
| 2 conditions | **3** conditions (marker in stem / live extension with real siblings / base sibling, same dir) |
| `test-tidy-keep-tiers.sh` 77 cases | **112** cases |
| Suite total 1373 | **1408** pass / 0 fail |
| Negative control 33 assertions | **51** assertions fail when `commands/repo/tidy.md` alone is reverted |

Also added in that pass: an extension that itself carries a marker is never a live source
extension (resolves the `.orig` / SAFE-tier overlap), and the `why:` line may only be
written from what conditions 2 and 3 established — if it cannot be written truthfully the
file is not reported. `CHANGELOG.md` describes the final three-condition behaviour.


rjwalters:
## Changes Requested

One blocking finding in the detection heuristic. Everything else in this PR verified clean — the hard constraints all hold, CI passes, merge state is CLEAN, and the doc prose is genuinely good. The block is narrow and the fix is small.

---

### Verification performed (all pass)

| Constraint | Result |
|---|---|
| Safety rule 1 byte-identical | ✅ `commands/repo/tidy.md:464` is byte-identical to `origin/main:373`. Diffed the entire numbered safety-rules region (`main` 360-420 vs PR 451-511) — identical at a uniform +91 offset. |
| No `### Verify after write` in `tidy.md` | ✅ 0 occurrences. `test-verify-fix-persistence.sh` unmodified, 79/79 pass. |
| Tidy never *executes* `git rm` | ✅ Fence guard is real. I tried three ways to fool it and all three were caught: (a) top-level ` ```bash ` fence with `git rm`, (b) a 4-space-**indented** ` ```bash ` fence nested in a list item, (c) `git rm` injected directly into the existing indented `git ls-files` fence at `tidy.md:299`. All produced `FAIL: no ```bash fence in tidy.md contains 'git rm'`. The `sub(/^[ \t]+/,"",t)` in `fence_lines()` (`test-tidy-keep-tiers.sh:483`) handles indentation correctly, and all 10 fence markers in `tidy.md` are balanced (5 blocks), so the walker's state never desyncs. |
| No new `/repo:*` command | ✅ |
| Only `commands/` edited | ✅ `git diff --name-only origin/main` = `CHANGELOG.md`, `commands/repo/tests/test-tidy-keep-tiers.sh`, `commands/repo/tidy.md`, `hooks/repo/tests/run.sh`. |
| Suite wired + counts real | ✅ Ran `./hooks/repo/tests/run.sh` myself: **PASS: 1373 FAIL: 0**, `test-tidy-keep-tiers.sh -> 77 cases pass`. Matches the claim exactly. |
| Negative control meaningful | ✅ Reverting only `tidy.md` fails **32** assertions (PR body says 33 — trivial off-by-one, see minor notes). I listed all 32: every one pins substantive prose (`whose remedies are **opposite**`, `Never match against a hardcoded global extension list`, `does **not** collide`, the sample-report sub-headers, …). None are tautological. |
| CHANGELOG | ✅ Exactly one `## Unreleased` heading (`CHANGELOG.md:3`), entry added under it, house style matches (bolded lead citing `(#120)`, then substantive prose). |
| CI / merge state | ✅ `pnpm test` pass (1m20s); `mergeStateStatus` = `CLEAN`. |

I re-derived the heuristic **independently in Python from the prose in `tidy.md` step 2** rather than reusing the suite's awk, precisely because the test and the doc were written together. Against a repo I constructed myself:

```
connectors_backup_20260427_163100.kicad_sch   -> FLAGGED   (correct, AC #3)
foo.kicad_sch.backup-20260427                 -> not flagged (correct, AC #3)
connectors.kicad_sch.backup-20260427_163100   -> not flagged (correct)
```

AC #2 also holds under the independent implementation: the extension set is derived from `git ls-files` per-repo, and removing the real sibling flips the same untouched file out of the set. Case-insensitivity and empty-sub-header suppression both confirmed.

---

### 🔴 Blocking: condition 1 flags ordinary source files, and the printed remedy is `git rm`

The two conditions in `commands/repo/tidy.md:300-312` are stated as a complete mechanical rule ("A tracked file is a name collision when **both** conditions hold"), and condition 1 is a bare substring test — `backup` or `copy` anywhere in the stem. That matches a *topic* just as readily as a *provenance stamp*.

Running my independent implementation against a synthetic but entirely ordinary repo (12 tracked files, no backups anywhere in it):

```
src/backup.py                 -> FLAGGED
src/copy.py                   -> FLAGGED
src/copyright.py              -> FLAGGED
src/copy_utils.ts             -> FLAGGED
src/BackupManager.ts          -> FLAGGED
src/useCopyToClipboard.ts     -> FLAGGED
src/deepcopy_helpers.py       -> FLAGGED
docs/backup-strategy.md       -> FLAGGED
```

8 of 12 files. Every one gets the `name collision — tracked AND parsed as real source; gitignoring fixes nothing:` sub-block and a copy-pasteable `git rm src/backup.py`.

Why I'm treating this as blocking rather than a note, despite the change being print-only:

1. **The `why:` line becomes a false statement.** The sample at `tidy.md:378-380` templates "N real `.kicad_sch` files in this tree, so KiCad opens this backup as a schematic sheet and its contents are counted twice." Rendered for `src/BackupManager.ts` that reads "…so the TypeScript compiler opens this backup as a module and its contents are counted twice" — which describes the file working as designed. The report doesn't just misfile it; it argues for its deletion with a fabricated rationale.
2. **The framing amplifies it.** `tidy.md:342-344` says an empty sub-header "reads as a finding, and this one is meant to read as an alarm." Print-only is the right safety posture, but it isn't a defense against an alarm that is wrong most of the time — the whole point of the split (per the issue) was to *stop burying the signal*, and an 8-of-12 false-positive rate re-buries it, just in the other tier.
3. **`git rm` is the one recipe where a wrong flag is unrecoverable-ish.** A `see /repo:gitignore` pointer that misfires costs nothing. A `git rm src/backup.py` that misfires and gets pasted costs a source file.

Note this repo and `rjwalters/loom` both flag **0** files, so the suite passing here doesn't exercise the failure mode at all — the fixtures only contain names where the marker *is* a provenance stamp.

**Suggested narrowing** (not prescriptive — pick whatever you like, but it should be a third condition, not a bigger word list): require the marker to read as provenance by demanding a **base sibling** — stripping the marker run (and its adjacent separator/timestamp) from the stem must yield another tracked file with the same extension.

- `connectors_backup_20260427_163100.kicad_sch` → `connectors.kicad_sch` ✅ exists — still flagged (motivating case preserved)
- `schematic copy.kicad_sch` → `schematic.kicad_sch` ✅ exists — still flagged
- `board_backup_20260101.wibble` → `board.wibble` ✅ exists — derived-extension case preserved
- `src/backup.py` → strips to nothing; `copyright.py` → no `copyri.py`; `copy_utils.ts` → no `_utils.ts`; `BackupManager.ts` → no `Manager.ts` — **all 8 false positives drop out**

This is consistent with the curated ACs: AC #2 constrains where `<source-ext>` comes from and forbids a hardcoded extension table, and adding a third name-shape condition doesn't touch either. It's also squarely in the spirit of the curator's "Naming pattern precision" item, which asked the pattern not to misfire. Two synthetic fixtures would need base siblings added (`SHEET_Copy.kicad_sch`, `Connectors_BACKUP_20260427.kicad_sch` in the `markers` repo currently have no matching base file) — that's a fixture edit, not a contract change.

If you'd rather keep the rule as-is, the alternative I'd accept is an explicit **judgment clause** in the sub-case: state plainly that a marker word appearing as the file's *subject* (`backup.py`, `copy_utils.ts`, `BackupManager.ts`) is not a collision, that the marker must read as a provenance stamp on an existing file, and that the `why:` line must be omitted (and the file not reported) when that can't be asserted truthfully. That's weaker — it relies on the reading agent — but it at least stops the doc from mandating the false positive.

---

### Minor (non-blocking)

1. **`commands/repo/tests/test-tidy-keep-tiers.sh:138`** — the transcription implements the `.orig` marker as `stem ~ /(^|\.)orig$/`, but `tidy.md:307` documents the pattern as `*.orig.`. These diverge in two directions: `orig.rs` (stem `orig`) is matched by the code but not by the documented pattern, and `foo.orig.bar.rs` (stem `foo.orig.bar`) matches the documented pattern but not the code. Cosmetic today; worth aligning so the transcription stays a faithful mirror.

2. **`.orig` can become a derived "live source extension."** In a repo with unmarked merge leftovers `a.rs.orig` / `b.rs.orig`, `orig` acquires real siblings, so `sheet_backup_20260101.orig` gets flagged as a name collision (verified). Nothing parses `.orig` as source, and `test-tidy-keep-tiers.sh:300-304` explicitly assigns `*.orig` to the SAFE tier's merge-leftover rule — so this is the two rules overlapping. Low priority.

3. **PR body says "reverting `tidy.md` fails **33** of the new assertions"** — measured value is 32. Worth correcting in the body since the number is cited as evidence.

Everything else here is strong work — the fence-level guard in particular is the right way to pin printed-vs-executed, and it survived every attempt I made to break it. Fix condition 1's breadth and I expect a fast re-approval.

rjwalters:
## Feedback addressed — condition 3 (base sibling) added

Adopted the suggested narrowing as-is. Thanks for the independent re-derivation — it found a real hole, and I reproduced your exact 8-of-12 result before fixing it.

### The blocking finding

`commands/repo/tidy.md` now states **three** conditions, not two. Condition 3:

> The marker reads as a **provenance stamp on an existing file**, not as the file's subject. Strip the **marker run** off the *end* of the stem — the marker word, the separator run (space, `_`, `-`, `.`) in front of it, and any timestamp or copy index (digits and separators) behind it — and what remains must be a non-empty stem that names a **base sibling**: another tracked file `<base>.<ext>`, same extension, **same directory**.

Two deliberate refinements on top of your sketch, both to make the rule harder to satisfy by accident:

1. **The marker run must be at the *end* of the stem, with a separator in front of it.** This does the stamp/topic discrimination structurally rather than by word list. `copyright`, `backupmanager`, `deepcopy_helpers`, `usecopytoclipboard` have no marker *run* at all — the marker is glued into a longer word — so they never reach the sibling lookup. `copy_utils` has the marker at the front with no base in front of it. `backup.py` / `copy.py` strip to nothing. This matters: a strip that only deletes the marker *word* would turn `copy_utils.ts` into `utils.ts` and `BackupManager.ts` into `Manager.ts`, both of which are perfectly plausible files to have in the same tree.
2. **Same directory.** A same-named file elsewhere in the tree is a coincidence, and coincidences are exactly what condition 3 exists to reject.

### Verification that the false positives are gone

I rebuilt your synthetic repo (the 8 flagged names plus unmarked companions) and ran the **old** transcription against it first, to confirm I was fixing the thing you found:

```
$ # old heuristic, PR as reviewed
docs/backup-strategy.md   src/BackupManager.ts   src/backup.py   src/copy.py
src/copy_utils.ts         src/copyright.py       src/deepcopy_helpers.py
src/useCopyToClipboard.ts
```

All 8, matching your list exactly. With condition 3 the same repo yields **nothing**. `rjwalters/repo` itself and `rjwalters/loom` also both yield 0, unchanged.

That repo is now a fixture (`-- condition 3: the marker must be a provenance STAMP, not a topic --`), with each of the 8 pinned by name. Its unmarked companions are **adversarial on purpose**: `src/utils.ts`, `src/Manager.ts`, `src/helpers.py` and `docs/strategy.md` are precisely the base siblings a naive "delete the marker word" strip would find for `copy_utils.ts`, `BackupManager.ts`, `deepcopy_helpers.py` and `backup-strategy.md`. So the fixture fails a weaker implementation of the same idea, not just the old one. The same repo then gets `src/main_backup_20260101.py` added and asserts it is the **only** hit — the rule discriminates within one tree rather than by refusing to fire.

Preserved: all three motivating `.kicad_sch` backups, the 8 inert `.kicad_sch.backup-<ts>`, the derived-`.wibble` case, the derived-extension flip test, tracked-only scope, the `backups/` parent-directory case. Two fixtures gained base siblings as you predicted (`Connectors.kicad_sch`, `SHEET.kicad_sch` in the `markers` repo) — fixture edit only.

### On the case you asked me to evaluate critically

You asked whether this wrongly drops a genuinely harmful file. **It does, and I did not paper over it.** A real backup whose base file was since renamed or deleted, or that was moved into a `backups/` directory away from its original, has no base sibling and is not flagged even though a tool still parses it. I kept the narrowing anyway and documented the trade explicitly (`The trade is deliberate — precision bought with recall`), with two mitigations:

- Such a file is **not lost** — it stays in the `generated` sub-case with the `/repo:gitignore` pointer, exactly where the inert shapes go. It loses the alarm and the recipe, not its visibility.
- The recall loss lands exactly where the report would have had to lie anyway. Which is why I also added your **alternative** on top of the base-sibling rule rather than instead of it: the `why:` line may only be written from what conditions 2 and 3 established (the sibling count, and the *named* base file), and **if that sentence cannot be written truthfully the file is not a collision and must not be reported here**. The sample `why:` now leads with `backup of connectors.kicad_sch`, so the false statement you quoted is unwritable by construction — there is no template slot left to fill with a fabricated rationale.

There is one deliberate false negative I did not chase: `foo_backup_final.kicad_sch` (a trailing word rather than a timestamp) does not strip. Widening the tail to arbitrary words reopens the `data_copy_helpers.py` class of match, so digits-and-separators it is.

### Minor notes

1. **Doc/transcription divergence on `.orig`** — fixed by making both simpler and identical. The doc now says the marker is the substring `backup`, `copy`, or `orig`, case-insensitively, and the transcription is `index(stem, "orig")`. The `(^|\.)orig$` special case is gone entirely: condition 3 already supplies the delimiting (`original.rs` strips to nothing), so condition 1 no longer needs a second, differently-shaped rule for one of its three markers.
2. **`.orig` as a derived live extension** — resolved in the SAFE tier's favour, by a general rule rather than a special case: **an extension that itself carries a marker is never live**, however many files share it. `.orig` and `.backup-20260427_163100` are provenance suffixes; nothing parses them as source. This also makes the inert-shape exclusion explicit instead of emergent from the sibling count. New fixture `origlive` builds your exact scenario (`a.rs.orig`, `b.rs.orig` making `orig` look live) and asserts `sheet_backup_20260101.orig` is not flagged.
3. **PR body 33 vs 32** — corrected. The current negative control is **51** assertions failing when `commands/repo/tidy.md` alone is reverted to `origin/main` (112 total, 61 pass).

### Constraints re-checked

- Safety rule 1 byte-identical — the whole numbered safety-rules region diffs clean against `origin/main` at a uniform +136 offset.
- 0 occurrences of `### Verify after write`; `test-verify-fix-persistence.sh` unmodified, 79/79.
- Fence guard passes: no ```` ```bash ```` fence contains `git rm`; the sample-output fence still carries the recipe. No execution path added.
- No new `/repo:*` command (`git ls-tree commands/repo/` identical to main); `.claude/` untouched; changed files are `CHANGELOG.md`, `commands/repo/tidy.md`, `commands/repo/tests/test-tidy-keep-tiers.sh` (+ `hooks/repo/tests/run.sh` wiring from the original commit).
- CHANGELOG entry under the existing single `## Unreleased` updated — the mechanism description changed materially, so it now documents all three conditions, the marker-extension exclusion, the `why:`-line rule, and the recall trade.

**`./hooks/repo/tests/run.sh`: PASS 1408, FAIL 0** (`test-tidy-keep-tiers.sh` 77 → 112 cases). `pnpm check:ci` green.

rjwalters:
## ✅ Approved — second (final) Judge pass

The blocking finding from the first pass is fully resolved. I re-derived the heuristic independently from the prose again — this time from the **three**-condition text at `commands/repo/tidy.md:303-331`, in Python, without reading the suite's awk first — and every claim in the Doctor's account checks out. Approving.

---

### 1. Independent re-derivation vs. the 8 false positives

My implementation of `commands/repo/tidy.md:303-331`: cond 1 = marker substring in the stem (case-insensitive); cond 2 = trailing ext has ≥1 other tracked unmarked file, **and** the ext must not itself carry a marker; cond 3 = strip `[sep-run](backup|copy|orig)[digits/sep]*$` off the stem, remainder non-empty, `<dir>/<remainder>.<ext>` tracked.

Against a rebuilt 15-file ordinary repo containing all 8 names **plus** the adversarial companions (`src/utils.ts`, `src/Manager.ts`, `src/helpers.py`, `docs/strategy.md`):

```
(no output — 0 of 15 flagged)
```

All 8 unflagged. One derivation note that only strengthens the doc: my first attempt anchored the strip as `^(.*\S)[sep]+...` and produced `sheet -` rather than `sheet` for `sheet - Copy 2.kicad_sch`. `commands/repo/tidy.md:323` says "the separator **run** … in front of it", which forces the base to not end in a separator — the prose is precise enough to catch my own bug, and the awk's leftmost-longest `sub(/[ ._-]+(backup|copy|orig)[0-9 ._-]*$/…)` at `test-tidy-keep-tiers.sh:149` gets it right by construction.

### 2. Motivating cases still fire

```
FLAG connectors_backup_20260427_163100.kicad_sch  base=connectors.kicad_sch  siblings=6
FLAG schematic copy.kicad_sch                     base=schematic.kicad_sch   siblings=6
FLAG sheet - Copy 2.kicad_sch                     base=sheet.kicad_sch       siblings=6
FLAG parser.orig.rs                               base=parser.rs             siblings=2
```

All four documented shapes at `tidy.md:327-329` fire under my implementation; the 8 inert `connectors.kicad_sch.backup-<ts>` fire on none (they fail cond 1 *and* cond 2 independently — marker is in the ext, not the stem). Suite pins the same at `test-tidy-keep-tiers.sh:260-283`.

### 3. The tightenings are load-bearing, and the adversarial fixtures really bite

Verified the claim at `test-tidy-keep-tiers.sh:375-378` rather than taking it. I implemented two naive strips against the same fixture repo:

| Naive variant | False positives it produces |
|---|---|
| strip marker word + adjacent separators | `docs/backup-strategy.md → docs/strategy.md`, `src/BackupManager.ts → src/Manager.ts`, `src/copy_utils.ts → src/utils.ts` (**3**) |
| keep the tail after the marker | the above **plus** `src/deepcopy_helpers.py → src/helpers.py` (**4**) |

So the companions are genuinely adversarial: a weaker implementation of the *same* idea fails this fixture, not just the pre-fix one. The "marker run at the end of the stem, with a separator in front" tightening is what rejects all four, and `src/main_backup_20260101.py` being the sole hit after it is added (`:396-400`) proves the rule discriminates within one tree rather than by refusing to fire.

### 4. The documented false negative — acceptable, and stated honestly

`tidy.md:358-366` states it plainly ("The trade is deliberate — precision bought with recall"), names both loss shapes (base renamed/deleted; backup moved into `backups/`), and states the mitigation: the file "stays in the **generated** sub-case with the [[gitignore]] pointer" — visibility retained, alarm and recipe withheld. I accept the trade: the withheld output is exactly the output that could not be written truthfully, and the asymmetry of costs is right — a missed alarm costs a stale file, a false alarm costs a source file. `test-tidy-keep-tiers.sh:422-432` pins the false negative in both directions so it stays deliberate rather than drifting into a surprise.

### 5. The `why:` honesty clause holds

`tidy.md:370-375` binds the reason to conditions 2 and 3 only ("the sibling count … and the base sibling the file is a copy of"), with **"If that sentence cannot be written truthfully, the file is not a collision and must not be reported here"** and "never assert that a file is a backup of something tidy could not name". The sample at `tidy.md:425` now leads `why: backup of connectors.kicad_sch` — the template's first slot is the *named* base file, which only condition 3 can supply. The false rendering I quoted last time ("…so the TypeScript compiler opens this backup as a module and its contents are counted twice" for `BackupManager.ts`) is now unwritable twice over: that file can never reach the sub-case, and the template has no slot that could be filled without naming a base. Pinned at `test-tidy-keep-tiers.sh:595-602`.

### 6. `.orig` / SAFE-tier overlap — general rule, no surprising consequence

Stated in prose at `tidy.md:315-320` (**"An extension that itself carries a marker is never live"**, with the `*.orig` leftover explicitly assigned to SAFE), transcribed as `!marked(ext[i])` at `test-tidy-keep-tiers.sh:178`, and exercised by the `origlive` fixture (`:352-355`) that builds my exact scenario. I looked for a surprising consequence and found none material: no real source extension contains `backup`/`copy`/`orig` as a substring, so the rule only ever excludes provenance suffixes. The one theoretical cost is a repo using an extension like `.origami`, which is not a case worth designing for.

### 7. Doc/test lockstep restored

The `(^|\.)orig$` vs `*.orig.` divergence is gone. `tidy.md:306` says "the substring `backup`, `copy`, or `orig`"; `test-tidy-keep-tiers.sh:138` is `index(s, "orig")`. Identical rules, and simpler on both sides — condition 3 now supplies the delimiting that the special case used to.

### 8. Test counts and negative control — reproduced

```
$ ./hooks/repo/tests/run.sh
  ok   test-tidy-keep-tiers.sh  -> 112 cases pass
  ok   test-verify-fix-persistence.sh -> 79 cases pass   (unmodified)
PASS: 1408   FAIL: 0
```

Negative control: reverting **only** `commands/repo/tidy.md` to `origin/main` → **61 pass / 51 fail** of 112. Matches the Doctor's claim exactly; the guards bite hard, and I spot-checked that the failures pin substantive prose (the three conditions, the marker-extension exclusion, the `why:` clause, the recall trade), not tautologies. `pnpm check:ci` green locally; `shellcheck -S warning` clean on the new suite.

### 9. First-pass constraints re-verified on the new head

| Constraint | Result |
|---|---|
| Safety rule 1 byte-identical | ✅ `commands/repo/tidy.md:509` identical to `origin/main:373`; diffed the entire `## Safety Rules` section through EOF — **39 lines, zero differences** |
| No `### Verify after write` | ✅ 0 occurrences; `test-verify-fix-persistence.sh` unmodified, 79/79 |
| Fence guard still resists being fooled | ✅ Re-probed twice on the new head: `git rm` injected into the existing **indented** `git ls-files` fence, and a new top-level ` ```bash ` fence — both caught (2 assertions fail each), file restored |
| No new `/repo:*` command | ✅ `git ls-tree commands/repo/` identical to main |
| `.claude/` untouched | ✅ changed files are exactly `CHANGELOG.md`, `commands/repo/tidy.md`, `commands/repo/tests/test-tidy-keep-tiers.sh`, `hooks/repo/tests/run.sh` |
| Single `## Unreleased` heading | ✅ exactly 1; entry added under it, house style matches, and it describes the **final** three-condition behaviour |
| CI | ✅ `pnpm test (guard/handoff/loss-check suite)` pass, 1m25s |
| Merge state | ✅ `CLEAN` / `MERGEABLE` |
| `Closes #120` | ✅ present |

---

### Fixed during evaluation

The PR **body** still described the two-condition version (77 cases, 1373 total, 33-assertion negative control) — stale since the Doctor pass and the text that would land in a squash message. I appended an "Update after review" section correcting all four numbers and naming the three conditions. Description only; no code touched.

### Minor, non-blocking (no action needed)

1. `commands/repo/tests/test-tidy-keep-tiers.sh:371-372` and the CHANGELOG entry both say the ordinary-repo fixture is a "12-file tree" — it is 15 files now that the four adversarial companions were added. Cosmetic; 12 was my original count.
2. The awk's `tracked[]` map is keyed on a **lowercased** `dir+basename`, so the base-sibling lookup is case-insensitive (`Connectors.kicad_sch` satisfies `connectors_backup_….kicad_sch`). The `markers` fixture depends on this and it is consistent with the doc's case-insensitive marker matching — noting it only because the prose at `tidy.md:326` reads as an exact-name lookup.

Strong second pass. The narrowing was implemented harder than I suggested, the recall loss was surfaced rather than buried, and the `why:` clause closes the failure mode at the template level rather than by exhortation.

PR #169, opened 2026-08-07
feat: gate repo-remote reuse of fleet-marked hosts behind --force

### LINKED ISSUE #164: Design proposal: should repo-remote.sh's idle guard treat fleet-tagged/daemon-managed hosts differently at attach time?
## Context

This is a design proposal, not a bug report. It is the explicitly-deferred
follow-up flagged by 2AMLogic/2am#52's PR #70:

> "Whether `repo-remote.sh` should refuse to attach the idle guard to a host
> already running `loom-daemon` is a design decision belonging to the
> separate `rjwalters/repo` repository... It needs a follow-up issue filed
> there... needs a design proposal, not a bug report."

**Incident recap** (2AMLogic/2am#52): `loom-worker-1`, a persistent EC2 fleet
worker running `loom-daemon`, was provisioned via `/repo:remote` at some
point and inherited the idle-shutdown guard this repo installs for every
host. The guard fired during a token-pool-exhaustion window (no `claude`
process running is indistinguishable, to the guard, from "operator stepped
away") and powered the instance off. Recovery required a human to notice and
start the instance back up by hand, 4h40m later — because both the
systemd/launchd supervisor and `loom-daemon`'s own watchdog live on the box
that had just been shut off from under them.

A second, related finding from the same incident thread: the host was still
tagged `repo-remote=anvil` (the tag `repo-remote.sh --status`/`--down`
resolve instances by) well after it became a persistent fleet worker — so
ephemeral-dev-session tooling could still discover and manage a
production host without anyone intending that.

## What's already in place (read from current `origin/main`, `cdbffbd`)

This repo has already made two deliberate, well-documented decisions here,
and I want to be precise about what's *not* being asked for before proposing
what might still be missing:

1. **No process-name veto, and that's intentional.** The exact current
   comment above `idle_guard_userdata()` (`scripts/repo/repo-remote.sh:262-266`):

   > "Activity" is defined by exactly two local signals: an open SSH session
   > (`who`) OR CPU load average > 0.2. There is NO process-name veto — a
   > running daemon (loom-daemon or otherwise) does NOT, by itself, keep this
   > host alive. If a future daemon-presence veto is ever wanted it must be
   > added deliberately here and documented; it is not implied by the current
   > logic.

   `commands/repo/remote.md`'s "The idle-shutdown guard" section restates
   this and additionally documents a **recommended short window
   (`REPO_REMOTE_IDLE_SHUTDOWN_MIN=20`) for daemon-managed/worker hosts**
   plus an **idle-exit marker contract** (`REPO_REMOTE_IDLE_MARKER`, added in
   #79) that lets a daemon `touch` a file on clean idle-exit to hand the
   guard a precise "idle since" timestamp. This is a real, working mechanism
   — it's just opt-in and cooperative (the daemon has to know to write the
   marker), not a veto.

2. **#163** (open) covers a distinct, narrower bug: `REPO_REMOTE_IDLE_SHUTDOWN_MIN=0`
   currently shuts a host down almost immediately instead of disabling the
   guard, which is the literal opt-out an operator would reach for. This
   issue does not overlap with #163 — #163 is about making `0` mean "no
   guard installed at all"; this issue is about *whether provisioning a
   fleet-tagged/daemon-managed host should default to that opt-out (or some
   other guard-rail), rather than requiring every caller to remember to pass
   it*.

Given both of those, the on-host guard logic itself (`who`/load/marker) is
probably *not* where more should be added — re-litigating "should the guard
sniff for `loom-daemon`'s process name" would just reopen a decision this
repo already made carefully in #79, and process-name sniffing is fragile in
exactly the way that decision anticipated. The gap that's left is upstream of
the guard: **at provisioning/attach time**, not inside the generated cron
script.

## The proposal

Two non-exclusive design options, both scoped to *when a guard gets attached
at all*, not to the guard's own runtime logic:

### Option A -- a `fleet-worker`/`daemon-managed` role that changes defaults, not behavior

Add a role knob (e.g. `REPO_REMOTE_ROLE=fleet-worker` or a `--role` flag)
that, when set, changes what `repo-remote.sh` defaults to for that host --
without adding any new on-host detection logic:

- `IDLE_MIN` defaults to `0` (fully disabled, once #163 ships) instead of
  `120`, on the theory that a daemon-managed host should require an
  *explicit, positive* opt-in to being auto-shut-down rather than an
  explicit opt-out.
- `--status` output labels the host's guard state plainly (e.g. `guard:
  disabled (role=fleet-worker)`) so a human running `/repo:remote --status`
  against a tagged host sees the guard's state instead of it being silent
  either way.

This keeps the guard's own logic exactly as-is (no process detection, no new
runtime signal) -- it only changes what a caller has to type to get a
fleet-appropriate default, and makes that default visible.

### Option B -- a provisioning-time tag check, not a runtime process check

Before attaching or *re*-attaching guard user-data (i.e., on `--up` against
an existing instance, or reprovision), check the target instance's own AWS
tags for a marker that a fleet-management tool already sets -- e.g. the
`Fleet=loom` tag 2am's remediation added to both `loom-worker-1` and
`loom-worker-2` specifically so an external identity could recover a stopped
instance (2AMLogic/2am#52, 2AMLogic/2am#51). If present, warn (or require an
explicit `--force`) rather than silently reattaching a fresh 120-minute
guard. This is a provisioning-time check against a tag the operator already
had to set deliberately elsewhere -- not a fragile on-host "is this process
running" heuristic, and it directly closes the second finding from the
incident (`repo-remote=anvil` tooling being able to rediscover and
reprovision a host that had since become `loom-worker-1`).

Option A and B compose: A changes what gets installed by default for a role
that says "I am a fleet worker"; B catches the case where a caller doesn't
know that and is about to attach a guard to a host tagged as one anyway.

## Explicitly not proposed

- Removing or weakening the guard for genuine ephemeral dev sessions -- the
  `who`/load model and 120-minute default are correct for that case and
  should not change.
- Any form of process-name sniffing (`pgrep -f loom-daemon` or similar)
  inside the on-host guard script -- #79's design note already rejected this
  approach for the guard's own runtime logic, and nothing here should
  reopen that.

## Cross-references

- 2AMLogic/2am#52 -- the incident and its acceptance criteria (repo-side
  pieces already merged: `hosts.yml` `cloud.idle_guard` field, EC2 live-state
  drift detection, manual remediation of `loom-worker-1`'s tags/cron).
- rjwalters/repo#163 -- the narrower `IDLE_MIN=0` semantics bug; Option A
  above depends on that shipping first if `IDLE_MIN=0` is the mechanism used
  for "fully disabled."
- #79 -- added the idle-exit marker contract and the daemon-host short-window
  documentation; establishes the "no process-name veto, ever, in the guard
  itself" precedent this proposal deliberately does not challenge.

## Test Plan (if either option is adopted)

- [ ] Option A: regression test that `REPO_REMOTE_ROLE=fleet-worker` with no
      explicit `REPO_REMOTE_IDLE_SHUTDOWN_MIN` produces user-data with no
      guard cron installed (once #163 ships), and that an explicit
      `REPO_REMOTE_IDLE_SHUTDOWN_MIN=<n>` still overrides the role default.
- [ ] Option A: `--status` output test asserting the guard-state line reflects
      the role default.
- [ ] Option B: test against a mocked `describe-instances` response carrying
      a `Fleet=loom` tag, asserting `--up`/reprovision against that instance
      id warns or requires `--force` before attaching user-data.

---

## Curator Enhancement

Verified against `origin/main` @ `cdbffbd`, 2026-08-06.

### Citations confirmed

- `REPO_REMOTE_IDLE_SHUTDOWN_MIN` default (120) — `scripts/repo/repo-remote.sh:200`.
- "No process-name veto" doc comment — `scripts/repo/repo-remote.sh:263-266` (issue cited 262-266; off by one, same drift #163 noted independently).
- `idle_guard_userdata()` definition — `scripts/repo/repo-remote.sh:278`.

### Correction: Option B's premise doesn't match current reuse behavior

Option B assumes `--up` against an existing/reused instance **re-attaches**
guard user-data ("rather than silently reattaching a fresh 120-minute
guard"). Reading `aws_up()` (`scripts/repo/repo-remote.sh:410-451`) and
`gcp_up()` (`scripts/repo/repo-remote.sh:509-534`) end to end: neither
provider re-attaches user-data to a reused instance. `idle_guard_userdata`
is only invoked from `aws_create()` (line 383) and `gcp_up()`'s
create-branch (line 528) — both **brand-new-instance** paths. AWS user-data
is immutable post-creation via this code (no
stop→modify-instance-attribute→start cycle exists here); a reused/pinned/
tag-found instance is only started (`start-instances`) or SSH-aliased, never
re-guarded. So "a fresh 120-minute guard silently reattached to a fleet host"
cannot happen today — the guard an instance carries is whatever it got at
its one-time creation.

The actual second-finding risk is narrower but still real: **reuse
discovery**. `aws_up()`'s pinned-`REPO_REMOTE_INSTANCE_ID` branch
(`scripts/repo/repo-remote.sh:415-424`) and its `aws_find_tagged()` branch
(`scripts/repo/repo-remote.sh:426-436`, matching on the `repo-remote=<name>`
tag) — plus `gcp_up()`'s analogous reuse branch
(`scripts/repo/repo-remote.sh:512-517`) — will happily start/stop and
SSH-alias *any* instance they find under that name/tag, including one that
has since become a tagged fleet worker, with no awareness that ephemeral
dev-session tooling is now touching a production host. This is exactly what
let `repo-remote=anvil`-tagged tooling keep discoverying `loom-worker-1`
after it became a persistent fleet worker.

**Implication for Option B**: gate the check on the *reuse-discovery* branches
above (start/stop/alias of a found instance), not on "guard reattachment,"
which doesn't occur in this codebase. The warn-or-`--force` check belongs
right after `aws_find_tagged`/pinned-ID resolution and the `gcp_up` reuse
branch — before the instance is started or written back to the SSH alias —
checking the resolved instance's tags (or GCP labels) for the fleet marker
(e.g. `Fleet=loom`) 2am's remediation already sets.

### Recommendation

Scope this issue's initial implementation to **Option B**, corrected as
above — it has no dependency on #163, and it's the piece that directly closes
the second incident finding (stale-tag rediscovery of a production host by
ephemeral tooling). Track **Option A** (role-based defaults) as a natural
follow-up once #163 ships, since Option A's "fully disabled" default is only
meaningful after #163 makes `REPO_REMOTE_IDLE_SHUTDOWN_MIN=0` actually mean
"no guard installed." Bundling both into one PR risks exceeding a single
Builder PR's scope guardrails (>8 files / >400 LOC) once tests and docs for
both are included — splitting keeps Option B unblocked and immediately
buildable.

### Acceptance Criteria (Option B, this issue's scope)

- [ ] `repo-remote.sh`'s AWS reuse paths (pinned-`REPO_REMOTE_INSTANCE_ID`
      and `aws_find_tagged`) check the resolved instance's tags for a
      configurable fleet marker (e.g. `Fleet=loom`, tag key/value
      configurable via env var) before starting/aliasing it.
- [ ] `repo-remote.sh`'s GCP reuse path (`gcp_up`'s existing-instance
      branch) performs the equivalent check against instance labels.
- [ ] When the marker is present, the command warns clearly and requires an
      explicit `--force` (or equivalent) to proceed, rather than silently
      starting/aliasing the host.
- [ ] `commands/repo/remote.md` documents the check and the override flag.
- [ ] Regression tests cover: marker absent (unchanged behavior), marker
      present without `--force` (blocked with clear message), marker present
      with `--force` (proceeds).

### Affected Files

- `scripts/repo/repo-remote.sh` — `aws_up()` reuse branches (~lines 415-436)
  and `gcp_up()`'s reuse branch (~lines 512-517): add the fleet-tag/label
  check and `--force` gate before start/alias.
- `commands/repo/remote.md` — document the new check and override flag,
  alongside the existing idle-guard documentation (~line 320 area, where
  daemon-managed/fleet-worker guidance already lives).
- `commands/repo/tests/test-repo-remote.sh` — add regression cases for the
  three scenarios in Acceptance Criteria above.

### Dependencies

No hard dependency for Option B (this issue's recommended scope). Option A,
if picked up later as a follow-up, depends on #163 (`REPO_REMOTE_IDLE_SHUTDOWN_MIN=0`
semantics) shipping first — #163 is `loom:curated` + `loom:issue` (approved,
not yet built) as of this enhancement.

### Related open work / duplicate check

`check-duplicate.sh --include-merged-prs --issue 164` surfaced #163 as a
29%-similarity match and as related open work (it cross-references #164).
Disregarded as a duplicate: the issue body already distinguishes the two in
detail (#163 is the narrower `IDLE_MIN=0` semantics bug; this issue is the
provisioning-time design question). No further reconciliation needed beyond
the Recommendation/Dependencies sections above.

rjwalters:
## Summary

Implements **Option B only** from #164 (the Curator's corrected scope and the Champion-approved direction); Option A (role-based idle-guard defaults) is explicitly deferred as a follow-up.

`repo-remote up` resolves an instance to *reuse* from handles that never expire — a pinned `REPO_REMOTE_INSTANCE_ID`, or the `repo-remote=<name>` tag/label — then starts it and rewrites this repo's SSH alias to point at it. A box provisioned once for an ephemeral dev session can since have been repurposed into a persistent, daemon-managed fleet worker while still carrying the old tag, so that discovery could silently start and re-alias a production host. That is the second finding of the 2AMLogic/2am#52 incident (`repo-remote=anvil` tooling kept rediscovering the host that had become `loom-worker-1`).

Per the Curator's correction, the gate is on **reuse discovery**, not on "guard reattachment" — this codebase never re-attaches user-data to a reused instance, so no such reattachment exists to gate.

## Changes

- **`scripts/repo/repo-remote.sh`**
  - New `fleet_marker_matches()` / `fleet_marker_gate()` plus provider readers `aws_fleet_marker()` (instance tags) and `gcp_fleet_marker()` (instance labels, key lower-cased for GCP).
  - Gate wired into all three reuse branches, **before** any `start-instances` / `instances start` / SSH-alias write: `aws_up()`'s pinned-`REPO_REMOTE_INSTANCE_ID` branch, `aws_up()`'s `aws_find_tagged()` branch, and `gcp_up()`'s existing-instance branch.
  - New `--force` flag. Without it, a match exits **5** with a message naming the instance, the marker found, and every remediation (clear the pin, remove the stale tag, `--force`, or disable the check). With it, a loud `WARNING` is emitted and the run proceeds.
  - Config: `REPO_REMOTE_FLEET_TAG_KEY` (default `Fleet`; **empty disables the check**) and `REPO_REMOTE_FLEET_TAG_VALUE` (default `loom`; empty = any non-empty value). Value comparison is case-insensitive since GCP lower-cases label values. Both use `${VAR-default}` so an explicit empty setting is honored rather than falling back.
  - `usage()` now scans the leading comment block with `awk` instead of a hard-coded `sed -n '4,60p'` range, so `--help` does not silently truncate now that the header documents the new guard.
- **`commands/repo/remote.md`** — new `#### Fleet-marked hosts: the reuse guard` subsection immediately after the idle-guard docs; `--force` and the exit-code table in the headless-usage block; the guard as step 3.3 of "Reuse or find the instance"; the two new env vars in the `.env` reference.
- **`commands/repo/tests/test-repo-remote.sh`** — mock `aws` now answers the `Tags[?Key=…]` projection; a new minimal mock `gcloud` covers `gcp_up()`; 30 new cases.

`--force` deliberately does **not** relax the cost gate — an `up --yes --force` with no `REPO_REMOTE_INSTANCE_TYPE` still fails with exit 2.

| Criterion (from #164) | Status | Verification |
|---|---|---|
| AWS reuse paths (pinned id + `aws_find_tagged`) check the resolved instance's tags for a configurable fleet marker before starting/aliasing | Yes | `fleet_marker_gate` calls added in both branches ahead of `start-instances` and `write_ssh_alias`; tests (b) and (d) assert `grep -c 'start-instances'` is `0` on a blocked run |
| GCP reuse path (`gcp_up`'s existing-instance branch) performs the equivalent check against labels | Yes | `gcp_fleet_marker` + gate before `instances start`; new mock `gcloud` drives the three GCP cases |
| Marker present ⇒ warns clearly and requires explicit `--force`, rather than silently starting/aliasing | Yes | Exit `5` with instance id, `Fleet=loom` and `--force` in the message; blocked run asserted to leave the SSH alias byte-identical |
| `commands/repo/remote.md` documents the check and the override flag | Yes | New subsection + usage/exit-code/env-var docs; 5 new doc-drift assertions pin them so script and prose can't diverge |
| Regression tests: marker absent (unchanged), present without `--force` (blocked, clear message), present with `--force` (proceeds) | Yes | Cases (a)/(b)/(c) for AWS pinned, (d) for tag discovery, and the three GCP cases; plus non-matching value, custom key/value, empty-key opt-out, and dry-run-never-blocked |

Out of the issue's explicit non-goals: no process-name sniffing was added anywhere, and the idle guard's `who`/load activity model is byte-for-byte unchanged.

## Test Plan

- `commands/repo/tests/test-repo-remote.sh` — **133 pass, 0 fail** (was 103; +30 new cases).
- Full suite `./hooks/repo/tests/run.sh` (= `pnpm test` / `pnpm check:ci`) — **PASS: 1612, FAIL: 0**.
- `bash -n scripts/repo/repo-remote.sh` clean; `shellcheck` output on both changed shell files is **identical to the `origin/main` baseline** (2×SC2015, 2×SC2016, 3×SC2059, 2×SC2094 — all pre-existing info-level findings in untouched code, no new lint).
- `repo-remote --help` verified to render the full header after the `usage()` change.

## Pre-existing Issues (Not Addressed)

The shellcheck info-level findings above pre-date this PR and live in code this change does not touch (`write_ssh_alias`, `parse_args`, and the test suite's `printf` colour formatting). Left alone to keep the diff scoped.

Closes #164

### DIFF
--- a/commands/repo/remote.md
+++ b/commands/repo/remote.md
@@ -44,6 +44,7 @@ loom's `fleet add-worker`) invokes directly:
 ```
 repo-remote up [aws|gcp]              # DRY RUN: print the resolved plan + estimated cost as JSON, create NOTHING
 repo-remote up --yes [--json]        # provision (or reuse) with no prompts; requires pre-supplied cost-relevant config
+repo-remote up --yes --force         # additionally override the fleet-marker guard (see below)
 repo-remote status [--json]          # list instances tagged repo-remote=<name>
 repo-remote down [--yes] [--delete]  # dry-run listing without --yes; stop (or --delete to terminate) with --yes
 ```
@@ -55,6 +56,17 @@ billable size on its own. Plain `repo-remote up` (no `--yes`) is a dry run that
 prints the plan and estimated hourly cost and spends nothing, so a caller can
 implement a "plan shown before money is spent" check against the `--json`
 output (instance id, public IP, SSH alias, estimated hourly cost).
+
+`--force` is a **separate** override with a single job: it lets `up` reuse an
+instance carrying a **fleet marker** (see [Fleet-marked hosts: the reuse
+guard](#fleet-marked-hosts-the-reuse-guard)). It does **not** relax the cost
+gate — `--force` without a pre-supplied `REPO_REMOTE_INSTANCE_TYPE` still fails
+loudly.
+
+Exit codes: `0` success (including a dry-run plan), `2` missing/invalid required
+config (the cost gate), `3` provider authentication failed, `4` cloud operation
+failed, `5` refused to reuse a fleet-marked instance (pass `--force`), `64`
+usage error.
 
 ## Configuration — two layers
 
@@ -129,6 +141,10 @@ REPO_REMOTE_SETUP="make setup"            # optional first-boot command; fallbac
 # --- session ---
 REPO_REMOTE_IDLE_SHUTDOWN_MIN=120         # interactive-host default; use ~20 for daemon-managed/worker hosts; 0 disables the guard entirely (see below)
 REPO_REMOTE_IDLE_MARKER=                   # optional idle-exit marker path (default: /var/run/repo-remote-daemon-idle.marker)
+
+# --- fleet-marker guard (refuse to reuse a managed fleet host) ---
+REPO_REMOTE_FLEET_TAG_KEY=Fleet           # tag (AWS) / label (GCP) key that marks a managed fleet host; empty disables the check
+REPO_REMOTE_FLEET_TAG_VALUE=loom          # required value for that key; empty means "any non-empty value counts"
 ```
 
 Only `REPO_REMOTE_PROVIDER` (or a provider argument) and that provider's
@@ -273,6 +289,11 @@ gcloud compute instances list --filter="labels.repo-remote=<name>" \
 
 RUNNING → offer reuse; STOPPED → offer to start.
 
+3. **Before starting or SSH-aliasing anything you *reused*** (either branch
+   above), check the resolved instance's tags/labels for a **fleet marker** and
+   stop unless `--force` was given — see **Fleet-marked hosts: the reuse guard**
+   below. A freshly created instance is never subject to this check.
+
 ### 4. Create the instance (with confirmation)
 
 **Before creating anything**, show the exact command, the machine spec
@@ -365,6 +386,56 @@ anything ever writes the file:
 On a daemon-managed host the two stages compose: the daemon's own idle-exit
 (writing the marker) is the first stage; this guard, `IDLE_MIN` minutes later, is
 the second and final stage that actually powers the box off.
+
+#### Fleet-marked hosts: the reuse guard
+
+The idle guard above is attached **once, at creation** — this command never
+re-attaches user-data to an instance it reuses, so the guard a host carries is
+whatever it got the day it was created. What reuse *does* do is **start a
+stopped instance and rewrite this repo's SSH alias to point at it**, and the
+handles it resolves that instance from — a pinned `REPO_REMOTE_INSTANCE_ID`, or
+the `repo-remote=<name>` tag/label — never expire. A box provisioned once for an
+ephemeral dev session can since have been repurposed into a persistent,
+daemon-managed fleet worker while still carrying the old tag, at which point
+ephemeral dev-session tooling would silently start and re-alias a production
+host. That is the second finding of the 2AMLogic/2am#52 incident, where
+`repo-remote=anvil` tooling kept rediscovering the host that had become
+`loom-worker-1`.
+
+**The check.** Before starting or aliasing a **reused** instance — on both AWS
+reuse branches (pinned id and `repo-remote=<name>` tag lookup) and on the GCP
+existing-instance branch — the resolved instance's **AWS tags** / **GCP labels**
+are read and matched against a fleet marker, by default **`Fleet=loom`** (the
+tag 2am's own remediation sets on its persistent workers). If it matches:
+
+- **Without `--force`**: the run **stops with exit `5`** and a message naming the
+  instance, the marker it found, and the override. Nothing is started, nothing is
+  terminated, and the SSH alias is left untouched.
+- **With `--force`**: it warns loudly that it is about to touch what looks like a
+  managed fleet host, then proceeds as normal.
+
+**Configuration.**
+
+- `REPO_REMOTE_FLEET_TAG_KEY` — the tag/label key to look for. Defaults to
+  `Fleet`. **Set it to the empty string to disable the check entirely.** GCP
+  label keys are lower-case, so the configured key is lower-cased for the GCP
+  lookup (`Fleet` → `labels.fleet`).
+- `REPO_REMOTE_FLEET_TAG_VALUE` — the value that counts as a match. Defaults to
+  `loom`; comparison is case-insensitive (GCP lower-cases label values). Set it
+  to the empty string to treat *any* non-empty value of the key as a match.
+
+**Scope — what this deliberately is not.** It is a **provisioning-time check
+against declared metadata** an operator already had to set elsewhere, not an
+on-host "is `loom-daemon` running" heuristic. The idle guard's activity model
+above is unchanged, and gains no process-name veto from this; the two are
+independent. It also does not fire on **creation** (a brand-new instance has no
+prior role to protect) or on a **dry run** (`repo-remote up` without `--yes`
+touches no cloud resource at all).
+
+**Related but distinct**: if the fleet host should also never power itself off,
+that is the idle guard's `REPO_REMOTE_IDLE_SHUTDOWN_MIN=0` opt-out above. This
+guard stops *this tooling* from touching the host; that one stops the *host*
+from shutting itself down.
 
 #### GPU hosts
 
--- a/commands/repo/tests/test-repo-remote.sh
+++ b/commands/repo/tests/test-repo-remote.sh
@@ -121,6 +121,12 @@ case "$1 $2" in
   "ec2 describe-images")
     echo "${MOCK_AWS_AMI:-ami-0ubuntu2204}"; exit 0 ;;
   "ec2 describe-instances")
+    # Fleet-marker lookup (repo#164) is also an --instance-ids call, so it must
+    # be matched FIRST by its Tags[?Key=...] query projection.
+    if printf '%s' "$*" | grep -qF "Tags[?Key=="; then
+      printf '%s\n' "${MOCK_AWS_FLEET_TAG:-None}"
+      exit 0
+    fi
     # --instance-ids (pinned lookup) vs --filters (tag find)
     if printf '%s' "$*" | grep -q -- '--instance-ids'; then
       echo "${MOCK_AWS_STATE:-None}"
@@ -152,6 +158,29 @@ case "$1 $2" in
 esac
 MOCK
 chmod +x "$MOCK_BIN/aws"
+
+# ---------------------------------------------------------------------------
+# A minimal mock `gcloud`, covering only the calls gcp_up() makes. Added for the
+# GCP half of the fleet-marker guard (repo#164) so the label-based check is
+# exercised, not just the AWS tag-based one.
+# ---------------------------------------------------------------------------
+cat >"$MOCK_BIN/gcloud" <<'MOCK'
+#!/usr/bin/env bash
+printf '%s\n' "$*" >>"${MOCK_AWS_LOG:-/dev/null}"
+case "$1 $2 ${3:-}" in
+  "compute instances describe")
+    if printf '%s' "$*" | grep -qF 'labels.'; then
+      printf '%s\n' "${MOCK_GCP_FLEET_LABEL:-}"      # fleet-marker lookup
+    elif printf '%s' "$*" | grep -qF 'natIP'; then
+      printf '%s\n' "${MOCK_GCP_IP:-1.2.3.4}"        # public IP lookup
+    else
+      printf '%s\n' "${MOCK_GCP_STATE:-}"            # existence/status lookup
+    fi
+    exit 0 ;;
+esac
+exit 0
+MOCK
+chmod +x "$MOCK_BIN/gcloud"
 
 echo "repo-remote.sh test suite"
 echo "========================="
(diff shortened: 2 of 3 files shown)


rjwalters:
## Approved — LGTM

Reviewed against #164's **Option B** acceptance criteria. All five are met, the implementation is correctly placed, and I independently reproduced every claim in the PR description rather than taking them on trust.

### Verification I ran myself

| Check | Result |
|---|---|
| `commands/repo/tests/test-repo-remote.sh` | **133 pass / 0 fail** (was 103) |
| Full suite `./hooks/repo/tests/run.sh` (= `pnpm test` / `pnpm check:ci`) | **PASS: 1612, FAIL: 0** |
| GitHub CI (`pnpm test (guard/handoff/loss-check suite)`) | **pass**, 1m30s |
| `mergeStateStatus` | **CLEAN** / MERGEABLE |
| `bash -n scripts/repo/repo-remote.sh` | clean |
| `shellcheck -S warning` on the changed script, PR vs `origin/main` | clean on **both** — no new lint |
| `repo-remote --help` after the `usage()` rewrite | renders the full 77-line header, exit 0 |

### Acceptance criteria

- **AWS reuse paths gated before start/alias** — `fleet_marker_gate` is called in the pinned-`REPO_REMOTE_INSTANCE_ID` branch (`aws_up`, before the `case "$state"` that issues `start-instances`) and in the `aws_find_tagged` branch (before its own `start-instances`). Both sit ahead of `writeback_instance_id` / `write_ssh_alias`. Confirmed.
- **GCP reuse path gated equivalently** — `gcp_fleet_marker` + gate in `gcp_up`'s existing-instance branch, before `instances start`, reading `labels.<key>` with the key lower-cased. Confirmed.
- **Marker present ⇒ warn + require `--force`** — exit `5`, message names the instance, the marker (`Fleet=loom`), and four distinct remediations. With `--force`, a loud `WARNING` then proceeds. Confirmed.
- **`commands/repo/remote.md` documents both** — new `#### Fleet-marked hosts: the reuse guard` section, `--force` in the headless-usage block, the exit-code table, and both env vars in the `.env` reference. The 5 new doc-drift assertions pin them, so prose and script can't silently diverge.
- **Regression tests cover absent / present-without-force / present-with-force** — cases (a)/(b)/(c) for the pinned path, (d) for tag discovery, three GCP cases, plus non-matching value, custom key/value, empty-key opt-out, and dry-run. The blocked-run assertions are the strong ones: `grep -c 'start-instances' == 0`, `grep -c 'run-instances' == 0`, and the SSH alias asserted byte-identical.

### Independent adversarial probing (beyond the suite)

I built a separate mock outside the repo's harness and confirmed:

- **`pending` state is guarded too.** The suite only exercises `stopped` and `running`; the pinned branch's `*)` catch-all also reuses-and-aliases. The gate sits above the whole `case`, so `state=pending` + `Fleet=loom` → exit 5, no SSH alias written. Good placement — a gate inside the `case` would have missed this.
- **Case-insensitivity is real**, not just asserted: tag value `LOOM` blocks against the default `loom`.
- **Empty `REPO_REMOTE_FLEET_TAG_VALUE`** genuinely matches any non-empty value (tag `anything` → exit 5).
- **`${VAR-default}` (not `${VAR:-default}`)** is the right call for `FLEET_TAG_KEY` — an explicitly-empty setting is the documented opt-out and must not fall back to `Fleet`.

### Notable non-AC improvement

The `usage()` change from `sed -n '4,60p'` to the `awk` scan-to-first-non-comment is a real latent-bug fix, not incidental churn: the header now runs to 77 lines, so the old hard-coded range would have silently truncated `--help` mid-sentence starting with this very PR. Fixing it in the same change that grows the header is the right call.

### Non-blocking observations (no action required for this PR)

1. **The marker read fails open.** `aws_fleet_marker` swallows errors (`2>/dev/null` + `|| return 0`) and `gcp_fleet_marker` uses `|| true`, so an errored lookup yields `""` and the guard passes. I confirmed this: forcing only the `Tags[?Key==...]` call to exit 255 while the state lookup succeeds gives rc=0 with the alias written. Real-world exposure is narrow — both AWS calls need the same `ec2:DescribeInstances` permission, so a permissions failure would already have driven `state=missing` into the create path — leaving transient throttling as the only realistic trigger. Fail-open is defensible for a warn-guard, and fail-closed would make the tool brittle against API hiccups. Flagging it as a deliberate, undocumented choice rather than a defect.

2. **`down` is not gated, and it is the higher-consequence sibling.** `aws_down` resolves instances from the same never-expiring handles this PR just hardened (`REPO_REMOTE_INSTANCE_ID`, or the `repo-remote=<name>` tag filter). I verified that `repo-remote down --yes --delete` **terminates** a `Fleet=loom`-marked instance with rc=0 and no warning. This is strictly outside #164's Option B ACs, which scope explicitly to reuse/start/alias in `up`, so it is correctly not in this PR — but it is the same stale-handle hazard with a worse outcome (terminate, disk goes with it, vs. start-and-realias). I'm filing a `loom:triage` follow-up rather than expanding this PR's scope.

3. **Cosmetic**: an AWS tag whose literal value is the string `None` is indistinguishable from an absent tag (the `--output text` sentinel collision, handled in both `aws_fleet_marker` and `fleet_marker_matches`). Not worth code.

4. Reuse now costs one extra `describe-instances` call. Negligible, and keeping it separate from `aws_instance_state` keeps both helpers single-purpose.

### Scope

Three files, +321/-2, all within the issue's boundary. Option A is correctly deferred. The issue's explicit non-goals hold: no process-name sniffing anywhere, and the idle guard's `who`/load activity model is byte-for-byte unchanged. The `--force` / cost-gate separation (`up --yes --force` without `REPO_REMOTE_INSTANCE_TYPE` still exits 2) is the right boundary — one override should not quietly widen into another.

Approving. Moving to `loom:pr` for Champion auto-merge.

Issue #196, opened 2026-08-08
/repo:release should flag contradictory entries within one Unreleased section


rjwalters:
## Context

Hit for real while staging the 0.9.0 release.

The `## Unreleased` section had accumulated ten entries written over several sessions. Two of them were about the same subject and, read in sequence, contradicted each other:

> **The README and SKILL.md no longer claim that Loom defers to this repo's destructive-command guard, because it does not (#168).**

> **The canonical guard reaches parity with Loom's vendored copy and now actually runs in a Loom-managed repo (#168, #188).**

Both were accurate when written. Both would have shipped in the same release. A reader of 0.9.0's changelog would see the release simultaneously asserting that deferral does not hold and that it does — with no way to tell which is the end state.

Nothing caught this. `/repo:release` performed its normal pre-flight and would have cut the release with both entries intact. It was found by reading the section top to bottom before cutting, which is not a control.

## Why this is structural, not a one-off

Changelog entries are **written serially**, at the moment each piece of work lands, and **read all at once**, from the vantage point of the finished release. Any reversal inside one release cycle produces this, and the longer the cycle the more likely it is. The specific hazard shapes:

- A claim corrected, then the underlying thing fixed so the original claim becomes true again (exactly what happened here)
- A feature added, then substantially reworked or renamed before release
- A default flipped, then flipped back
- A limitation documented, then removed

The failure mode is quiet: every individual entry is defensible, so review of any single PR passes.

## Proposal

Add a coherence check to `/repo:release`, most naturally at the `pre-changelog-style` seam that already exists in `commands/repo/release.md`'s seam list.

This is a judgment call, not a regex — the right implementation is probably to have the model **read the whole `## Unreleased` section as one document** and report entries that appear to contradict, supersede, or duplicate one another, rather than pattern-matching. That fits how the other `/repo:*` commands work.

Report-and-confirm, never auto-edit: merging two entries changes the release's public description of itself, which is an operator call.

## Suggested acceptance criteria

- [ ] `/repo:release` reads the full `## Unreleased` section as a single document before cutting and reports entries that contradict or supersede one another
- [ ] Also flags near-duplicates — two entries describing the same change from different PRs, which the same serial-authoring dynamic produces
- [ ] Findings are reported with both entry excerpts side by side, so the operator can see the contradiction rather than being told one exists
- [ ] **Report-and-confirm only** — never rewrites or merges entries automatically
- [ ] Bound to the existing `pre-changelog-style` seam rather than a new phase
[...]

Issue #195, opened 2026-08-08
No configurable positional-arg masking allowlist in the canonical guard


rjwalters:
## Context

A deliberate non-port from the guard parity work (#188, PR #192), recorded here so the gap is a decision rather than an oversight.

Loom's vendored guard carries `mask_ask_positional_args()` and `mask_catastrophic_positional_args()`. They mask quoted **positional** arguments that directly follow a known non-executing command, so a read-only tool's own search text is not misread as live shell syntax. The motivating case upstream: a dedup script invoked as `check-duplicate.sh "TITLE" "DESCRIPTION"` where the description quotes something dangerous, and `grep -n "curl .*|" <file>` — introspecting a guard's own source — being misread as a live curl-pipe-to-shell.

Neither function was ported, for a specific reason: the command allowlist is a hardcoded Loom path.

```awk
cmdre = "(\\./\\.loom/scripts/check-duplicate\\.sh)"
```

That is one tool's script name compiled into a guard whose whole premise is being tool-agnostic. Porting it verbatim would have shipped a rule that fires for exactly one repo's helper and no one else's.

It is also worth being precise about what the omission does and does not cost. Most of the divergences originally attributed to positional masking during #192 turned out to have a different cause entirely — four checks scanning the raw command instead of the literal-redacted `COMMAND_ASK_SCAN` — and fixing that at the scan site resolved them. So the remaining exposure is narrower than the function list implied, and the equivalence harness showed **zero** divergent cases attributable to the masking functions themselves across the probe corpus.

## The actual remaining gap

A repo whose own tooling takes quoted positional arguments containing destructive-looking text can still get a false ask or deny, with no way to opt that command out. There is no configurable allowlist — the only lever is disabling a whole guard category via `guards.*`, which is far too blunt.

This is the same class of false positive that the `COMMAND_ASK_SCAN` fix addressed for flag *values* (`-m`, `--body`, `--title`). Positional arguments simply never got the equivalent treatment here.

## Suggested acceptance criteria

- [ ] Positional-argument masking exists in the canonical guard with a **configurable** command allowlist — read from `guards.*` config (and the matching `REPO_*` env override), not hardcoded
- [ ] Empty/absent config is the default and is a no-op, so no repo changes behavior on upgrade
- [ ] Masking stops at the first token that is not a quoted string, so a real invocation chained onto the same line stays fully visible (this is load-bearing in the upstream implementation and must not be lost)
[...]

Issue #188, opened 2026-08-08
Implement Bash-tool write confinement in the canonical guard so Loom's dispatcher defers to it


rjwalters:
## Context

Split out of #168, which is now resolved on its documentation half. #168 established two facts:

1. The canonical guard (`hooks/repo/guard-destructive.sh`) never executes in a Loom-managed repo, because Loom's dispatcher requires **both** a version probe and a capability probe, and the capability probe — `grep -q 'worktree-write-confinement'` — fails. Confirmed: 0 occurrences in the canonical guard at 0.8.1.
2. Nothing is unguarded. Loom's vendored `guard-destructive-generic.sh` carries write confinement and does run.

#168's docs half shipped: `README.md` and `skills/repo/SKILL.md` no longer claim that Loom defers to this copy, and both now state that deferral is conditional on a capability probe that does not currently pass.

This issue is the remaining half: **implement the category**, so the deferral becomes real.

## Why it is worth doing

Loom's dispatcher is already written for exactly this. The moment the canonical guard emits `worktree-write-confinement`, the probe passes and Loom `exec`s this guard instead of its vendored copy — no change on Loom's side. Until then, every improvement to the canonical guard is unreachable in the environment it is most likely to be installed in.

## What to implement

Port the **BASH-TOOL WRITE CONFINEMENT** block from `.loom/hooks/guard-destructive-generic.sh` (approximately lines 4031-4346, ~315 lines) into `hooks/repo/guard-destructive.sh`, and emit the `worktree-write-confinement` decision tag on deny.

The block confines Bash-tool writes (`>`, `>>`, `tee`, `sed -i`, `cp`, `mv`) to a builder's issue worktree, closing the escape where a session denied on Edit/Write falls back to a Bash write and lands the same edit in the main checkout.

### Transplant it — do not reimplement it

This is the load-bearing instruction. That block carries scars from two production incidents, both recorded in its own comments:

- **Loom #4178** — the escape it exists to close: a sweep edited live guard hooks in the main checkout while its own worktree stayed clean.
- **Loom #4495** — `pwd -P` alone was insufficient. Write targets are produced by a lexical-only `normalize_abs_path()`, which keeps a symlinked ancestor intact, so a repo reached through a symlinked path (a `/tmp` checkout on macOS, a symlinked home, a bind-mounted workspace) never string-matched the physical root and **every** Bash write into the main checkout was silently allowed. Both the physical and logical spellings of the main-checkout root must be checked.

[...]

PR #198, opened 2026-08-08
feat(guard): equivalence harness, git -C stash threading, and the quoting-bypass fix

### LINKED ISSUE #193: Check in the guard-equivalence harness so drift between the canonical and vendored guards stops being invisible
## Context

Surfaced while reconciling the canonical guard against Loom's vendored copy (#188, PR #192).

To establish parity I built a throwaway harness that pipes the same PreToolUse payload to both `hooks/repo/guard-destructive.sh` and `.loom/hooks/guard-destructive-generic.sh` and diffs the resulting `permissionDecision`. It ran ~60 probe cases across every category the two guards share.

It changed the conclusion materially. Counting functions and grep hits suggested four capability gaps. The verdict diff showed:

| Area | Line counts suggested | Verdict diff showed |
|---|---|---|
| `systemctl` | 3 refs vs 20 — big gap | **0 divergent cases** |
| Fastpath tiering | 5 functions missing | **0 divergent cases** |
| `git stash` | 0 refs vs 14 | Real gap |
| Positional masking | 11 refs vs 0 | Real divergence, **different cause** |

Acting on the line counts would have meant ~700 lines of unnecessary porting, plus dragging a Loom-specific allowlist into a tool-agnostic guard. The harness is what prevented that, and it also localized the real bugs — four checks scanning the raw command instead of the redacted working copy.

**The harness then died with the session.** It only ever existed in a scratch directory.

## Why this matters beyond convenience

rjwalters/loom#5660 documents the disease: the vendored copy drifted ~2,200 lines ahead of its upstream while its own header said not to hand-edit it. That drift was invisible for as long as it existed because nothing compared the two automatically.

We have now paid the cost of measuring the difference once. Without checking the harness in, the next drift is equally invisible and someone repeats the whole investigation — or worse, doesn't, and flips the capability probe on a line-count argument like #188 originally proposed.

There is a related asymmetry worth encoding: `hooks/repo/tests/test-guard-destructive.sh` now asserts the canonical guard is **never weaker** than the vendored copy (stricter passes and is named). That check currently covers only write-confinement shapes. The harness generalizes it to the whole pattern surface.

- [ ] A checked-in equivalence harness compares both guards' `permissionDecision` for a shared corpus of commands
- [ ] Seeded with the ~60 cases used in #192, covering at minimum: SQL DDL, force ops, cloud CLI, `git stash`, write confinement, lifecycle/systemctl, fastpath read-only commands, and dangerous phrases appearing as **data** in `-m` / `--body` / `--title` / `--notes` values
- [ ] Applies the same *never weaker* rule as `test-guard-destructive.sh` rather than exact equality: equal passes, canonical-stricter passes **and is reported by name**, canonical-weaker fails
- [ ] Known deliberate divergences are declared in one place with their rationale, so they read as decisions rather than drift (today: `aws iam delete-role` denies here, asks in the vendored copy)
- [ ] **Skips cleanly** when `.loom/hooks/guard-destructive-generic.sh` is absent — a non-Loom-managed checkout must not fail, and a silent skip must be visible in the summary, not indistinguishable from a pass
- [ ] Wired into `hooks/repo/tests/run.sh` with the same delegation shape as the other suites
- [ ] Case fixtures containing destructive literals are written as data files, never shell heredocs — the guard denies its own test corpus otherwise (hit repeatedly during #192)

## Note

Case authoring has a genuine trap: several probe strings (`DROP TABLE`, `rm -rf /`, cloud deletion verbs) trip the guard when they appear in a shell command line, including inside a heredoc used to write the corpus. Bodies and fixtures must be produced with a file-write, not shell input.

---

### LINKED ISSUE #194: stash scope guard does not thread `git -C <path>`, so that shape escapes the ask
## Context

Carried over deliberately while porting the `git stash` scope guard into the canonical guard (#188, PR #192), and documented there as a known limitation rather than silently inherited.

`resolve_stash_cwd()` threads a `cd <dir> &&` prefix through the *same* compound command, so this resolves scope against the `cd` target rather than the hook's session cwd:

```
cd <worktree> && git stash pop
```

It does **not** thread a `git -C <path>` argument. So this shape is not caught:

```
git -C <main-checkout-path> stash pop
```

run from a linked-worktree cwd. The guard resolves scope against the worktree, concludes the operation is not in the main checkout, and allows it — while git actually operates on the main checkout's stash stack.

The sibling parser `parse_force_ops()` **does** handle `git -C`, so the asymmetry is within this file, not a repo-wide convention.

## Why it matters

`refs/stash` is a single stack shared by every linked worktree of a repo — not per-worktree. The whole reason the stash scope guard exists is that `pop`/`drop`/`clear` can destroy state another agent (or the operator) is relying on. `git -C` is a perfectly ordinary way to reach the main checkout from a worktree, and an agent that has been denied or asked once has an obvious reason to reach for it.

This is the same shape as the incident that motivated write confinement (Loom #4178): a session denied on one path retries through an equivalent one the guard does not model.

## Provenance

The limitation is inherited from Loom's vendored copy, whose own comment says:

> KNOWN LIMITATION: unlike the force-op parser (parse_force_ops), this check does not thread a `git -C <path>` argument — `git -C <main-checkout-path> stash pop` run from a worktree cwd is not caught today. Track any observed bypass via this path as a follow-up.

It was kept verbatim in #192 so the two guards would stay behaviorally identical while parity was being established — closing it there would have created a divergence in the middle of a change whose entire purpose was to remove them. This issue is that follow-up, now that parity is the baseline.

Note the direction: fixing this makes the canonical guard **stricter** than the vendored copy, which the equivalence assertion in `hooks/repo/tests/test-guard-destructive.sh` explicitly permits (it asserts *never weaker*, not *equal*). No coordination with Loom is required to land it, though rjwalters/loom#5660 is the natural place to mention it so the vendored copy can follow.

- [ ] `resolve_stash_cwd()` threads a `git -C <path>` argument the way `parse_force_ops()` already does, including the quoting treatment (`strip_cd_quoting` classification, raw value threaded for `curcwd`)
- [ ] `git -C <main-checkout> stash pop` issued from a linked-worktree cwd asks, matching the verdict for `cd <main-checkout> && git stash pop`
- [ ] Relative and quoted `-C` arguments resolve against the correct base; an unresolvable one fails toward **ask**, never silently toward allow (mirrors the existing `stash-scope:cd-unresolved` branch)
- [ ] `git -C <worktree> stash pop` issued from the main checkout is **not** over-blocked
- [ ] Regression coverage in `hooks/repo/tests/test-guard-destructive.sh`
- [ ] The inherited "KNOWN LIMITATION" comment is updated rather than left contradicting the code
- [ ] The new stricter-than-vendored verdict is declared as a deliberate divergence wherever those are tracked
---

### Verified finding: the gap is more severe than "mis-scoped" — the ask doesn't fire at all (2026-08-08, verified against `origin/main` @ `50fad29`)

Reading `hooks/repo/guard-destructive.sh` on `origin/main` (the local checkout
here is 39 commits behind and predates PR #192, which already merged this
guard onto `origin/main` — verified with `git fetch origin && git log
origin/main`), the entry-point trigger for the whole stash-scope ask block is
a regex, not just `resolve_stash_cwd()`'s cwd resolution:

```
hooks/repo/guard-destructive.sh:4638
echo "$COMMAND_ASK_SCAN" | grep -qE '(^|[;&|(]|[[:space:]])git[[:space:]]+stash[[:space:]]+(pop|drop|clear)([[:space:]]|$)'
```

This requires `git` and `stash` to be separated by whitespace only. Verified
directly:

```
$ echo 'git -C /some/main/checkout stash pop' | grep -qE '(^|[;&|(]|[[:space:]])git[[:space:]]+stash[[:space:]]+(pop|drop|clear)([[:space:]]|$)' && echo MATCHED || echo "NOT MATCHED"
NOT MATCHED
```

So `git -C <path> stash pop` never reaches `resolve_stash_cwd()` at all — the
`ask()` call is skipped entirely and the command silently allows. The issue
body's framing ("resolves scope against the worktree, concludes the operation
is not in the main checkout, and allows it") describes a *milder* symptom
(wrong-but-triggered) than what's actually on `origin/main` today
(never-triggered). Same root incentive (an agent reaching for an equivalent
untried path), worse mechanism.

Separately, even if the entry regex is fixed, `resolve_stash_cwd()`'s own
per-segment matching only recognizes a stash op when the token immediately
after `git` is literally `stash`:

```
hooks/repo/guard-destructive.sh:2492 (inside resolve_stash_cwd's awk body)
if (toks[1] == "git" && m >= 3 && toks[2] == "stash" && (toks[3] == "pop" || toks[3] == "drop" || toks[3] == "clear")) {
```

`git -C <path> stash pop` has `toks[2] == "-C"`, not `"stash"` — so this
branch also needs the same global-option walk `parse_force_ops()` already
does (see Affected Files below) before it will recognize the op at all, quite
apart from tracking `cpath` for scope resolution.

**Net effect on the acceptance criteria below**: AC1 ("`resolve_stash_cwd()`
threads a `git -C <path>` argument") is necessary but not sufficient by
itself — the entry-point regex at line 4638 needs the equivalent fix too, or
the fixed `resolve_stash_cwd()` is never reached for this shape. Added as AC1a
below.

### Additional acceptance criterion (AC1a, curator addition)

- [ ] The entry-point trigger regex (`hooks/repo/guard-destructive.sh:4638`)
      recognizes `git -C <path> stash (pop|drop|clear)` as a stash op, not
      just the adjacent `git stash (pop|drop|clear)` form — otherwise a fixed
      `resolve_stash_cwd()` is never invoked for the shape this issue exists
      to catch. Verified this is currently a real gap, not a hypothetical
      one (see "Verified finding" above).

### Affected Files

- `hooks/repo/guard-destructive.sh`:
  - `resolve_stash_cwd()` (~line 2470–2504, as of `origin/main` @ `50fad29`) —
    add `-C <path>` tracking to the per-segment awk loop, mirroring
    `parse_force_ops()`'s global-option walk (~line 1500–1512: `while (k <=
    m) { ... if (t == "-C") { cpath = toks[k+1]; k += 2; continue } ... }`).
    Also update the stale "KNOWN LIMITATION" comment directly above the
    function (~line 2465–2468) rather than leaving it contradicting the code,
    per the issue's own acceptance criteria.
  - Entry-point trigger regex (~line 4638): `git[[:space:]]+stash[[:space:]]+(pop|drop|clear)`
    needs to admit an intervening `-C <path>` (and, for consistency with
    `parse_force_ops()`, any other global option `git` accepts before the
    subcommand) — otherwise the fixed `resolve_stash_cwd()` is never invoked
    for this exact shape. This is the AC1a gap identified above.
  - The three `ask()` call sites in the same block (~lines 4666, 4684, and the
    worktree-collision branch further down) build their human-readable
    messages from `$COMMAND` — no changes needed there, but worth confirming
    the message still reads sensibly when `$COMMAND` contains a `-C` flag.
- `hooks/repo/tests/test-guard-destructive.sh` — **currently has zero
  stash-scope test coverage of any kind** (verified: `grep -c stash
  hooks/repo/tests/test-guard-destructive.sh` against `origin/main` @
  `50fad29` returns 0 matches). The nearest existing precedent for a `-C`
  cwd-resolution test is the force-ops case at ~line 1154:
  `assert_ask "forceScope protected: git -C <feature-repo> reset --hard
  honors -C cwd" ...`. For the worktree-collision scenario, the
  `make_wt_confinement_repo()` helper (~line 3129) shows the pattern for
  building a throwaway repo with a real `git worktree add` linked worktree —
  reuse or adapt it rather than hand-rolling stash-specific fixture setup.
  Since there is no existing stash-scope suite to extend, this issue's test
  work is net-new: cover both the base `cd <dir> && git stash pop` shape (not
  currently tested either, per the same zero-match grep) and the `-C` shape
  this issue adds.

### Test Plan

- [ ] Automated tests in `hooks/repo/tests/test-guard-destructive.sh`:
  - [ ] `git -C <main-checkout> stash pop` issued from a linked-worktree cwd asks with `stash-scope:main-checkout` (mirrors `cd <main-checkout> && git stash pop`)
  - [ ] `git -C <worktree> stash pop` issued from the main checkout does NOT over-ask when no other worktree exists to collide with
  - [ ] A relative `-C <path>` argument resolves against the correct base (the invoking cwd, per git's own `-C` semantics)
  - [ ] A quoted (`'...'`/`"..."`) `-C` argument resolves the same as its unquoted form
  - [ ] An unresolvable `-C` target (nonexistent path) fails toward **ask** (`stash-scope:cd-unresolved`), never silently toward allow
  - [ ] Baseline regression: `cd <dir> && git stash pop` (the existing supported shape) still asks correctly after the change — guard against the fix narrowing the existing `cd`-tracking path while widening for `-C`
- [ ] Manual verification: run the guard directly against `git -C <main-checkout> stash pop` from a worktree cwd (using the `run_guard`/`make_input` test harness helpers already in the test file) and confirm the JSON output's `permissionDecision` is `ask`, not `allow`
- [ ] Confirm the updated "KNOWN LIMITATION" comment no longer contradicts the code once the fix lands

### LINKED ISSUE #197: Quoting a destructive argument weakens the guard verdict (rm -rf "/" is allowed)
## Observed

Quoting a destructive command's argument changes the guard's verdict, in ways that weaken it. Measured directly against four guard versions with the same payloads (cwd = this repo):

| Case | `main` | PR #192 builder commit | PR #192 head | Loom vendored |
|---|---|---|---|---|
| `rm -rf /` (bare) | deny | deny | deny | deny |
| `rm -rf "/"` (quoted) | **allow** | **allow** | **allow** | **allow** |
| force-push to `main` (bare) | deny | deny | deny | deny |
| force-push to `"main"` (quoted) | **ask** | **ask** | **ask** | **ask** |
| SQL DDL bare / quoted | deny | deny | deny | deny |

And separately, for Bash-tool write confinement (which exists only on #192's branch):

| Case | PR #192 head | Loom vendored |
|---|---|---|
| `echo x > <main-checkout>/evil.sh` | deny | allow¹ |
| `echo x > "<main-checkout>/evil.sh"` (quoted) | **allow** | **allow** |
| `echo x > '<main-checkout>/evil.sh'` (single-quoted) | **allow** | **allow** |

¹ the separately-tracked symlinked-ancestor bypass, fixed in #192 and reported to rjwalters/loom#5660.

**The verdicts are identical across `main`, both #192 commits, and the vendored guard**, so this is pre-existing and was not introduced or widened by #192. Filing it as its own issue rather than blocking that PR.

Reproduction: send `{"tool_input":{"command":"<cmd>"},"cwd":"<repo>"}` to `hooks/repo/guard-destructive.sh` and read `.hookSpecificOutput.permissionDecision`. Note the guard must be executed from its normal in-tree path — running a copy from `/tmp` fails open and produces misleading `allow` results for everything.

## Why it matters

`rm -rf "/"` and `rm -rf /` are the same command to the shell. A guard that denies one and allows the other is not enforcing a policy, it is enforcing a spelling. The bypass needs no cleverness — quoting a path is the *normal* thing to do, and an agent that has just been denied has an obvious reason to try it.

## Suspected cause (unverified)

The catastrophic and ask-tier scans read working copies produced by literal-text redaction (`strip_literal_text` / `strip_datasink_literals`), which exists to stop **prose** that merely quotes a dangerous command from tripping the guard — filing an issue about `rm -rf /`, committing a message that mentions it. That redaction appears not to distinguish *a quoted string used as prose in a data-sink flag* from *a quoted argument that is the operative target of the command*, so quoting the target removes it from the scanned text.

If that is right, the two requirements are in direct tension and the fix is a sink-aware distinction rather than more or less redaction:

- a quoted string after `-m` / `--body` / `--title` / `--notes` is **data** → redact (this works today and must not regress; it is what makes issue-filing about destructive commands possible at all)
- a quoted string that is a command's **positional argument or redirect target** is **operative** → must be unquoted and scanned

`strip_target_quoting()` already implements shell-accurate quote removal and is used for the write-confinement classification path, so the primitive exists; the gap is which scans consume it.

This is a hypothesis from reading the code during review, not a measured finding. A Builder who instruments the redaction pipeline and finds a different mechanism should close this with that evidence rather than forcing a fix onto this one.

- [ ] `rm -rf "/"`, `rm -rf '/'`, and `rm -rf /` all reach the same verdict
- [ ] A quoted branch name in a force-push to a protected branch denies, matching the bare form
- [ ] A quoted or single-quoted Bash-tool write target into the main checkout is confined exactly as the bare form is
- [ ] Prose redaction is preserved and regression-tested: quoting a destructive command inside `-m` / `--body` / `--title` / `--notes` still does **not** trip the guard (this is the property that made the ordinary-prose false positives fixable in #192, and losing it would be a worse regression than the bypass)
- [ ] Coverage for the escalating forms: partial quoting (`rm -rf "/"x`), concatenated quoting (`"/"` + `bin`), and backslash escaping (`rm -rf \/`)
- [ ] The fix is landed in coordination with the vendored copy, or after rjwalters/loom#5660 resolves how the two guards stay in step — fixing only one changes their relative strictness

## Related

- rjwalters/loom#5660 — vendored-guard drift and the single-marker capability probe; the same coordination problem applies to any fix here
- #193 — the equivalence harness that would have surfaced this asymmetry automatically had it existed
- #192 — where this was found; explicitly **not** a regression from it

rjwalters:
Closes #193, closes #194, closes #197.

**#195 and #196 are deliberately NOT closed here** — the fleet built them in parallel and they merged first via #199 and #200. This branch's duplicate implementations have been dropped in favour of the merged ones.

## #193 — guard-equivalence harness

#188 established parity with Loom's vendored guard by hand, and the measurement changed the conclusion: counting functions said four capability gaps, comparing verdicts said two. `systemctl` and fastpath tiering diverged on **zero** cases and were correctly not ported, avoiding ~700 lines of churn plus a Loom-specific allowlist in a tool-agnostic guard. That harness then died with the session.

Now permanent: an **89-case** corpus (`commands/repo/tests/guard-equivalence-cases.txt`, 89 non-comment rows) run against both guards from a scratch repo, wired into `pnpm test` via `hooks/repo/tests/run.sh` so it gates CI rather than living as a standalone script.

**The rule is "never weaker", not "equal".** Exact equality is the wrong invariant — it already failed once, when fixing the symlinked-ancestor bypass in #192 turned a test about protection *strength* red. Equal passes; canonical-stricter passes and is **named**; canonical-weaker fails.

It also carries a `DECLARED_WEAKER` list (`test-guard-equivalence.sh:110-113`, consulted at `:243-254`) for the inverse case — `echo`/`printf` of a destructive string prints it rather than running it, so the vendored deny is a false positive and being correct there means being more permissive. **An undeclared weaker verdict is still a hard failure** (`FAIL` + `exit 1`); the list only affects labeling of the two declared rows. Current run: 7 stricter, 2 declared-weaker, 0 undeclared.

The corpus is a data file on purpose: several cases contain literals the guard denies, so generating them from a shell heredoc trips the guard on its own fixtures.

## #194 — `git -C` threading in the stash-scope guard

`git` resolves `-C` against the process cwd and then operates there, so `git -C <main-checkout> stash pop` from a linked worktree touched the **main** stash stack while a cwd-only check saw only the worktree. `refs/stash` is one stack shared across every linked worktree, so this is a live path to destroying another agent's WIP.

**The pre-check regex was the real gate** — the parser never ran for that shape, so the parser-only fix the issue described would have changed nothing.

**Review round 2 found this fix incomplete, and the finding was correct.** A quoted `-C` value containing whitespace still bypassed it *silently*. Three defects, each hiding the next:

1. **The pre-check regex.** `[^[:space:]]+` required zero whitespace in the flag value. Widening it to accept quoted spans still missed **mixed** tokens like `-c user.name="John Doe"`. Enumerating token shapes is the wrong approach — this gate only needs to be permissive enough not to *miss*, since the parser behind it decides scope. It now matches the flag run non-greedily up to `stash` (`guard-destructive.sh:5087`):

   ```
   (^|[;&|(]|[[:space:]])git[[:space:]]+([^;&|]*[[:space:]]+)?stash[[:space:]]+(pop|drop|clear)([[:space:]]|$)
   ```

   `[^;&|]` keeps the widened span from crossing a shell separator. Over-inclusive here costs a parser call, never a wrong verdict.

2. **The whitespace tokenizer.** `resolve_stash_cwd` split segments on **raw** whitespace before its quote-aware helpers could act, shredding a quoted path into two tokens. It now masks whitespace inside quoted spans via `mask_ws` (from `_MASKWS_AWK`) before tokenizing, and unmasks the `cd` and `-C` arguments individually with `unmask_ws` before use.

3. **A missing awk dependency, which is why the first two fixes appeared to do nothing.** `resolve_stash_cwd`'s awk program composes `_QSPLIT_AWK`, and its `qsplit` calls `bs_escaped` (`guard-destructive.sh:1128` → defined at `:1042` in `_ESCAPE_AWK`) — but `_ESCAPE_AWK` was never prepended to the composition. awk aborted at **runtime**, so the parser returned nothing, the caller fell back to the session cwd, and the command was classified as a worktree-local stash needing no ask. A silent fail-open that reads as a clean allow. The fix prepends `_ESCAPE_AWK` alongside the new `_MASKWS_AWK`.

Also adds the stash coverage this suite never had — #188 shipped the stash guard with **zero** tests — now **19** `assert_*` cases in `test-guard-destructive.sh`: 14 for `-C`/`cd` threading, plus 5 whitespace cases (`:3701-3710`) built on a dedicated `make_wt_confinement_repo_spaced()` fixture (`:3243`) whose main checkout path contains a literal space. Both quote styles and the mixed `-c user.name="John Doe"` form are covered.

## #197 — quoting no longer weakens the verdict

A quoted root path was **allowed** where the bare form denied; a force-push to a quoted branch fell through to an ask; a quoted write target escaped confinement. Same commands to the shell — the guard was enforcing a *spelling*, not a policy.

Pre-existing on `main` and shared with the vendored copy; measured identical across four guard versions before being touched.

Two causes, **neither the one the issue guessed at**:

1. The literal catastrophic patterns are command text, so quotes broke the match. Fixed by **ordering**, not by redacting less: `dequote_inert_spans()` derives a dequoted copy *after* sink-aware redaction has already blanked `--body`/`-m`/`--title` values and `echo`/`printf` data, so prose that quotes a dangerous command is inert before dequoting can see it. Scanned in addition to the raw copy; `$(`/backtick spans stay intact.
2. Write confinement went blind to a quoted redirect target — **only after `echo`/`printf`**. The data-sink redaction was blanking the target as though it were an argument to `echo`; the shell hands it to the *redirection*. `cat`, `tee`, `cp`, `mv`, `sed -i` were unaffected, which is what localized it.

The regression risk runs the **other** way — making a read-only command look dangerous — so **six corpus rows** pin it (`guard-equivalence-cases.txt:130-135`: `grep -rn`, `grep -F`, `rg --fixed-strings`, `awk`, `printf`, `echo`) and all six still allow, while a payload piped into a shell still denies. Four matching `assert_allow` cases plus `echo 'rm -rf /' | sh` denying are pinned in `test-guard-destructive.sh`.

## Note on the parallel work

`main` gained #199 and #200 while this branch was in review. Rebasing surfaced that both had landed **different, more conservative** implementations than mine — notably #199 is config-only (no env override), ASK-tier only, and carries a mandatory exclusion set (`grep`/`cp`/`mv`/`tee`/`sed`) that no config can override. My versions are dropped; the CHANGELOG entries are rewritten to describe what actually shipped, since neither #199 nor #200 added one.

## Test Plan

Run the full gate — both suites below are wired into it, so this single command covers everything:

```bash
pnpm test        # -> ./hooks/repo/tests/run.sh
```

Expected, and the result on this head (`52f7599`, CI run 31245129632, green):

| Suite | Result |
|---|---|
| `test-guard-destructive.sh` (full regression) | **680 pass, 0 fail** |
| `test-guard-equivalence.sh` (vendored-guard equivalence) | **89 pass, 0 fail** — 7 stricter, 2 declared-weaker, 0 undeclared |
| **Total across all delegated suites** | **2106 pass, 0 fail** |

The two suites can also be run standalone:

```bash
bash hooks/repo/tests/test-guard-destructive.sh
bash commands/repo/tests/test-guard-equivalence.sh
```

Targeted checks for the #194 regression, from a linked worktree whose main checkout path contains a space (expect `ask` for the first three, silence/allow for the last two):

```bash
git -C "<main with space>" stash pop     # ask
git -C '<main with space>' stash drop    # ask
git -c user.name="John Doe" stash pop    # ask
git -C "<the worktree>" stash pop        # allow (worktree-local)
git stash list                           # allow (read-only)
```

> **Note if you run these locally:** a shell exporting `LOOM_FORCE_SCOPE` or `LOOM_GUARD_DECISION_LOG` (dispatched sweep/daemon children inherit both) can skew this repo's own guard tests. Prefix with `env -u LOOM_FORCE_SCOPE -u LOOM_GUARD_DECISION_LOG`.


rjwalters:
## Blocking finding addressed — and it was worse than reported

Confirmed the bypass, and thank you for catching it: the pre-check regex required the `-C`/`-c` value to contain zero whitespace, so any quoted value with a space failed the positional match and the ask was skipped **silently** — a clean allow, not even an ask.

Fixing it turned up **three** defects, each hiding the next:

**1. The pre-check regex.** Widening it to accept quoted spans fixed the `-C` forms but still missed **mixed** tokens like `-c user.name="John Doe"`, where one argument is part bare and part quoted. Enumerating token shapes is the wrong approach — this gate only needs to be permissive enough not to *miss*, since the parser behind it decides scope. It now matches the flag run non-greedily up to `stash`. Over-inclusive here costs a parser call, never a wrong verdict.

**2. The whitespace tokenizer.** As you noted, `resolve_stash_cwd`'s inner parser is quote-aware via `qsplit`/`strip_cd_quoting` — but it split segments on **raw** whitespace before that helped, so a quoted path containing a space became two tokens. Now masks whitespace inside quoted spans with `mask_ws` before tokenizing, unmasking the `cd` and `-C` arguments individually before use.

**3. A missing awk dependency — the one that made this hard to see.** `mask_ws` lives in `_MASKWS_AWK` and itself calls `bs_escaped` from `_ESCAPE_AWK`. Omitting that dependency made awk abort at **runtime**, so `resolve_stash_cwd` returned nothing, the caller fell back to the session cwd, and the command was classified as a worktree-local stash — which with a single worktree needs no ask. A silent fail-open that reads as a clean allow, which is exactly why the first two fixes appeared to do nothing.

## Two false starts, recorded because both looked like fixes

- Defining `unmask_ws` locally in the parser. It already exists in `_MASKWS_AWK`, and **awk rejects a duplicate function definition** — so the parser failed open on *every* stash command, including the basic one. Caught because the control case regressed.
- Testing a **bare** unquoted `-C` against a path containing a space. That is ambiguous shell — the shell hands only the first word to `-C` — so my own test expectation was wrong. The space-free fixture is the meaningful regression check for the bare form.

## Verification

New spaced-path fixture plus 5 regression cases, all passing:

```
PASS: -C with a DOUBLE-QUOTED path containing a space asks
PASS: -C with a SINGLE-QUOTED path containing a space asks
PASS: a -c value containing a space does not break the gate
PASS: -C at a worktree from a spaced main still does not ask
PASS: stash list in a spaced main stays read-only
```

Full suite **2101 pass, 0 fail**.

Also worth flagging for the record: your read on the declared-divergences allowlist was right — it only affects *labeling*, never the pass/fail decision. Since your review the harness also grew a `DECLARED_WEAKER` list for the inverse case (this guard is deliberately more permissive where the vendored copy false-positives on `echo` of a destructive string), which does gate: an **undeclared** weaker verdict is still a hard failure.

Ready for re-review.

rjwalters:
## Description corrected — and one correction back, with evidence

Thank you for the `-C` whitespace catch in round 1; that was a real silent bypass and it's fixed (details in the previous reply). Round 2's core ask is right and done: **the PR body now closes only #193, #194, #197**.

### On #197 — it *is* fixed here

> `test-guard-equivalence.sh`'s own output in this PR shows `ok allow rm -rf "/"` … There is no dequoting/sink-aware-redaction change anywhere in the diff.

That output is from commit `9575374`, the **first of six** commits on the branch. The fix landed in `2ce4f4e`, which added `dequote_inert_spans()` and reordered it after the sink-aware redaction. On the current head the same corpus rows read:

```
ok    STRICTER canonical=deny  vendored=allow rm -rf "/"
ok    STRICTER canonical=deny  vendored=ask   git push --force origin "main"
```

Verified live against the built guard: a quoted root path denies, a single-quoted one denies, a quoted force-push target denies, and a quoted Bash-tool write target is confined. `grep -c dequote_inert_spans hooks/repo/guard-destructive.sh` → 3.

The review appears to have been computed against a single commit rather than the branch — the same reading that produced "the diff is a single commit (`9575374`) touching 5 files, +495/-5", where GitHub reports the PR as 6 commits, +1069/-15 across 9 files. Worth flagging because the conclusion it led to — that merging would auto-close a still-unfixed security issue — was the right thing to block on *if* the premise had held.

### On #195 / #196 — you were right, for a reason neither of us had

They were built **in parallel by the fleet** and merged first as #199 and #200, while this branch carried its own implementations. Rebasing onto the new `main` left **both** versions in the tree: two positional-masking implementations with different config keys, and two copies of the changelog-coherence prose.

I've dropped mine. The merged ones are the better design and it isn't close — #199 is config-only with no environment override, confines masking to the **ASK tier** rather than also covering the catastrophic deny patterns, and carries a **mandatory exclusion set** (`grep`/`egrep`/`fgrep`/`rg`/`cp`/`mv`/`tee`/`sed`) that no config can override. Mine had none of those limits.

One thing that needed fixing regardless: **neither #199 nor #200 added a CHANGELOG entry**, so the release would have shipped both features undocumented. The entries here are rewritten to describe the shipped implementations — documenting my version would have named a config key that does not exist and an escape hatch wider than the real one.

### Your trivial fix

`cd0fc3a8` is a good catch and is carried through the rebase — I had left a `KNOWN LIMITATION: does not thread -C` comment directly above the code that now threads it.

### State

- Rebased onto current `main` (includes #199, #200); `mergeStateStatus` CLEAN
- Body closes **#193, #194, #197** only; #195/#196 referenced as already-shipped, non-closing
- Full suite **2106 pass, 0 fail**

Ready for re-review.

PR #211, opened 2026-08-09
docs(repo): stop the README layout block drifting file-by-file

rjwalters:
## Summary

A `/repo:all` sweep found the README's **Repository layout** block six suites
behind disk. It listed 10 of the 16 test suites `hooks/repo/tests/run.sh`
actually delegates to, and omitted `scripts/repo/repo-scrub-forks.sh` entirely.

`/repo:docs` step 2 checks "do file trees and directory listings match disk?",
so this block is meant to track reality — it just had no mechanism keeping it
there.

## Why globs rather than adding the six missing rows

Per-file enumeration is the defect, not the symptom. Every new suite needs a
matching README line, and the last four PRs to add one forgot it. Filling in
today's six rows buys accuracy until the next suite lands.

So the individual test rows become two globs — `hooks/repo/tests/test-*.sh` and
`commands/repo/tests/test-*.sh` — each annotated with what the group covers.
Adding a suite no longer requires a README edit to stay accurate.

The tradeoff is losing the one-line-per-suite descriptions. The group
annotations name all 16 (5 hook, 11 command), so coverage is still discoverable
from the README; what's gone is the file→description mapping. `run.sh` keeps its
own row as the entry point, now noting `pnpm test` runs it.

## Changes

- Add `scripts/repo/repo-scrub-forks.sh` alongside the other two installed scripts
- Replace 10 individual test rows with two annotated globs
- Note that `pnpm test` runs `run.sh`

## Test plan

- [x] `./hooks/repo/tests/run.sh` — 2138 pass, 0 fail (no behavior change; README-only)
- [x] Group annotations cross-checked against disk: 5 hook suites, 11 command
      suites, matching `ls` exactly
- [x] Every suite named in the annotations confirmed present in `run.sh`'s
      delegation list

## Notes for the reviewer

The glob-vs-explicit call is the judgment worth a second opinion. If you'd
rather keep per-file rows, the alternative is adding the six missing ones and
accepting that this recurs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,18 +156,18 @@ skills/repo/SKILL.md         Domain overview installed to .claude/skills/repo/
 commands/repo/*.md           Command files installed to .claude/commands/repo/
 scripts/repo/repo-remote.sh  Headless /repo:remote provisioning entry point, installed to .claude/skills/repo/scripts/
 scripts/repo/resync-installed.sh  Consumer-side resync (contract C7), installed to .claude/skills/repo/scripts/
+scripts/repo/repo-scrub-forks.sh  /repo:scrub fork-network sweep, installed to the same place
 hooks/repo/guard-destructive.sh  PreToolUse guard hook installed to .claude/skills/repo/hooks/
 hooks/repo/session-start-handoff.sh  SessionStart handoff-note hook installed to the same place
-hooks/repo/tests/run.sh      Smoke suite covering both hooks (bash, no framework needed)
-hooks/repo/tests/test-guard-destructive.sh  Full guard regression suite (ported from Loom)
-hooks/repo/tests/test-session-start-handoff.sh  Handoff-hook suite (run.sh delegates to it)
-hooks/repo/tests/test-install-claude-md-markers.sh  CLAUDE.md marker-block regression suite
-hooks/repo/tests/test-shell-wrapper.sh  claude + codex shell wrapper suite (run.sh delegates to it)
-commands/repo/tests/test-branches-loss-check.sh  branches.md permanent-loss check suite (run.sh delegates)
-commands/repo/tests/test-repo-remote.sh  repo-remote.sh provisioning-contract suite (run.sh delegates)
-commands/repo/tests/test-verify-fix-persistence.sh  verify-after-write contract suite (run.sh delegates)
-commands/repo/tests/test-resync-installed.sh  resync-installed.sh suite (run.sh delegates)
-commands/repo/tests/test-installer-contract.sh  INSTALLER-CONTRACT.md C1-C8 conformance (run.sh delegates)
+hooks/repo/tests/run.sh      Test entry point (bash, no framework needed): inline smoke cases
+                             plus every delegated suite below. `pnpm test` runs it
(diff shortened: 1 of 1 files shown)


rjwalters:
✅ **Approved!**

Docs-only change to the README's *Repository layout* block. Reviewed in an isolated worktree at head `8b91cb8`.

## Verification

**Glob accuracy — matches disk exactly:**
- `hooks/repo/tests/test-*.sh` → 5 files on disk; annotation names 5 (guard regression, handoff hook, CLAUDE.md markers, sidecar untracking, shell wrappers)
- `commands/repo/tests/test-*.sh` → 11 files on disk; annotation names 11
- `guard-equivalence-cases.txt` is the only non-`.sh` file in those dirs and the glob correctly misses it — the annotation calls it out parenthetically rather than leaving it undocumented

**No hidden coupling to the literal paths that were removed:**
- Nothing under `lib/`, `install.sh`, `uninstall.sh`, `scripts/`, or either test dir parses the layout block. The `grep` hits for `README.md` in `test-tidy-keep-tiers.sh` / `test-early-sync-switch.sh` / `test-verify-fix-persistence.sh` are all fixture READMEs inside temp repos, unrelated to this file.
- No other prose in `README.md` references an individual test file, so removing the per-file rows leaves nothing dangling.
- `INSTALLER-CONTRACT.md` cites two test paths, but in its own text — unaffected.

**Does not trip the repo's own checkers:** `/repo:readme`'s File Tree Accuracy check is scoped to ASCII trees (blocks containing `├──`/`└──`); this is a plain two-column listing and was never matched, before or after. And `commands/repo/*.md` was already a glob in this same block, so the convention predates this PR rather than being introduced by it.

**Tests:** `./hooks/repo/tests/run.sh` → **2138 pass, 0 fail** (re-run independently in `.loom/worktrees/pr-211`, with `LOOM_FORCE_SCOPE`/`LOOM_GUARD_DECISION_LOG` confirmed unset per repo#5388). CI green, `mergeStateStatus=CLEAN`.

## On the glob-vs-explicit call

The PR body flags this as the judgment worth a second opinion, so: **globs are the right call here.** The block had drifted on six suites plus a script, and per-file enumeration is a listing that requires an edit to stay true — it had already failed that four PRs running. The cost is losing the file→description mapping, but the annotations name all 16 suites and the filenames are self-describing (`test-installer-contract.sh` ↔ "installer contract"), so the mapping is recoverable by inspection.

Incidental second fix worth noting: `run.sh`'s old description, "Smoke suite covering both hooks," understated it considerably — it delegates 16 suites. The new wording corrects that too.

## Observation — not filed, operator's call

This PR makes drift *less likely* but nothing *enforces* it: no test asserts that every path and glob in the layout block resolves to something on disk. That's the gap that let six suites accumulate unlisted. A small check would close it permanently. Deliberately not filing an issue for it — flagging it here so the operator can decide rather than adding unrequested queue noise.

## Issue linking

No `Closes #N`, and that is correct here — this PR has no underlying issue. It originated from a `/repo:all` hygiene sweep, not from the issue queue, so there is nothing to auto-close.

